PR #195, opened 2021-08-19
add openapi def to docs

olensmar:


api/v1/kubtest.yaml
```diff
@@ -1,23 +1,19 @@
...
+  title: Kubtest API
+  description: 'Kubtest provides a Kubernetes-native framework for test definition, execution and results'
   contact:
-    email: api@kubtest.io
+    email: kubtest@kubshop.io
   license:
     name: MIT
     url: https://opensource.org/licenses/MIT
-  version: 0.0.5
+  version: 1.0.0
```
exu, line 11: @olensmar can we put here tag ?

exu (changes requested):

api/v1/kubtest.yaml
exu: or take it somehow when building (maybe some placeholder ? don't know what possibilities are there

exu (approved):

api/v1/kubtest.yaml
olensmar: yeah - not sure either... let's go with this for now

PR #234, opened 2021-08-25
added async calls to start execution 

exu:
+ - fixed concurency problem with mongo+workers

## Changes

- start script is async since now
- we can get result with additional `script excution` command 

## Fixes

- invalid QueuePull method inexecutor

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cmd/kubectl-kubtest/commands/scripts/start_script.go
```diff
@@ -39,26 +41,40 @@ var StartScriptCmd = &cobra.Command{
...
+
+		execution := scriptExecution.Execution
 
-		result := scriptExecution.Execution.Result
 		switch true {
 
-		case scriptExecution.Execution.IsSuccesful():
-			fmt.Println(result.RawOutput)
-			ui.Success("Script execution completed with sucess")
+		case execution.IsPending():
```
exu: ispending 2x

PR #249, opened 2021-08-30
Multiple executors handling + input from git

exu:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/v1/kubtest.yaml
```diff
@@ -500,9 +500,29 @@ components:
...
           description: script content as string (content depends from executor)
+        input-type:
+          type: string
+          description: >
+            script content type can be: 
+            - direct content - created from file, 
+            - git repo directory checkout in case when test is some kind of project or have more than one file, 
+            - or file uri
+          enum:
+            - content
```
olensmar, line 514: should we have a binary-content? (which would be an encoded string)

api/v1/kubtest.yaml
```diff
@@ -500,9 +500,29 @@ components:
...
+            - git
```
olensmar, line 515: suggest repo - with a corresponding "repository" input object containing the properties:
- type (enum, always git)
- uri
- branch
- folder

api/v1/kubtest.yaml
```diff
@@ -507,20 +507,30 @@ components:
...
           enum:
             - content
             - git
-            - uri
+        repository: 
+          $ref: '#/components/schemas/Repository'
+
+    Repository: 
+      description: scripts execution request body
+      type: object
```
olensmar, line 520: add required section - type, uri, branch?

api/v1/kubtest.yaml
exu: i think we could easily refactor when we'll get some executor which could execute such input

olensmar (approved):

nicufk (approved):

exu:
@olensmar  added repository as object

exu:
@olensmar added `/executions` and `/executions/:id` endpoints

olensmar:
thanks @exu - although I don't see it in the OpenAPI definition!?

exu:
> thanks @exu - although I don't see it in the OpenAPI definition!?

added @olensmar

exu:
> thanks @exu - although I don't see it in the OpenAPI definition!?

added openapi routes too

pkg/executor/server/handlers.go
```diff
@@ -0,0 +1,57 @@
...
+
+		s.Log.Infow("starting new execution", "execution", execution)
+		c.Response().Header.SetStatusCode(201)
+		return c.JSON(execution)
+	}
+}
+
+func GetExecution(s server.HTTPServer, repository result.Repository) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		execution, err := repository.Get(context.Background(), c.Params("id"))
```
nicufk, line 50: It will be better to use the context from the request.

nicufk (approved):

cmd/kubectl-kubtest/commands/scripts/create_script.go
```diff
@@ -4,16 +4,22 @@ import (
...
 
 func init() {
-	// TODO find a good way to handle short flags
 	CreateScriptsCmd.Flags().StringP("name", "n", "", "unique script name - mandatory")
 	CreateScriptsCmd.Flags().StringP("file", "f", "", "script file - will be read from stdin if not specified")
 
+	// TODO - type should be autodetected
 	CreateScriptsCmd.Flags().StringP("type", "t", "postman/collection", "script type (defaults to postman-collection)")
+
+	CreateScriptsCmd.Flags().StringP("uri", "", "", "if resource need to be loaded from URI")
```
jasmingacic, line 20: PersistentFlags can inject value directly into a variable. It is rather convenience so you don't have to get them `cmd.Flag("prop_name").Value.String()`

cmd/kubectl-kubtest/commands/scripts/create_script.go
```diff
@@ -4,16 +4,22 @@ import (
...
 	"github.com/kubeshop/kubtest/pkg/ui"
 	"github.com/spf13/cobra"
 )
 
 func init() {
-	// TODO find a good way to handle short flags
 	CreateScriptsCmd.Flags().StringP("name", "n", "", "unique script name - mandatory")
 	CreateScriptsCmd.Flags().StringP("file", "f", "", "script file - will be read from stdin if not specified")
 
+	// TODO - type should be autodetected
```
jasmingacic, line 17: Each cobra CLI command has `Prerun`  where auto detection can be implemented.

cmd/kubectl-kubtest/commands/scripts/create_script.go
```diff
@@ -47,11 +56,27 @@ var CreateScriptsCmd = &cobra.Command{
 			ui.Failf("Script with name '%s' already exists in namespace %s", name, namespace)
 		}
 
-		if len(content) == 0 {
+		if len(content) == 0 && uri == "" {
```
jasmingacic, line 59: you can do `len(uri)==0` for sake of aesthetics

cmd/kubectl-kubtest/commands/scripts/create_script.go
jasmingacic:
```
var output string 

cmd.PersistentFlags().StringVarP(&output, "out", "o", "default value", "something cool to say")
```

pkg/executor/server/handlers.go
exu: 👍🏻  good catch

exu: @nicufk created separate issue for that - as there is quite a lot to change

api/v1/kubtest.yaml
```diff
@@ -15,17 +15,95 @@ externalDocs:
...
+          description: 'problem with getting script executions from storage'
+          content:
+            application/problem+json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Problem'
+  /executions:
+    get:
+      parameters:
```
olensmar, line 69: i think you can get rid of this parameter

PR #257, opened 2021-09-02
check for KUBECONFIG and if APISERVER_PORT is not taken

jasmingacic:
This is probably not the best place to do it but it will save some time.


Makefile
```diff
@@ -8,7 +8,7 @@ DATE ?= $(shell date -u --iso-8601=seconds)
 COMMIT ?= $(shell git log -1 --pretty=format:"%h")
 
 run-api-server: 
-	APISERVER_PORT=8080 go run cmd/api-server/main.go
+	APISERVER_PORT=8087 go run cmd/api-server/main.go
```
exu, line 11: This port is used by kubectl kubtest direct client - and  need to be changed too

exu (changes requested):

cmd/api-server/main.go
```diff
@@ -23,12 +27,32 @@ func init() {
 }
 
 func main() {
+
+  if os.Getenv("KUBECONFIG") == "" {
+		fmt.Println("KUBECONFIG not present. Exiting...")
+		os.Exit(1)
+	}
+
+	port := os.Getenv("APISERVER_PORT")
```
olensmar, line 36: indentation seems wrong!?

olensmar (approved):

cmd/api-server/main.go
exu: no it's TAB - depends on display in editor/viewer :)

exu (approved):

PR #263, opened 2021-09-03
fixed openapi definition and changed listExecutions operations …

olensmar:
…to return ExecutionSummary

This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


nicufk (approved):
👍

internal/app/api/v1/server.go
```diff
@@ -46,8 +46,8 @@ func (s kubtestAPI) Init() {
 
 	executions := s.Routes.Group("/executions")
 
-	executions.Get("/executions", s.ListExecutions())
-	executions.Get("/executions/:id", s.GetScriptExecution())
+	executions.Get("/", s.ListExecutions())
+	executions.Get("/:executionID", s.GetScriptExecution())
```
exu, line 50: It will brake the code in handler 
Additionally keep in mind that both routes (this one and one on top of script resource )  are managed by single handler and Paramus need to be the same

exu (changes requested):

internal/app/api/v1/server.go
exu: It's ok just checked you've fixed this 😁

exu (approved):
LGTM

hamzahsn (approved):
LGTM! I think we also can keep the name of the execution to show it as the executionID 😄

jasmingacic (approved):

PR #278, opened 2021-09-08
added v to versions

exu:
For fixing deps in executors

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


pkg/version/version_test.go
```diff
@@ -0,0 +1,41 @@
...
+	t.Run("test get newest version with v prefix", func(t *testing.T) {
+		versions := []string{
+			"v0.1.1",
+			"0.3.2",
+			"v0.3.3",
+			"0.2.2",
+		}
+
+		newest := GetNewest(versions)
+		assert.Equal(t, "0.3.3", newest)
```
nicufk, line 20: Isn't it supposed to test version with v prefix?

exu: yes it is look at top of change - generally semver package strips v in version string - but is able to read and process it (it's behaving like some kind of prefix and is often omitted)

exu: @nicufk

nicufk: 👍

nicufk (approved):

jasmingacic (approved):

PR #297, opened 2021-09-15
first try to integrate an agent into kubtest

jasmingacic:
This PR replaces call to executor endpoint and spins up k8s jobs instead.

Missing parts:
- Persistent wolumes
- Agents images pushed to docker image repo. Also agents need to be normalized so each of them accepts the same input right now there is a significant difference between cypress and other agents. for more details see `https://github.com/kubeshop/kubtest-executor-k8sjobs/pull/1`

p.s. implementation is very crude as I want to open up a discussion.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


exu (commented):
Put this in WIP draft PR mode

internal/app/api/v1/scripts.go
```diff
@@ -173,21 +175,34 @@ func (s kubtestAPI) ExecuteScript() fiber.Handler {
...
-				l.Debugw("saving - debug", "scriptExecution", scriptExecution)
+			// l := s.Log.With("executionID", e.Id, "status", e.Status, "duration", e.Duration().String())
+			// l.Infow("saving", "result", e.Result)
+			// l.Debugw("saving - debug", "scriptExecution", scriptExecution)
 
-				scriptExecution.Execution = &e
-				return s.Repository.Update(ctx, scriptExecution)
-			})
-		}(scriptExecution, executor)
+			result := jobClient.LaunchK8sJob(scriptExecution.Id, getImageFromCRType(scriptCR.Spec.Type_), options)
```
exu, line 196: just thinking if we could go async here 
- we have pending `ABORT` feature - which will need to be implemented - probably we could run this job and keep some ID of it to eventually kill execution.

internal/app/api/v1/scripts.go
```diff
@@ -131,10 +132,10 @@ func (s kubtestAPI) ExecuteScript() fiber.Handler {
 		}
 
 		// get executor from kubernetes CRs
-		executor, err := s.Executors.Get(scriptCR.Spec.Type_)
-		if err != nil {
-			return s.Error(c, http.StatusInternalServerError, fmt.Errorf("can't get executor: %w", err))
-		}
+		// executor, err := s.Executors.Get(scriptCR.Spec.Type_)
```
exu, line 135: I would leave job based executor/runner to be registered as executor/runner ? What do you think ?

internal/app/api/v1/scripts.go
```diff
@@ -173,21 +175,34 @@ func (s kubtestAPI) ExecuteScript() fiber.Handler {
...
 
+		}
+		fmt.Println("*******ID:", scriptExecution.Id)
+		jobClient, err := jobs.NewJobClient()
+		if err != nil {
+			return s.Error(c, http.StatusInternalServerError, fmt.Errorf("can't get k8s client: %w", err))
+		}
 		// save watch result asynchronously
-		go func(scriptExecution kubtest.ScriptExecution, executor client.HTTPExecutorClient) {
+		go func(scriptExecution kubtest.ScriptExecution) error {
```
exu, line 189: This will not work as HTTP client, There is Watch method which handles status change

internal/app/api/v1/scripts.go
```diff
@@ -158,11 +159,12 @@ func (s kubtestAPI) ExecuteScript() fiber.Handler {
 			Params:     request.Params,
 		}
 		s.Log.Infow("calling executor with options", "options", options)
-		execution, err := executor.Execute(options)
+		// execution, err := executor.Execute(options)
```
exu, line 162: I would leave HTTP based executors and implement both methods - it's quite nice feature for possible integrations - not all integrations could be fat like Cypress.

pkg/jobs/jobclient.go
```diff
@@ -0,0 +1,190 @@
...
+		Pods(c.Namespace).
+		GetLogs(podName, &podLogOptions)
+	stream, err := podLogRequest.Stream(context.TODO())
+	if err != nil {
+		return "", err
+	}
+
+	defer stream.Close()
+
+	for {
```
exu, line 168: good place to make it async - we could think about setting raw output as some kind log - and append new job output lines to it
would be also great place for integration of getting real time log output.

exu: log lines could be passed as channel of string ?

internal/app/api/v1/scripts.go
jasmingacic: job's name is the same as execution ID so it shouldn't be a problem to do that

internal/app/api/v1/scripts.go
jasmingacic: Literally no need for it now, we could perhaps recycle it so we can get appropriate image for the pod

internal/app/api/v1/scripts.go
jasmingacic: yes i'm aware of that but overall process was quite different so we will have to make this better

internal/app/api/v1/scripts.go
jasmingacic: Sounds good we only need to figure out a way how to do that. Which also opens the question if we are going to need typed executors anymore?

pkg/jobs/jobclient.go
jasmingacic: Let's add more stuff to this.

jasmingacic:
Recent comments have reminded me that we need to figure out the way to "watch" pods the same way it was done by "watching" mongodb for changes

internal/app/api/v1/scripts.go
exu: > Recent comments have reminded me that we need to figure out the way to "watch" pods the same way it was done by "watching" mongodb for changes

even better - http watch simply was asking in loop for get and checked for status change (success|error)
here we can easily implement watch based on logs (just need to tag those lines which are logs somehow

internal/app/api/v1/scripts.go
nicufk: how will you know then which agent to run?

jasmingacic: Based on CR, we have field Type_ which should be enough to make the call.

exu (changes requested):

nicufk (changes requested):
For not spamming the slack channel and also could you please make separate commits after changes because without amend to ease tracking of changes.

pkg/api/client/direct.go
```diff
@@ -156,6 +156,23 @@ func (c DirectScriptsAPI) ListScripts(namespace string) (scripts kubtest.Scripts
 	return
 }
 
+func (c DirectScriptsAPI) AbortExecution(scriptID, id string) error {
+	fmt.Println("TADA")
+	uri := fmt.Sprintf(c.URI+"/v1/scripts/%s/executions/%s/abort", scriptID, id)
```
nicufk, line 161: Can we have it like a const? and v1 here should not be hardcoded.

pkg/api/client/proxy.go
```diff
@@ -159,6 +159,19 @@ func (c ProxyScriptsAPI) ListScripts(namespace string) (scripts kubtest.Scripts,
 	return c.getScriptsFromResponse(resp)
 }
 
+// GetExecutions list all executions in given script
+func (c ProxyScriptsAPI) AbortExecution(scriptID, id string) error {
+	uri := fmt.Sprintf("v1/scripts/%s/executions/%s/abort", scriptID, id)
```
nicufk, line 164: Same as previous. Can we have it like a const? and v1 here should not be hardcoded.

pkg/executor/client/rest.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+
+	}()
+
+	return events
+}
+
+func (c RestExecutorClient) Get(id string) (execution kubtest.Execution, err error) {
+	uri := fmt.Sprintf(c.URI+"/v1/executions/%s", id)
```
nicufk, line 61: Hardcoded v1.

pkg/executor/client/rest.go
```diff
@@ -0,0 +1,137 @@
...
+// Execute starts new external script execution, reads data and returns ID
+// Execution is started asynchronously client can check later for results with Get
+func (c RestExecutorClient) Execute(options ExecuteOptions) (execution kubtest.Execution, err error) {
+	request := MapExecutionOptionsToExecutionRequest(options)
+	body, err := json.Marshal(kubtest.ExecutionRequest(request))
+	if err != nil {
+		return execution, err
+	}
+
+	resp, err := c.client.Post(c.URI+"/v1/executions/", "application/json", bytes.NewReader(body))
```
nicufk, line 83: Hardcoded v1.

nicufk (changes requested):

exu (approved):
LGTM with changes from @nicufk passed as new task to review as it's a rabbit hole ;)

jasmingacic:
Opened issue #308 and merging this.

pkg/jobs/jobclient.go
```diff
@@ -0,0 +1,347 @@
...
+		case v1.ClaimBound:
+			return true, nil
+		case v1.ClaimLost:
+			return false, nil
+		}
+		return false, nil
+	}
+}
+
+func (c *JobClient) GetPodLogs(podName string, containerName string, endMessage string) (string, error) {
```
nicufk, line 232: We can structure the code to have the public methods first and then the private ones.

nicufk (approved):

pkg/jobs/jobclient.go
jasmingacic: This method will be used for watch so we need it to be public

pkg/jobs/jobclient.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	return &kubtest.ExecutionResult{
+		Status:    kubtest.ExecutionStatusSuceess,
+		RawOutput: result,
+	}, nil
+}
+
+// connectToK8s returns ClientSet
+func connectToK8s() (*kubernetes.Clientset, error) {
```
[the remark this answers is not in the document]
jasmingacic: I would move this for some other PR as this one has become huge

PR #330, opened 2021-09-27
removed persistent volume and pvc from k8s jobs

jasmingacic:
I yanked out PVs and PVCs from current implementation as we said that artifacts should be stored in min.io. 
So back to the drawing board


fog1985 (approved):
LGTM

nicufk (approved):

PR #351, opened 2021-10-01
added storage interface

jasmingacic:
Storage interface with implementation for minio storage provider


exu (commented):
Looks good for now :)

nicufk (approved):

jasmingacic:
as a proof of concept i used this to setup minio in a single pod https://faun.pub/minio-object-storage-deployment-on-kubernetes-83f81fba1d03

Also I built a file "scrapper" (leaving it here for posterity)
```
	cli, err := minio.NewClient("localhost:9000", "api", "minioadmin", "minioadmin", "minioadmin", false)

	err = cli.CreateBucket("test")
	if err != nil {
		fmt.Println(err)
		return
	}
	err = filepath.Walk(".",
		func(path string, info os.FileInfo, err error) error {
			if err != nil {
				return err
			}

			if !info.IsDir() {
				fmt.Println(path, info.Size())
				err = cli.SaveFile("test", path)
				if err != nil {
					fmt.Println(err)
				}
			}

			return nil
		})
	if err != nil {
		log.Println(err)
	}
```

PR #472, opened 2021-11-03
executions/{id}/artifacts endpoints list and get

jasmingacic:
For whichever reason testkube.yaml got changed from double quotes to single quotes but it isn't that relevant as the other changes are


internal/app/api/v1/scripts.go
```diff
@@ -424,3 +424,31 @@ func NewExecutionFromExecutionOptions(options client.ExecuteOptions) testkube.Ex
...
+	return func(c *fiber.Ctx) error {
+
+		executionID := c.Params("executionID")
+		fileName := c.Params("filename")
+
+		file, err := s.Storage.DownloadFile(executionID, fileName)
+		if err != nil {
+			return s.Error(c, http.StatusInternalServerError, err)
+		}
+		return c.JSON(file)
```
exu, line 452: Should stream content not output json as we discussed yesterday

exu (changes requested):

internal/app/api/v1/scripts.go
exu: Just was thinking how to return JSON metadata for single file maybe /executions/{id}/artifacts/{path}/metadata endpoint? or here pass JSON and stream on /executions/{id}/artifacts/{path}/stream?

api/v1/testkube.yaml
```diff
@@ -65,89 +65,155 @@ paths:
...
+          schema:
+            type: string
+          required: true
+          description: ID of the script execution
+      tags:
+        - artifacts
+        - executions
+        - api
+      summary: "Get execution's artifacts by ID"
+      description: "Returns arifacts of the given executionID"
```
olensmar, line 101: typo

api/v1/testkube.yaml
```diff
@@ -65,89 +65,155 @@ paths:
...
+      operationId: downloadFile
+      responses:
+        200:
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Artifact"
+        500:
+          description: "problem with getting script executions from storage"
```
olensmar, line 145: wrong wording?

api/v1/testkube.yaml
```diff
@@ -564,7 +627,21 @@ components:
...
+      type: object
+      description: API server artifact
+      properties:
+        name:
+          type: string # https://api-server/exec_id/subdirectory/file_name
+          description: artifact file path
+        size:
+          type: integer
+          description: file size in bytes
+        data:
```
olensmar, line 642: an alternative could be to not return the data inline but have a binary/octet-stream response in the operation - the client accept header would specify which to return!?

api/v1/testkube.yaml
```diff
@@ -65,89 +65,155 @@ paths:
...
+          schema:
+            type: string
+          required: true
+          description: name of the file in the bucket
+      operationId: downloadFile
+      responses:
+        200:
+          description: successful operation
+          content:
+            application/json:
```
olensmar, line 141: see comment further below - we could have a separate response with content type application/octet-stream - see https://stackoverflow.com/a/53339910/249414 for an example - the client would need to specify an accept header to control which response to return

jasmingacic:
@nicufk I merged your changes into this PR can you please review if i messed up testkube.yaml

api/v1/testkube.yaml
```diff
@@ -612,8 +676,18 @@ components:
...
+          $ref: "#/components/schemas/ExecutionResult"
 
+    Artifact:
+      type: object
+      description: API server artifact
+      properties:
+        name:
+          type: string # https://api-server/exec_id/subdirectory/file_name
+          description: artifact file path
+        size:
```
olensmar, line 688: can we provide some metadata on content-type? (we can always guess from the file extension) - this would help clients with further processing... i.e. if they know the file is a pdf, avi, png, etc...

jasmingacic: minio will return full name with extension

jasmingacic: Just like this
image

olensmar: so you're saying we should let the client decide based on extension only? I'm thinking that other executors might provide this metadata without the client having to guess... but we can perhaps cross that bridge when we get there....

jasmingacic: I tried to pull content-type (MIME) from minio but when listing objects from a bucket you don't get information.
Below is everything that minio returns about a single object:
```

(minio.ObjectInfo) {
 ETag: (string) (len=32) "7f932a200f44503cebd0c2ceeed8efb5",
 Key: (string) (len=7) "main.go",
 LastModified: (time.Time) 2021-11-04 10:32:32.845 +0000 UTC,
 Size: (int64) 471,
 ContentType: (string) "",
 Expires: (time.Time) 0001-01-01 00:00:00 +0000 UTC,
 Metadata: (http.Header) <nil>,
 UserMetadata: (minio.StringMap) <nil>,
 UserTags: (map[string]string) <nil>,
 UserTagCount: (int) 0,
 Owner: (minio.Owner) {
  XMLName: (xml.Name) {
   Space: (string) (len=39) "http://s3.amazonaws.com/doc/2006-03-01/",
   Local: (string) (len=5) "Owner"
  },
  DisplayName: (string) (len=64) "02d6176db174dc93cb1b899f7c6078f08654445fe8cf1b6ce98d8855f66bdbf4",
  ID: (string) (len=5) "minio"
 },
 Grant: ([]minio.Grant) <nil>,
 StorageClass: (string) (len=8) "STANDARD",
 IsLatest: (bool) false,
 IsDeleteMarker: (bool) false,
 VersionID: (string) "",
 ReplicationStatus: (string) "",
 Expiration: (time.Time) 0001-01-01 00:00:00 +0000 UTC,
 ExpirationRuleID: (string) "",
 Restore: (*minio.RestoreInfo)(<nil>),
 Err: (error) <nil>
}
```

exu (approved):

olensmar (approved):

PR #475, opened 2021-11-04
added segmentio telemetry

jasmingacic:
Added data anonymous data collection.  Basically it collects information about the host in form like this:
```
{
  "anonymousId": "a4652358effb311a074bf84d2aed5a7d270dee858bff10e847df2a9ea132bb38",
  "context": {
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    }
  },
  "event": "testkube-heartbeat",
  "integrations": {},
  "messageId": "2021-11-04 19:54:40.029549 +0100 CET m=+0.148209228",
  "originalTimestamp": "2021-11-04T19:54:40.029571+01:00",
  "receivedAt": "2021-11-04T18:54:41.004Z",
  "sentAt": "2021-11-04T18:54:40.029Z",
  "timestamp": "2021-11-04T18:54:41.004Z",
  "type": "track"
}
```

If we are going to proceed with this we need to make sure that LD_FLAG is passed 
```
LD_FLAGS += -X github.com/kubeshop/testkube/pkg/telemetry.telemetryToken=$(TELEMETRY_TOKEN)
```

Also `TESTKUBE_TELEMETRY_DISABLE` can be passed to the API server to disable data collection.

I've created integration with my personal account for testing purposes which I can share in slack

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -11,6 +11,7 @@ import (
 	v1API "github.com/kubeshop/testkube/internal/app/api/v1"
 	"github.com/kubeshop/testkube/internal/pkg/api/repository/result"
 	"github.com/kubeshop/testkube/internal/pkg/api/repository/storage"
+	"github.com/kubeshop/testkube/pkg/telemetry"
 	"github.com/kubeshop/testkube/pkg/ui"
 )
 
@@ -26,6 +27,10 @@ func init() {
 }
 
 func main() {
+
+	if _, telemetryNotEnabled := os.LookupEnv("TESTKUBE_TELEMETRY_DISABLED"); !telemetryNotEnabled {
+		telemetry.CollectAnonymousInfo()
+	}
 
 	port := os.Getenv("APISERVER_PORT")
 
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -28,9 +28,7 @@ func init() {
 
 func main() {
 
-	if _, telemetryNotEnabled := os.LookupEnv("TESTKUBE_TELEMETRY_DISABLED"); !telemetryNotEnabled {
-		telemetry.CollectAnonymousInfo()
-	}
+	telemetry.CollectAnonymousInfo()
 
 	port := os.Getenv("APISERVER_PORT")
 
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -2,7 +2,7 @@
 
 To improve the end-user experience TestKube collects anonymous data about usage and sends it to us.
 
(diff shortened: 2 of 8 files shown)


exu:
Ldflags are added only in makefile 
I think we have also some docker bake commands in GitHub action so probably it'll need to added there also ?

exu:
Do we want to embed key in binary ? I think it'll be quite easy to retrieve

exu:
Missing docs describing what kind of data are we sending and how to disable it on demand

jasmingacic:
Let's find the best options for deployment and I will add docs. Open questions are how the users will be able to opt out of telemetry.

Embedding the key into the app is the easiest way plus even if we exposed it no harm done. Also let's see what @fog1985 suggests on how to add that to GH actions

exu (changes requested):

cmd/api-server/main.go
```diff
@@ -27,6 +28,10 @@ func init() {
 
 func main() {
 
+	if _, telemetryNotEnabled := os.LookupEnv("TESTKUBE_TELEMETRY_DISABLED"); !telemetryNotEnabled {
```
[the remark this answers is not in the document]
jasmingacic: by default it is enabled, unless you specify otherwise
we want to get as much data as possible

exu (changes requested):
still missing docs

fog1985:
> Let's find the best options for deployment and I will add docs. Open questions are how the users will be able to opt out of telemetry.
> 
> Embedding the key into the app is the easiest way plus even if we exposed it no harm done. Also let's see what @fog1985 suggests on how to add that to GH actions

Hey @jasmingacic ,
Not sure what is meant to be done on GH's side. Please clarify so that I could respond. properly.

jasmingacic:
@fog1985 so when we bake our release we should inject [LD_FLAG](https://github.com/kubeshop/testkube/blob/telemetry/Makefile#L10) which is passed to build [process](https://github.com/kubeshop/testkube/blob/telemetry/Makefile#L25) the value of LD_FLAG should be stored as a secret somewhere in GH actions.

docs/telemetry.md
```diff
@@ -0,0 +1,26 @@
+# Telemetry
+
+To improve the end-user experience TestKube collects anonymous data about usage and sends it to us.
+
+Telemetry collects and scrambles infromation about the host when the API server is bootstrapped for the first time.
```
olensmar, line 5: ", allowing us to ... " ?

olensmar, line 5: typo - infromation

exu (approved):

PR #481, opened 2021-11-10
exposed CLI interface to fetch files (direct client)

jasmingacic:
Exposed CLI subcommand artifacts:

```
# go run ../cmd/kubectl-testkube/main.go  --client direct artifacts -e something13
parsing client version - : Invalid Semantic Version
                    NAME                   | SIZE (KB)  
+------------------------------------------+-----------+
  Health.postman_collection.json           |       775  
  Local-API-Health.postman_collection.json |       763  
  LocalHealth.postman_collection.json      |       996  
  LongRun.postman_collection.json          |      3508  
  TestKube-Sanity.postman_collection.json  |     14395  
  curl.json                                |       182  
  e2e_test.go                              |      3380  
  local-postman.yaml                       |       356  
  test.postman_collection.json             |      1253  
  testkube.go                              |      2457  
```

for now it only works with client=direct I still haven't implemented proxy client for downloading files. First I need to deploy the entire setup.

This PR also changes how file scrapping is done. Now the files are uploaded to minio by name not by full path. The reason for this was that the client was not able to get bucket's directory's files. 

This command will download file in the execution path of the  binary.
```
# go run ../cmd/kubectl-testkube/main.go  --client direct artifacts -e something13 -f curl.json
parsing client version - : Invalid Semantic Version
File testkube.go downloaded.
```

@exu @nicufk  Should we add another subcommand like download and allow users to provide destination directory for download?

Also our current deployment does not provide these env vars for REST API 
```
STORAGE_SSL
STORAGE_ENDPOINT
STORAGE_ACCESSKEYID
STORAGE_SECRETACCESSKEY
STORAGE_LOCATION
STORAGE_TOKEN
```
Similar env vars need to be passed down to job executors as shown here https://github.com/kubeshop/testkube-executor-cypress/blob/main/pkg/runner/cypress.go#L17-L24 

### DIFF
--- a/internal/app/api/v1/server.go
+++ b/internal/app/api/v1/server.go
@@ -73,6 +73,7 @@ func (s testkubeAPI) Init() {
 	executions.Get("/", s.ListExecutions())
 	executions.Get("/:executionID", s.GetExecution())
 	executions.Get("/:executionID/artifacts", s.ListArtifacts())
+	executions.Get("/:executionID/logs", s.ExecutionLogs())
 	executions.Get("/:executionID/artifacts/:filename", s.GetArtifact())
 
 	scripts := s.Routes.Group("/scripts")
--- a/internal/app/api/v1/server.go
+++ b/internal/app/api/v1/server.go
@@ -1,8 +1,6 @@
 package v1
 
 import (
-	"os"
-
 	"github.com/gofiber/fiber/v2/middleware/cors"
 	"github.com/kelseyhightower/envconfig"
 	executorscr "github.com/kubeshop/testkube-operator/client/executors"
@@ -46,12 +44,24 @@ type testkubeAPI struct {
 	ExecutorsClient *executorscr.ExecutorsClient
 	Metrics         Metrics
 	Storage         storage.Client
+	storageParams   storageParams
+}
+
+type storageParams struct {
+	SSL             bool
+	Endpoint        string
+	AccessKeyId     string
+	SecretAccessKey string
+	Location        string
+	Token           string
 }
 
 func (s testkubeAPI) Init() {
 	var err error
-	_, minioSSL := os.LookupEnv("MINIO_SSL")
-	s.Storage, err = minio.NewClient(os.Getenv("MINIO_STORAGE_ENDPOINT"), os.Getenv("MINIO_STORAGE_ACCESSKEYID"), os.Getenv("MINIO_STORAGE_SECRETACCESSKEY"), os.Getenv("MINIO_STORAGE_LOCATION"), os.Getenv("MINIO_STORAGE_TOKEN"), minioSSL)
+
+	envconfig.Process("STORAGE", &s.storageParams)
+
+	s.Storage, err = minio.NewClient(s.storageParams.Endpoint, s.storageParams.AccessKeyId, s.storageParams.SecretAccessKey, s.storageParams.Location, s.storageParams.Token, s.storageParams.SSL)
 	if err != nil {
 		s.Log.Warnf("error occured while instantiating storage provider:", err)
 	}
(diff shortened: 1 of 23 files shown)


nicufk (approved):
👍

exu:
Yes would be best to have list and download commands if it's quite easy to implement

jasmingacic:
> Yes would be best to have list and download commands if it's quite easy to implement

Done can you review again

pkg/api/v1/client/interface.go
```diff
@@ -29,6 +29,9 @@ type Client interface {
 	GetExecutor(name string) (executor testkube.ExecutorDetails, err error)
 	ListExecutors() (executors testkube.ExecutorsDetails, err error)
 	DeleteExecutor(name string) (err error)
+
+	GetExecutionArtifacts(executionID string) (artifacts testkube.Artifacts, err error)
+	DownloadFile(executionID, fileName, destination string) (artifact string, err error)
```
exu, line 34: io.Reader instead of string should be used

exu: (For return values )

pkg/api/v1/client/proxy.go
```diff
@@ -392,3 +397,52 @@ func (c ProxyScriptsAPI) makeDeleteRequest(uri string, isContentExpected bool) e
...
+
+	return c.getArtifactsFromResponse(resp)
+
+}
+func (c ProxyScriptsAPI) DownloadFile(executionID, fileName, destination string) (artifact string, err error) {
+	uri := c.getURI("/executions/%s/artifacts/%s", executionID, fileName)
+	req := c.GetProxy("GET").
+		Suffix(uri)
+
+	resp := req.Do(context.Background())
```
exu, line 419: Use Stream() instead. It'll return io.ReadCloser look at Logs() method for details

exu (changes requested):

internal/app/api/v1/server.go
```diff
@@ -51,7 +51,7 @@ type testkubeAPI struct {
 func (s testkubeAPI) Init() {
 	var err error
 	_, minioSSL := os.LookupEnv("MINIO_SSL")
```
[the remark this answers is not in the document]
jasmingacic: since it is a flag I just check if it is present or not and therefore i set it the value can be anything

pkg/api/v1/client/interface.go
jasmingacic: Dowload file already downloads the file and stores it on the client machine i thought no need to pass the io.Reader

pkg/api/v1/client/proxy.go
jasmingacic: the API endpoint is sending file as c.SendFile since c.SendStream was not working https://github.com/kubeshop/testkube/blob/cli_artifacts/internal/app/api/v1/scripts.go#L484

jasmingacic:
Most recent commit fixes issue with passing full file name and allows subdirectories in buckets

pkg/api/v1/client/interface.go
jasmingacic: The function returns only file name and in the   fuction i construct the file out of datastream

jasmingacic: Take another look please

pkg/api/v1/client/proxy.go
jasmingacic: fixed it

internal/app/api/v1/server.go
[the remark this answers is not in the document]
jasmingacic: i think we can live with this

jasmingacic:
Added corresponding env vars to helm charts as well https://github.com/kubeshop/helm-charts/pull/62

nicufk (approved):

PR #528, opened 2021-11-18
fix: Error reporting for artifacts.

### LINKED ISSUE #518: Strange doubled message when listing artifacts without id passed
**Describe the bug**

➜  testkube git:(main) t artifacts list
Error: execution-id is a required parameter
execution-id is a required parameter

Use common ui framework for messages and infos

jasmingacic:
Strange doubled message when listing artifacts without id passed #518

Signed-off-by: jasmingacic <jasmin.gacic@gmail.com>

### DIFF
--- a/internal/app/api/v1/scripts.go
+++ b/internal/app/api/v1/scripts.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
 
 	"github.com/gofiber/fiber/v2"
@@ -443,7 +444,9 @@ func (s testkubeAPI) ListArtifacts() fiber.Handler {
 	return func(c *fiber.Ctx) error {
 
 		executionID := c.Params("executionID")
-
+		if s.Storage == nil || reflect.ValueOf(s.Storage).IsNil() {
+			return s.Error(c, http.StatusInternalServerError, fmt.Errorf("Storage interface not configured."))
+		}
 		files, err := s.Storage.ListFiles(executionID)
 		if err != nil {
 			return s.Error(c, http.StatusInternalServerError, err)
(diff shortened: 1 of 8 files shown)


exu:
Please add PR title (ideally with conventional commit fix: prefix )

internal/app/api/v1/scripts.go
```diff
@@ -443,7 +444,9 @@ func (s testkubeAPI) ListArtifacts() fiber.Handler {
 	return func(c *fiber.Ctx) error {
 
 		executionID := c.Params("executionID")
-
+		if s.Storage == nil || reflect.ValueOf(s.Storage).IsNil() {
```
exu, line 447: Reflection looks odd

nicufk: Agree. Can we find a more elegant way?

jasmingacic: it is either of these options
https://play.golang.org/p/QlObAWkiVdB
i could cast s.Storage.(*minio.Client) == nil

pkg/api/v1/client/proxy.go
```diff
@@ -443,12 +443,12 @@ func (c ProxyScriptsAPI) DownloadFile(executionID, fileName, destination string)
 		Suffix(uri).
 		SetHeader("Accept", "text/event-stream").
 		Stream(context.Background())
```
jasmingacic, line 445: @exu how do you check for status code in such calls?

internal/app/api/v1/scripts.go
jasmingacic: addressed it

pkg/api/v1/client/proxy.go
nicufk: @jasmingacic look into makeDeleteRequest, something like that
`		var code int
		resp.StatusCode(&code)
		if code != http.StatusNoContent {
			respBody, err := resp.Raw()
			if err != nil {
				return err
			}
			return fmt.Errorf("request returned error: %s", respBody)
		}`

exu (approved):
LGTM

pkg/api/v1/client/proxy.go
jasmingacic: I would if i had any response here.
Let's merge it and then we deal with it later

PR #544, opened 2021-11-19
fix: download artifacts in cli

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/api/v1/client/proxy.go
```diff
@@ -437,8 +437,9 @@ func (c ProxyScriptsAPI) GetExecutionArtifacts(executionID string) (artifacts te
 	return c.getArtifactsFromResponse(resp)
 
 }
+
 func (c ProxyScriptsAPI) DownloadFile(executionID, fileName, destination string) (artifact string, err error) {
-	uri := c.getURI("/executions/%s/artifacts/%s", executionID, fileName)
+	uri := c.getURI("/executions/%s/artifacts/%s", executionID, url.QueryEscape(fileName))
```
nicufk, line 442: I think it is better to use .Param() for the query params in the proxy.

jasmingacic: let's revisit this later

jasmingacic (approved):

PR #631, opened 2021-12-09
feat: openapi spec for test executions

exu:

- [x] tested locally
- [ ] tested on cluster
- [x] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


api/v1/testkube.yaml
```diff
@@ -649,6 +718,129 @@ paths:
...
+      properties:
+        namespace:
+          type: string
+        name:
+          type: string
+
+
+    Test:
+      type: object
+      properties:
```
olensmar, line 733: perhaps we should have a description property - testers like to describe what their tests do...

api/v1/testkube.yaml
```diff
@@ -649,6 +718,129 @@ paths:
...
+        - type: object
+          required:
+            - namepace 
+            - name
+          properties:
+            namespace:
+              type: string
+            name:
+              type: string
+            stopOnFailure:
```
olensmar, line 770: "stopTestOnFailure"?

api/v1/testkube.yaml
```diff
@@ -649,6 +718,129 @@ paths:
...
+              type: boolean
+              default: true
+
+    TestStepDelay: 
+      allOf:     # Combines the TestStep and the inline model
+        - $ref: '#/components/schemas/TestStep'
+        - type: object
+          required:
+            - duration
```
olensmar, line 779: delay ?

api/v1/testkube.yaml
```diff
@@ -795,6 +987,8 @@ components:
         - success
         - error
 
+    
+
     ExecutionResult:
```
olensmar, line 992: perhaps we should rename this to ScriptExecutionResult - to avoid confusion with TestExecutionResult ?

api/v1/testkube.yaml
```diff
@@ -649,6 +718,129 @@ paths:
...
+            prefix: "some-"
+        startTime:
+          type: string
+          description: 'test start time'
+          format: date-time
+        endTime:
+          type: string
+          description: 'test end time'
+          format: date-time
+        executionsResults:
```
olensmar, line 822: stepExecutionResults?  - all Steps should have a StepExecutionResult, some simpler than others?

api/v1/testkube.yaml
exu: Will script/test be conifusing for testers? Aren't they used to test/suite naming ?

exu: but on the bottom there is execution result which is the same (just going from different direction) 
from test we need to go by 2 levels, from script we're getting it directly

api/v1/testkube.yaml
exu: maybe stepResult - to not make those fields too hard to read for user ? @olensmar

olensmar (approved):

api/v1/testkube.yaml
```diff
@@ -649,6 +718,137 @@ paths:
...
+        - type: object
+          required:
+            - namepace 
+            - name
+          properties:
+            namespace:
+              type: string
+            name:
+              type: string
+            stopTestOnFailure:
```
olensmar, line 778: should this be in TestStep object?

jasmingacic (approved):

api/v1/testkube.yaml
```diff
@@ -649,6 +718,137 @@ paths:
...
+          format: date-time
+        stepResults:
+          type: array
+          description: "steps execution restults"
+          items:
+            $ref: "#/components/schemas/TestExecutionResult"
+            description: test execution results
+          
+
+    TestExecutionResult:
```
olensmar, line 838: perhaps "TestStepExecutionResult" is a better name?

api/v1/testkube.yaml
exu: I assume delay don't have failures :) what can go wrong with delay :D ??

olensmar: sure - just trying to avoid step-specific processing as most other step types will probably be able to fail - the Delay step will probably always succeed (unless the test is aborted while delaying?)

PR #814, opened 2022-01-20
fix: allow to delete tests

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/tests/deleteall.go
+++ b/cmd/kubectl-testkube/commands/tests/deleteall.go
@@ -1,0 +1,29 @@
+package tests
+
+import (
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common/validator"
+	"github.com/kubeshop/testkube/pkg/ui"
+	"github.com/spf13/cobra"
+)
+
+func NewDeleteTestsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete <testName>",
+		Short: "Delete tests",
+		Long:  `Delete tests by name`,
+		Args:  validator.TestName,
+		Run: func(cmd *cobra.Command, args []string) {
+			ui.Logo()
+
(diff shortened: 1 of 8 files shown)


cmd/kubectl-testkube/commands/tests/deleteall.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/kubeshop/testkube/pkg/ui"
+	"github.com/spf13/cobra"
+)
+
+func NewDeleteTestsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete <testName>",
+		Short: "Delete tests",
+		Long:  `Delete tests by name`,
+		Args:  validator.TestName,
```
vsukhin, line 14: Should be changed for all tests, we don't need name then

pkg/api/v1/client/direct.go
```diff
@@ -553,6 +553,25 @@ func (c DirectScriptsAPI) DeleteTest(name, namespace string) (err error) {
...
+		return fmt.Errorf("prepare request error: %w", err)
+	}
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("client.Do error: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if err := c.responseError(resp); err != nil {
+		return fmt.Errorf("api/delete-test returned error: %w", err)
```
vsukhin, line 569: delete-tests?

vsukhin (commented):
couple minor notes

cmd/kubectl-testkube/commands/tests/deleteall.go
exu: good catch! fixed

pkg/api/v1/client/direct.go
exu: good catch! fixed

vsukhin (approved):

PR #890, opened 2022-01-28
feat: common script source for git ops

exu:
## Pull request description 

Changed data structure for content - Needs new operator, and executors to follow up

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/executor/content/fetcher.go
```diff
@@ -0,0 +1,112 @@
...
+	case testkube.ScriptContentTypeGitDir:
+		return f.FetchGitDir(content.Repository)
+	default:
+		return path, fmt.Errorf("unhandled content type: '%s'", content.Type_)
+	}
+}
+
+// FetchString stores string content as file
+func (f Fetcher) FetchString(str string) (path string, err error) {
+	return f.saveTempFile(strings.NewReader(str))
```
vsukhin, line 41: Do we clean up this temp file somewhere?

pkg/executor/content/fetcher.go
```diff
@@ -0,0 +1,112 @@
...
+
+		gitURI.User = url.UserPassword(repo.Username, repo.Token)
+		return gitURI.String(), nil
+	}
+
+	return repo.Uri, nil
+}
+
+func (f Fetcher) saveTempFile(reader io.Reader) (path string, err error) {
+	tmpFile, err := ioutil.TempFile("", "fetcher-save-temp-file")
```
vsukhin, line 102: TempFile creates a new temporary file in the directory dir, opens the file for reading and writing, and returns the resulting *os.File

vsukhin (commented):
a few questions

vsukhin:
Will also need to update testkube operator with latest changes to Script model

pkg/executor/content/fetcher.go
exu: hmm, it's a very good question, probably it'll be wiped out when the job will end - as we don't store anything in any volumes (we've dropped volumes someday in past

pkg/executor/content/fetcher.go
```diff
@@ -0,0 +1,112 @@
...
+
+// FetchGitDir returns path to locally checked out git repo with partial path
+func (f Fetcher) FetchGitDir(repo *testkube.Repository) (path string, err error) {
+	uri, err := f.gitURI(repo)
+	if err != nil {
+		return path, err
+	}
+
+	// if . make full repo checkout
+	if repo.Path == "." {
```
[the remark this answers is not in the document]
exu: hmm good point - will be remove :)

pkg/executor/content/fetcher.go
exu: good catch - some copy-paste :O

pkg/executor/client/common.go
```diff
@@ -4,7 +4,8 @@ import (
 	"time"
 
 	executorv1 "github.com/kubeshop/testkube-operator/apis/executor/v1"
-	scriptv1 "github.com/kubeshop/testkube-operator/apis/script/v1"
+	// scriptv1 "github.com/kubeshop/testkube-operator/apis/script/v1"
```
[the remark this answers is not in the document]
exu: done

pkg/executor/content/fetcher.go
vsukhin: If we fetch data only in executor job, then yes, becuase it will be stored on container disk till container is alive

vsukhin (approved):

PR #1032, opened 2022-02-28
feat: add ga analytics sending to the cmd and heartbeat

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/analytics/analytics.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+
+	v1 "github.com/mjpitz/go-ga/client/v1"
+	"github.com/mjpitz/go-ga/client/v1/gatypes"
+
+	"github.com/kubeshop/testkube/cmd/tools/commands"
+	"github.com/kubeshop/testkube/pkg/telemetry"
+)
+
+const testkubeTrackingID = "UA-221444687-1"
```
vsukhin, line 13: Might be to pass it from env var?

vsukhin (commented):
one question

cmd/api-server/main.go
```diff
@@ -58,6 +60,7 @@ func runMigrations() (err error) {
 func main() {
 
 	telemetry.CollectAnonymousInfo()
```
exu, line 60: Shouldn't we remove this one ?

pkg/analytics/analytics.go
exu: But it would be quite hard to force user to set it :)

pkg/telemetry/telemetry.go
```diff
@@ -58,7 +58,7 @@ func CollectAnonymousCmdInfo() {
 	}
 }
 
-func machineID() string {
+func MachineID() string {
 	id, _ := generate()
```
exu, line 62: Could you check this one - it's generating id from machine name only so probably we'll have a lot of MacBook pro

cmd/api-server/main.go
nicufk: Done ✅

pkg/analytics/analytics.go
nicufk: What will be the best solution?

vsukhin: If I remember it right, we added config for kubectl plugin, so it can be store and read it from there for cli
For server side, we should be able to pass it through dashboard by api method call or siimilar from cli api call. The question is where to store it on server side and to keep it there, even if pod is destroyed

vsukhin (commented):
one comment

pkg/analytics/analytics.go
exu: Yeah will be a good place 
Someone could Use it's company if if he wants ;)

exu (changes requested):
Use config struct to populate analytics Id there is default config constructor for it and will be saved on first run

nicufk:
> Use config struct to populate analytics Id there is default config constructor for it and will be saved on first run

Done, passing it on build, do we need users to be able to modify it?

vsukhin:
Think so

vsukhin (commented):
looks ok, except changing the key

exu:
> > Use config struct to populate analytics Id there is default config constructor for it and will be saved on first run
> 
> Done, passing it on build, do we need users to be able to modify it?

My comment was tin context of saving it in config - if we don't do this - I would leave it it's not in scope IMO

exu (approved):

PR #1034, opened 2022-03-01
feat: webhooks 

### LINKED ISSUE #965: Webhooks - in API

As an integrator, I want to get feedback from Testkube when the test is finished

**Additional context**

Implement webhook mechanism, which will take all URIs and call each of them which have particular event defined

So `TestkubeEvent` CRD
should have fields:
```
uri: string 
events: []string  
```

we'll integrate notifier after test completion 
webhook manager should call uris asynchronously and send particular events 
for now it'll be TestCompletedTestkubeEvent openapi spec model
with subset of fields from TestExecution (like ID, status, testname etc) 

client should call /executions/ID to get more details

webhook manager will list all TestkubeEvents  filtered by event 
then will call them on top of some work queue. 

should be async. 

blocked by #966

### LINKED ISSUE #966: Webhook CRD

So `Webhook` CRD should have fields:
```
uri: string 
events: []string  **Additional context**
```

Add any other context, CLI input/output, etc. about the enhancement here.

Add CRD + client in operator  to use in API

exu:
## Pull request description 

closes #965 
closes #966 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


internal/app/api/v1/executions.go
```diff
@@ -114,10 +115,29 @@ func (s TestkubeAPI) executeTest(ctx context.Context, options client.ExecuteOpti
...
 
+func (s TestkubeAPI) sendEvent(eventType *testkube.WebhookEventType, execution testkube.Execution) error {
+	webhookList, err := s.WebhooksClient.GetByEvent(eventType.String())
+	if err != nil {
+		return err
+	}
+
+	for _, wh := range webhookList.Items {
+		s.Log.Debugw("Sending event", "uri", wh.Spec.Uri, "type", eventType, "execution", execution)
+		s.EventsEmitter.Send(testkube.WebhookEvent{
```
vsukhin, line 131: Is it always successful?

internal/app/api/v1/webhook.go
```diff
@@ -0,0 +1,75 @@
...
+	return func(c *fiber.Ctx) error {
+		name := c.Params("name")
+		ns := c.Query("namespace", "testkube")
+
+		err := s.WebhooksClient.Delete(name, ns)
+		if err != nil {
+			return s.Error(c, http.StatusBadRequest, err)
+		}
+
+		c.Context().SetStatusCode(204)
```
vsukhin, line 72: Why is different from above c.Status? Also I would use http.Status

pkg/api/v1/client/proxy.go
```diff
@@ -314,6 +314,63 @@ func (c ProxyAPIClient) DeleteExecutor(name string) (err error) {
...
+
+	request := testkube.WebhookCreateRequest(options)
+
+	body, err := json.Marshal(request)
+	if err != nil {
+		return executor, err
+	}
+
+	req := c.GetProxy("POST").Suffix(uri).Body(body)
+	resp := req.Do(context.Background())
```
vsukhin, line 330: So the difference with direct is only in above 2 lines, guess, it can be abstracted into some method with options returing response

internal/app/api/v1/webhook.go
exu: copy paste from executor ;) done in both

pkg/api/v1/client/proxy.go
exu: not only this two lines - also getting response is slightly different

nicufk (approved):
LGTM, Awesome work!

vsukhin (approved):

PR #1063, opened 2022-03-06
fix: remove non anonymous data and send data on post run root cmd

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -60,7 +60,7 @@ func SendAnonymouscmdInfo() {
 			ApplicationName:    "testkube",
 			ApplicationVersion: commands.Version,
 		},
-		CustomDimensions: gatypes.StringList(os.Args),
+		CustomDimensions: gatypes.StringList(os.Args[1:]),
 	}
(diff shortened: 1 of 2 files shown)


exu (approved):

pkg/analytics/analytics.go
```diff
@@ -60,7 +60,7 @@ func SendAnonymouscmdInfo() {
 			ApplicationName:    "testkube",
 			ApplicationVersion: commands.Version,
 		},
-		CustomDimensions: gatypes.StringList(os.Args),
+		CustomDimensions: gatypes.StringList(os.Args[1:]),
```
vsukhin, line 63: Do you check, that slice has more than 1 element?

vsukhin (commented):
one question

pkg/analytics/analytics.go
nicufk: Done ✅

PR #1070, opened 2022-03-09
feat: removed direct client

exu:
## Pull request description 

From now we'll use a single API client which is Kube Proxy based 
to test locally we need to run local kube-proxy which pass all traffic to local api-server instance 

```sh 
make api-proxy
```

## Checklist (choose what happened)

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/api/v1/client/factory.go
```diff
@@ -10,19 +12,19 @@ const (
...
-			return client, err
-		}
-		client = NewProxyAPIClient(clientset, NewProxyConfig(namespace))
-	default:
-		err = fmt.Errorf("Client %s is not handled by testkube, use one of: %v", clientType, []ClientType{ClientDirect, ClientProxy})
+	var overrideHost string
+	var clientset kubernetes.Interface
+
+	if clientType == ClientDirect {
+		overrideHost = "http://127.0.0.1:8080"
```
vsukhin, line 19: I would make it a const or env var

vsukhin (approved):
nice stuff! I like code to be rmoved)

pkg/api/v1/client/factory.go
exu: done (added possibility to override by env var

PR #1096, opened 2022-03-16
docs: Update Telemetry Doc

jfermi:
Updated telemetry.md and corrected broken image links in tests-creating.md.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


docs/telemetry.md
```diff
@@ -25,3 +29,25 @@ The collected data looks like this.
   "type": "track"
 }
 ```
+
+## **What We Collect**
+
+Analytics tracked include:
+- The unique id generated from the MAC address.
+- The testkube version.
+- The command being executed.
```
exu, line 38: Maybe we could point out that we don't track any custom parameters and flags passed to a command by user - only top level testkube commands for general stats ?

exu: @jfermi ^^

nicufk: @exu, in the details field  we are tracking the whole command

jfermi: So it sounds like we can't say "we don't track customer parameters and flags passed" right now?

docs/tests-creating.md
```diff
@@ -21,15 +21,15 @@ To create your first Postman collection in Testkube, export your collection into
 
 Right click on your collection name:
 
-![create postman collection step 1](img/test-create-1.png)
+![create postman collection step 1](img/script-create-1.png)
```
vsukhin, line 24: We don't have `script` anymore in our terms. Either it should be new test-something image or rename exsiting image, if it's already with tests

jfermi: I can rename the image file.  I left them the same in case they were used by something I was not aware of.

jfermi: I changed the image file names to test-create... and updated tests-creating.md.

vsukhin (approved):

PR #1098, opened 2022-03-16
feat: add labels to test suite executions

vsukhin:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/api/v1/testkube.yaml
+++ b/api/v1/testkube.yaml
@@ -133,6 +133,7 @@ paths:
         - $ref: "#/components/parameters/ExecutionsStatusFilter"
         - $ref: "#/components/parameters/StartDateFilter"
         - $ref: "#/components/parameters/EndDateFilter"
+        - $ref: "#/components/parameters/Selector"
       responses:
         200:
           description: successful operation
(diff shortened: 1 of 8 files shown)


exu (approved):

cmd/kubectl-testkube/commands/tests/run.go
```diff
@@ -59,21 +59,26 @@ func NewRunTestCmd() *cobra.Command {
 
 			printExecutionDetails(execution)
 
-			if watchEnabled {
-				watchLogs(execution.Id, client)
-			}
+			if execution.Id != "" {
```
exu, line 62: I think it's an error when there is no execution ID - we shouldn't be silent here.

exu (approved):
👍🏻

exu (approved):

cmd/kubectl-testkube/commands/tests/run.go
vsukhin: It works differently for scheduled tests. We don't have execution id after starting scheduled test. It will be just planned for execution and id will be created when scheduler triggers this test

PR #1403, opened 2022-04-27
fix: sync openapi spec

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/kubectl-testkube/commands/webhooks/create.go
```diff
@@ -52,6 +54,7 @@ func NewCreateWebhookCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&name, "name", "n", "", "unique webhook name - mandatory")
 	cmd.Flags().StringArrayVarP(&events, "events", "e", []string{}, "event types handled by executor e.g. start-test|end-test")
 	cmd.Flags().StringVarP(&uri, "uri", "u", "", "URI which should be called when given event occurs")
+	cmd.Flags().StringToStringVarP(&labels, "label", "l", nil, "label key value pair: --label key1=value1")
```
nicufk, line 57: should we be consistent when naming this (selectors and labels)?

internal/app/api/v1/webhook.go
```diff
@@ -72,3 +72,15 @@ func (s TestkubeAPI) DeleteWebhookHandler() fiber.Handler {
...
 }
+
+func (s TestkubeAPI) DeleteWebhooksHandler() fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		err := s.WebhooksClient.DeleteByLabels(c.Query("selector"))
+		if err != nil {
+			return s.Error(c, http.StatusBadRequest, err)
+		}
+
+		c.Status(204)
```
nicufk, line 83:
```suggestion
		c.Status(http.StatusNoContent)
```

nicufk (approved):
LGTM, shall we update documentation as well?

exu (approved):

cmd/kubectl-testkube/commands/webhooks/create.go
exu: yeah internally those are selectors (AFAIK) but from user point of view we want to sell them as labels :)

vsukhin: So, generally selector is an expression on labels, and for any Get, Delete, Run we use internally (in api) native k8s selectors(in api you can pass "key1 in (value1, value2)") , for Create and Update we simply use labels, but for cli users as @exu mentioned for simplicity, it's always labels

Issue #1394, opened 2022-04-26
Improve Slack integration


nicufk:
Register a testkube-bot to the slack app store and improve user experience on integration with the testkube

donnalux:
![Testkube_icon_slack](https://user-images.githubusercontent.com/102221761/167620237-77307c09-238e-4afa-95d4-c684baeef222.svg)

Let me know if you need a different format or size. From what I can see 36x36px is the size slack apps use.

donnalux:
[Slack_Payload.txt](https://github.com/kubeshop/testkube/files/8661239/Slack_Payload.txt)

PR #1513, opened 2022-05-11
feat: Add slack ouath endpoint api-server

nicufk:
## Pull request description 
#1394 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


.github/workflows/release.yaml
```diff
@@ -128,6 +128,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.CI_BOT_TOKEN }}
           ANALYTICS_TRACKING_ID: "${{secrets.TESTKUBE_CLI_GA_MEASUREMENT_ID}}"
           ANALYTICS_API_KEY: "${{secrets.TESTKUBE_CLI_GA_MEASUREMENT_SECRET}}"
+          SLACK_BOT_CLIENT_ID: "${{secrets.TESTKUBE_SLACK_BOT_CLIENT_ID}}"
+          SLACK_BOT_CLIENT_SECRET: "${{secrets.TESTKUBE_SLACK_BOT_CLIENT_SECRET}}"
```
vsukhin, line 132: Hope, this is not conflicting with @ypoplavs changes

pkg/slacknotifier/slack_oauth.go
```diff
@@ -0,0 +1,52 @@
...
+
+// OauthHandler creates a handler for slack authentication
+func OauthHandler() fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		ctx := c.Context()
+
+		errStr := c.Params("error")
+		if errStr != "" {
+			c.Status(http.StatusUnauthorized)
+			_, err := c.WriteString(errStr)
```
vsukhin, line 34: Should we return original errStr in such case or writing error?

vsukhin (commented):
a few questions

.github/workflows/release.yaml
exu: It shouldn't they are already merged

build/api-server/Dockerfile
```diff
@@ -9,8 +9,18 @@ ARG VERSION
...
+ARG SLACK_BOT_CLIENT_SECRET
 
-RUN cd cmd/api-server;go build -ldflags "-X github.com/kubeshop/testkube/internal/pkg/api.Version=$VERSION -X github.com/kubeshop/testkube/internal/pkg/api.Commit=$COMMIT -X github.com/kubeshop/testkube/pkg/analytics.TestkubeMeasurementID=$ANALYTICS_TRACKING_ID -X github.com/kubeshop/testkube/pkg/analytics.TestkubeMeasurementSecret=$ANALYTICS_API_KEY" -o /app -mod mod -a .
+RUN cd cmd/api-server; \
+    go build \
+        -ldflags "-X github.com/kubeshop/testkube/internal/pkg/api.Version=$VERSION \
+            -X github.com/kubeshop/testkube/internal/pkg/api.Commit=$COMMIT \
+            -X github.com/kubeshop/testkube/pkg/analytics.TestkubeMeasurementID=$ANALYTICS_TRACKING_ID \
+            -X github.com/kubeshop/testkube/pkg/analytics.TestkubeMeasurementSecret=$ANALYTICS_API_KEY \
+			-X github.com/kubeshop/testkube/pkg/slacknotifier.SlackBotClientID=$SLACK_BOT_CLIENT_ID \
```
exu, line 21: Do we really need to embed slack into binary? It need to change for each client?

.github/workflows/release.yaml
nicufk: No conflicts, I rebased the branch with the latest changes 😉

internal/app/api/v1/server.go
```diff
@@ -253,6 +254,9 @@ func (s TestkubeAPI) Init() {
 	labels := s.Routes.Group("/labels")
 	labels.Get("/", s.ListLabelsHandler())
 
+	slack := s.Routes.Group("/slack")
```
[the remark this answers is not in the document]
nicufk: Yes, it will be accessed by the slack, maybe it is better to have it in a separate service.

pkg/slacknotifier/slack_oauth.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/gofiber/fiber/v2"
+	"golang.org/x/oauth2"
+)
+
+var (
+	SlackBotClientID     = ""
+	SlackBotClientSecret = ""
+	oauthConfig          = &oauth2.Config{
+		ClientID:     SlackBotClientID,
```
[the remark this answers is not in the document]
nicufk: this endpoint will be the redirect url, we will need only one instance, maybe it is not a good place in the api-server, shall we make a new service for that?

build/api-server/Dockerfile
nicufk: we need only one instance because slack will access it to validate the bot, it is not needed for every instance.

pkg/slacknotifier/slack_oauth.go
nicufk: yes, if something happens from the slack side we should inform user.

vsukhin: I meant, this method will return nil usually in case of non empty errStr, because c.WriteString will be succesful

pkg/slacknotifier/slack_oauth.go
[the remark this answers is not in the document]
vsukhin: At least we need some docs about it

pkg/slacknotifier/slack_oauth.go
nicufk: yes, you are right, good catch 👍

nicufk: actually it is the error of the handler func, we care about the request status here, so it should be fine

vsukhin (approved):

PR #1547, opened 2022-05-19
fix: get logs from storage

### LINKED ISSUE #1414: Testkube logs should not ask for real pod log in case of status failed or completed
**Describe the bug**

Logs should be taken from pod only if status is "running"
If status is not "running" just get log output from storage

image

vLia:
## Pull request description 
Closes #1414

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


internal/app/api/v1/executions.go
```diff
@@ -509,6 +501,55 @@ func (s TestkubeAPI) GetExecuteOptions(namespace, id string, request testkube.Ex
 	}, nil
 }
 
+// streamLogsFromStorage writes logs from the output of executionResult to the writer
```
exu, line 504: Why from storage? is there any database used here? As name suggests that it's fetched from some "storage"

internal/app/api/v1/executions.go
```diff
@@ -359,35 +360,26 @@ func (s TestkubeAPI) ExecutionLogsHandler() fiber.Handler {
...
 				return
 			}
 
-			// loop through pods log lines - it's blocking channel
-			// and pass single log output as sse data chunk
-			for out := range logs {
-				s.Log.Debugw("got log", "out", out)
-				fmt.Fprintf(w, "data: ")
-				err = enc.Encode(out)
+			if isExecutionFinished(execution) {
```
exu, line 372: ExecutionResults has this method already

```
func (e *ExecutionResult) IsCompleted() bool {
```

internal/app/api/v1/executions.go
vLia: Storage as opposed to K8s job logs. It makes more sense in the method that it is called from, but there is not direct connection to the storage in this one, only the ExecutionResult that's already queried from there.

would streamFinishedExecutionLogs() be a better name for this?

internal/app/api/v1/executions.go
vLia: As I was testing it with the`test-api-local` tests, I saw that there is 2 more execution status `testkube.SUCCESS_Status` and `testkube.ERROR__Status`, so I checked for those as well in the `isExecutionFinished(...)` function. Tbh I was a bit confused as to why there are 2 different ways to describe the status, but wanted to make sure we catch both

exu: Holy moly! Good catch - we have file which is not used at all - some leftover after messing with moving status field to 

Please remove this file as it's not used and don't need to be handled anywhere. 

```
pkg/api/v1/testkube/model_status.go
```

And `execution.ExecutionResult.IsCompleted()` will be enough then

internal/app/api/v1/executions.go
exu: It streams execution result if I understand correctly what it does?

vLia: Yes, it streams the result from the ResultRepository

internal/app/api/v1/executions.go
vLia: Awesome, will do so!

vLia: Do you know how we can check if there are more files like this?

exu: probably linter could partially do that - detect unused vars

internal/app/api/v1/executions.go
exu: this function just streams ExecutionResult (not from repository as it's not fetching it) executionResult is passed from outside wherever it'll be so context of this function is not related to repository at all.

vLia: fair point, updated it to `streamLogsFromResult(...)`

exu (approved):

PR #1577, opened 2022-05-24
docs: add Custom Resources to the docs

vLia:
## Pull request description 

Add documentation on Custom Resources used in Testkube

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/docs/development-crds.md
+++ b/docs/development-crds.md
@@ -66,10 +66,10 @@ $ kubectl testkube get executors -o yaml | grep -A1 types
 
 When creating a Testkube Test, there are multiple supported input types:
 
-* string
+* String
 * Git directory
 * Git file
-* file URI
+* File URI
 
 Variables can be configured using the `variables` field as shown below.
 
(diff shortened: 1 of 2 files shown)


docs/development-crds.md
```diff
@@ -0,0 +1,166 @@
...
+    types:
+    - postman/collection
+--
+    types:
+    - soapui/xml
+```
+
+When creating a Testkube Test, there are multiple supported input types:
+
+* string
```
exu, line 69: Upper case to be consistent with rest

docs/development-crds.md
```diff
@@ -0,0 +1,166 @@
...
+  - git-file
+  - git-dir
+  features: 
+  - artifacts
+  - junit-report
+```
+
+## Webhooks
+
+Testkube Webhooks are event-based configurable scripts. They are executed when a test is either started or finished. This can be defined under `events`.
```
exu, line 153: I would not describe them as scripts as it could suggest that user can do anything with them 

Currently it's only standard web hook so only http post call with execution object as payload with current state of execution

exu (changes requested):

docs/development-crds.md
vLia: Updated this as well

nicufk (approved):
LGTM

exu (approved):

PR #1657, opened 2022-06-15
build: Trigger argocd image

ypoplavs:
## Pull request description 
Trigger build of ArgoCD Docker image in `kubeshop/testkube-argocd` repo, when a new release was created, passing a tag version.

- [ ] tested locally
- [ ] tested in CI
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,30 +3,30 @@ name: Release kubectl-testkube
 on:
   push:
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+-test"
-    branches:
-      - trigger-argocd-image
+      - "v[0-9]+.[0-9]+.[0-9]+-"
+#    branches:
+#      - trigger-argocd-image
 
 jobs:
-  # release_creation:
-  #   name: Prepares Release on GitHub
-  #   runs-on: ubuntu-latest
-  #   outputs:
-  #     id: ${{ steps.main_step.outputs.id }}
-  #   steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -6,25 +6,25 @@ on:
...
+#         env:
+#           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+#         with:
+#           tag_name: ${{ github.ref }}
+#           release_name: testkube CLI ${{ github.ref }}
+#           body: |
+#             Changes in this Release
+#             - Releasing new version of the testkube CLI for Kubectl.
+#           draft: false
+#           prerelease: false
```
vLia, line 27: Is there a reason to keep all these commented out lines? I know it's not part of your code, was just wondering

vLia (approved):
Looks good!

.github/workflows/release.yaml
ypoplavs: Frankly, I don't think we need it, since the pipeline is working and doesn't require any changes. I can remove it if no one else would mind.

vLia: Just remove it, it'll still be there in the Git history

vsukhin (approved):
Looks good

nicufk (approved):

PR #1660, opened 2022-06-15
feat: add execution number to executions

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/internal/app/api/v1/executions.go
+++ b/internal/app/api/v1/executions.go
@@ -147,6 +147,8 @@ func (s TestkubeAPI) executeTest(ctx context.Context, test testkube.Test, reques
 		request.Name = rand.Name()
 	}
 
+	request.Number = s.getNextExecutionNumber(test.Name)
+
 	// test name + test execution name should be unique
 	execution, _ = s.ExecutionResults.GetByNameAndTest(ctx, request.Name, test.Name)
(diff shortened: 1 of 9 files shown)


internal/app/api/v1/executions.go
```diff
@@ -569,6 +571,18 @@ func (s *TestkubeAPI) streamLogsFromJob(executionID string, w *bufio.Writer) {
 	}
 }
 
+func (s TestkubeAPI) getNextExecutionNumber(testName string) int {
+	execution, err := s.ExecutionResults.GetLatestByTest(context.Background(), testName, "number")
```
vsukhin, line 575: it will work only if we don't have parallel runnng tests

vsukhin (approved):

internal/app/api/v1/executions.go
exu: It should be done on top of some atomic counter, e.g. by using Mongos' findAndModify method on top of some counters collection, similar like posgresql sequences work.

nicufk: Done follow up task ✅

PR #1681, opened 2022-06-20
fix: capture init log errors

vsukhin:
## Pull request description 

Capture init log errors

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/jobs/jobclient.go
+++ b/pkg/jobs/jobclient.go
@@ -465,9 +465,7 @@ func (c *JobClient) TailPodLogs(ctx context.Context, pod corev1.Pod, logs chan [
 			return err
 		}
 
-		go func(ctr *string) {
-			defer close(logs)
-
+		go func() {
 			scanner := bufio.NewScanner(stream)
 
 			// set default bufio scanner buffer (to limit bufio.Scanner: token too long errors on very long lines)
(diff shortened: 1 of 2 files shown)


pkg/jobs/jobclient.go
```diff
@@ -377,79 +377,111 @@ func (c *JobClient) GetLastLogLineError(ctx context.Context, podNamespace, podNa
...
-	if err != nil {
-		return logs, err
+		podLogRequest := c.ClientSet.CoreV1().
+			Pods(c.Namespace).
+			GetLogs(pod.Name, &podLogOptions)
+
+		stream, err := podLogRequest.Stream(context.TODO())
+		if err != nil {
+			if len(logs) != 0 {
+				return logs, nil
```
vLia, line 409: is there a way to check for the type of error before returning nil?

pkg/jobs/jobclient.go
```diff
@@ -377,79 +377,111 @@ func (c *JobClient) GetLastLogLineError(ctx context.Context, podNamespace, podNa
...
+			return logs, err
+		}
+
+		defer stream.Close()
+
+		buf := new(bytes.Buffer)
+		_, err = io.Copy(buf, stream)
+		if err != nil {
+			if len(logs) != 0 {
+				return logs, nil
```
vLia, line 421: same here, should we make sure it's the right type of error before returning half-streamed logs with no error?

vLia (approved):
looks good 👍  had one minor question

pkg/jobs/jobclient.go
vsukhin: same approach

vsukhin: https://github.com/kubernetes/test-infra/blob/d408ead6e01bf53681bb1758f5a4728b28628fdd/prow/cmd/deck/main.go#L1368

pkg/jobs/jobclient.go
vLia: Looks good enough! 👍

PR #1741, opened 2022-06-30
fix: improved telemetry debug on cluster

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/internal/app/api/v1/server.go
+++ b/internal/app/api/v1/server.go
@@ -160,6 +160,19 @@ func (s *TestkubeAPI) WithTelemetry(enabled bool) {
 	s.TelemetryEnabled = enabled
 }
 
+// SendTelemetryStartEvent sends anonymous start event to telemetry trackers
+func (s TestkubeAPI) SendTelemetryStartEvent() {
+	if !s.TelemetryEnabled {
+		return
+	}
+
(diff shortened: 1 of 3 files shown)


internal/app/api/v1/server.go
```diff
@@ -160,6 +160,19 @@ func (s *TestkubeAPI) WithTelemetry(enabled bool) {
...
+func (s TestkubeAPI) SendTelemetryStartEvent() {
+	if !s.TelemetryEnabled {
+		return
+	}
+
+	out, err := telemetry.SendServerStartEvent(s.Config.ClusterID, api.Version)
+	if err != nil {
+		s.Log.Debug("telemetry send error", "error", err.Error())
+	}
+	s.Log.Debugw("sending telemetry server start event", "output", out)
```
vsukhin, line 173: Probably, should not be logged in case of error

vsukhin (approved):
one minor note

internal/app/api/v1/server.go
exu: good catch! fixed!

Issue #1647, opened 2022-06-13
Add `kubectl testkube debug` command which will generate markdown with cluster info for passing into new issue


exu:
**Is your feature request related to a problem? Please describe.**

As a Testkube user, I don't want to pass data manually when creating new issue.
 
Instead, I want to have nicely formatted markdown with all possible data which can help with debugging.

We could generate: 
- Testkube client version 
- Testkube cluster version 
- Testkube last X logs from API 
- Testkube last X logs from Operator 
- Testkube last X logs from Executors (try locate last execution pod) 

- ... more?

TheBrunoLopes:
nice idea, adding to the next release.

mszostok:
Hi, just was passing by :) 

>Is there any way to open Testkube issue automatically + copy to clipboard ?

Even better, via GitHub API you can actually create an issue. However, in this case, it will be much better to just open an issue form in browser, so user can still edit it. It can be achieved via [URL query params](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-url-query).

Here is a small example how it may work: https://github.com/mszostok/issue-shooter
Recording: 

https://user-images.githubusercontent.com/17568639/173876337-7e4f3b05-86f3-4d95-82f9-4659c5dd2bd4.mp4

---

BTW, I noticed a minor bug in [`enhancement.md`](https://github.com/kubeshop/testkube/blame/main/.github/ISSUE_TEMPLATE/enhancement.md#L5) template. The label on repo is called `enhancement` while in template it's `enhancement 🚀`. As a result, the label is not applied when creating an issue.

exu:
@mszostok thanks for great tips ;) @vLia can we utilize them ?

vLia:
@exu yes, we can :)

PR #1861, opened 2022-07-20
fix: update Testkube Swagger definition

vLia:
## Pull request description 
While working on issue #1647 it came up that some of the already implemented properties from the Go code do not appear in the Swagger definition.
This commit makes sure that the newly generated models using `make openapi-generate-model` will not break old functionality.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- updated swagger yaml to match go code

## Fixes

-

### DIFF
--- a/api/v1/testkube.yaml
+++ b/api/v1/testkube.yaml
@@ -1802,7 +1802,7 @@ components:
             description: test execution results
         labels:
           type: object
-          description: "test suite execution labels"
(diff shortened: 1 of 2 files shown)


vsukhin (approved):
good catch

PR #2003, opened 2022-08-05
feat: create bug ticket

### LINKED ISSUE #1937: Add `kubectl testkube debug create-issue` command to automatically create GitHub issue with debug info from the CLI
**Is your feature request related to a problem? Please describe.**
Follow-up to https://github.com/kubeshop/testkube/issues/1647

As a **Testkube user** I would like to **have my environment info automatically populated when opening bug reports**, so that I can **focus more on describing my issue**.

Issue#1647 implements the `tk debug info` command to show summary of the Testkube environment information. The purpose of this Issue is to implement using the GitHub API the creation of a new bug report.

**Describe the solution you'd like**
`kubectl testkube debug create-issue` should allow the user to populate a Testkube GitHub issue on the CLI and open up a browser to finalize the creation.

**Additional context**
The CLI UX of the `debug` command was promptly discussed during the UX meeting on 28.06.

vLia:
## Pull request description 
Closes #1937 

This PR opens a browser with a new issue from the command line populating the versions and pointing to `debug info` to add more context.

Adding the exact logs this way is not possible due to **URL size limitations** - the body of the issue is added as a URL param.
**Shortening the logs** to match the size limitations **ends us up with a few characters** from each pod - in scaled up Testkube clusters this might be even worse.

Another option would be to create the Issue **using the GitHub API** without opening a browser window. This would depend to have **logging into GitHub implemented** which we currently don't have - and it is not worth the effort at this point.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/kubectl-testkube/commands/debug/create.go
```diff
@@ -0,0 +1,22 @@
+package debuginfo
```
exu, line 1: Can we make it just `tk create-ticket`?

exu (approved):
One question

cmd/kubectl-testkube/commands/debug/create.go
vLia: We can, I just did not want to saturate the list of commands more here. But on it

PR #2063, opened 2022-08-15
feat: add username and token secrets

vsukhin:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/kubectl-testkube/commands/tests/create.go
```diff
@@ -164,6 +178,18 @@ func validateCreateOptions(cmd *cobra.Command) error {
 		}
 	}
 
+	if len(gitUsernameSecret) > 1 {
```
exu, line 181: will it work? looks like first or second will always fail

exu (approved):
one comment. as I'm not sure about logic behind.

PR #2085, opened 2022-08-17
feat: use config map for storing config

vsukhin:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/api-server/main.go
```diff
@@ -93,16 +94,35 @@ func main() {
...
-	configRepository := config.NewMongoRespository(db)
+	configRepository := configmongo.NewMongoRespository(db)
+	configMapConfig, err := configmap.NewConfigMapConfig(os.Getenv("APISERVER_CONFIG"), namespace)
+	ui.ExitOnError("Getting config map config", err)
 
+	ctx := context.Background()
 	// try to load from mongo based config first
-	telemetryEnabled, err := configRepository.GetTelemetryEnabled(context.Background())
+	telemetryEnabled, err := configMapConfig.GetTelemetryEnabled(ctx)
 	if err != nil {
```
exu, line 104: will this logic be still applicable? 

if err -> load from envs (fallback)

cmd/api-server/main.go
```diff
@@ -93,16 +94,35 @@ func main() {
...
 		// fallback to envs in case of failure (no record yet, or other error)
 		telemetryEnabled = envs.IsTrue("TESTKUBE_ANALYTICS_ENABLED")
 	}
 
-	clusterId, err := configRepository.GetUniqueClusterId(context.Background())
+	var clusterId string
+	config, err := configMapConfig.Get(ctx)
```
exu, line 110: We still have GetUniqueClusterId method? Why not use it here?

exu: Ok I see here is just writing to storage in case

exu (approved):

exu:
one commemnt

cmd/api-server/main.go
vsukhin: Yes, this logic was not changed, we try to read config map, if we can't do it (it doesn't exist), then we get it from env var

cmd/api-server/main.go
vsukhin: yes, we try to read from config map, if we can't get it, then we try to read mongo config, and if we don't have access to mongo config, we generate and save a new  cluster id

PR #2226, opened 2022-09-05
feat: obfuscate secret values in the execution output

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/internal/app/api/v1/executions.go
+++ b/internal/app/api/v1/executions.go
@@ -365,7 +365,7 @@ func (s *TestkubeAPI) streamLogsFromJob(execution *testkube.Execution, w *bufio.
 	enc := json.NewEncoder(w)
 	s.Log.Infow("getting logs from Kubernetes job")
 
-	executor, err := s.getExecutorByTestType(testType)
+	executor, err := s.getExecutorByTestType(execution.TestType)
 	if err != nil {
 		output.PrintError(os.Stdout, err)
 		s.Log.Errorw("getting logs error", "error", err)
@@ -373,7 +373,7 @@ func (s *TestkubeAPI) streamLogsFromJob(execution *testkube.Execution, w *bufio.
 		return
 	}
 
-	logs, err := s.Executor.Logs(execution)
+	logs, err := executor.Logs(execution)
 	s.Log.Debugw("waiting for jobs channel", "channelSize", len(logs))
 	if err != nil {
 		output.PrintError(os.Stdout, err)


pkg/api/v1/testkube/model_variables_extended.go
```diff
@@ -20,3 +13,13 @@ func VariablesToMap(v Variables) map[string]string {
 
 	return vars
 }
+
+func ObfuscateSecrets(output string, variables Variables) string {
+	for _, v := range variables {
+		if v.Type_ == VariableTypeSecret {
+			re := regexp.MustCompile(`(` + v.Name + `',\s*')([^\s']*)`)
```
vsukhin, line 20: Is this same pattern for all executors or particluar ones?

vsukhin (commented):
one question

pkg/api/v1/testkube/model_variables_extended.go
```diff
@@ -20,3 +15,13 @@ func VariablesToMap(v Variables) map[string]string {
 
 	return vars
 }
+
+func ObfuscateSecrets(output string, variables Variables) string {
+	for _, v := range variables {
+		if v.Type_ == VariableTypeSecret {
+			output = strings.ReplaceAll(output, `'`+v.Value+`'`, "********")
+			output = strings.ReplaceAll(output, `"`+v.Value+`"`, "********")
```
vsukhin, line 23: This is even more complex. We don't have values for secret vars with secret references. They are propagated when we run an executor pod

vsukhin (commented):
and one more question

pkg/api/v1/testkube/model_variables_extended.go
```diff
@@ -20,3 +18,31 @@ func VariablesToMap(v Variables) map[string]string {
...
+	var secretKeyValues map[string]string
+	if err == nil {
+		secretKeyValues, _ = secretClient.Get(secret.GetMetadataName(testName))
+	}
+
+	for _, v := range variables {
+		secretValue := ""
+		if *v.Type_ == SECRET_VariableType {
+			secretValue = v.Value
+		} else if *v.Type_ == SECRET_VariableType && v.SecretRef != nil {
```
vsukhin, line 38: this will never be called

pkg/api/v1/testkube/model_variables_extended.go
```diff
@@ -20,3 +18,31 @@ func VariablesToMap(v Variables) map[string]string {
...
+func ObfuscateSecrets(output string, variables Variables, testName string) string {
+	// TODO: this is ugly, does anybody have a better idea?
+	namespace := "testkube"
+	if ns, ok := os.LookupEnv("TESTKUBE_NAMESPACE"); ok {
+		namespace = ns
+	}
+	secretClient, err := secret.NewClient(namespace)
+	var secretKeyValues map[string]string
+	if err == nil {
+		secretKeyValues, _ = secretClient.Get(secret.GetMetadataName(testName))
```
vsukhin, line 31: It works this way only for git credentials. But for secret variables we need to take the name of secret and the secret key from SecretRef object. Also, check the error, if it's something rather than not found

vsukhin (commented):
more comments

pkg/api/v1/testkube/model_variables_extended.go
```diff
@@ -20,3 +21,35 @@ func VariablesToMap(v Variables) map[string]string {
...
+func ObfuscateSecrets(output string, variables Variables, testName string) string {
+	// TODO: this is ugly, does anybody have a better idea?
+	namespace := "testkube"
+	if ns, ok := os.LookupEnv("TESTKUBE_NAMESPACE"); ok {
+		namespace = ns
+	}
+	secretClient, err := secret.NewClient(namespace)
+	var secretKeyValues map[string]string
+	if err == nil {
+		secretKeyValues, err = secretClient.Get(secret.GetMetadataName(testName))
```
vsukhin, line 34: For variables the secret will be v.SecretRef.Name, if v.SecretRef != nil

vsukhin (commented):
one comment

pkg/api/v1/testkube/model_variables_extended.go
exu: We have Obfuscate logic already implemented

image

pkg/executor/client/interface.go
```diff
@@ -27,7 +27,7 @@ type Executor interface {
 	// Abort aborts pending execution, do nothing when there is no pending execution
 	Abort(id string) (execution *testkube.ExecutionResult)
 
-	Logs(id string) (logs chan output.Output, err error)
+	Logs(execution *testkube.Execution) (logs chan output.Output, err error)
```
exu, line 30: Why change interface of client?

exu (changes requested):
Obfuscate already there,

The problem is that it'll not handle use case where executor is logging execution.

Issue #2223, opened 2022-09-02
Add --executor-args support in argocd plugin


sharmajee1:
**Is your feature request related to a problem? Please describe.**
Currently, `testkube generate tests-crds` doesn’t have any option to pass `--executor-args` option.
Add this support so that we can use this in argocd plugin.

**Describe the solution you'd like**
```
source:
   repoURL: https://github.com/kubeshop/testkube-argocd.git
   targetRevision: HEAD
   path: examples/postman-collections
   plugin:
     name: testkube
     extraArgs:
       - '--insecure'
```

sharmajee1:
Additionally, please add support to use other Newman args like `--ssl-client-cert <path>`  `--ssl-client-key <path>`

sharmajee1:
> @sharmajee1 what is your use case, we are trying to understand what solution will be better for this, right now we implemented one that sets the executor args provided in the command line for every test.

Currently, we are using argocd plugin for test creation, and we need to create those test with extra args, for example `--insecure` `--ssl-client-cert <path> --ssl-client-key <path>` 

because for now `testkube generate tests-crds` doesn’t support those extra args

sharmajee1:
> so every test will need the same args right?

Yes

PR #2288, opened 2022-09-15
feat: add executor args and envs to generate crds

nicufk:
## Pull request description 

Closes https://github.com/kubeshop/testkube/issues/2223

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/crds/tests_crds.go
+++ b/cmd/kubectl-testkube/commands/crds/tests_crds.go
@@ -19,6 +19,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	executorArgs []string
+	envs         map[string]string
+)
+
 // NewCRDTestsCmd is command to generate test CRDs
 func NewCRDTestsCmd() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


vsukhin (commented):
Not sure, that this is an implementation we need, because we just add same parameters for all tests and all envs. And we also then need to change how we call argocd plugin

sharmajee1:
> Not sure, that this is an implementation we need, because we just add same parameters for all tests and all envs. And we also then need to change how we call argocd plugin

But if we use this argocd plugin with --insecure options that mean we intentionally want all tests to add that flag, but it should be pass during the plugin.

For example.

```
source:
   repoURL: https://github.com/kubeshop/testkube-argocd.git
   targetRevision: HEAD
   path: examples/postman-collections
   plugin:
     name: testkube
     extraArgs:
       - '--insecure
       
```

If I don't pass this tests should be created without the flag

vsukhin (approved):

sharmajee1:
I don’t see this command working in my environment 
```
testkube generate tests-crds . --executor-args '--insecure'
```
Test CRDs are getting generated but not with executor-args not getting updated.

Issue #2216, opened 2022-09-02
Add option to add configuration files to executions


vLia:
**Is your feature request related to a problem? Please describe.**
Related to issue https://github.com/kubeshop/testkube/issues/2185
Some testing frameworks allow configuration via files (variables files in Postman, settings.xml in Maven). Testkube users should be able to configure executions with configuration files, and for this they should be able to specify the content and the name of the files.

**Describe the solution you'd like**
There should be a way to specify both the names and the contents of these files. An option would be to pass it via cli arguments, another to use volumes.
// TODO: check if any of the executors require multiple configuration files to see how to design the commands for this

**Describe alternatives you've considered**
For maven a PR was already opened, but the solution can be improved: https://github.com/kubeshop/testkube-executor-maven/pull/10

sharmajee1:
Yes, this will be a useful feature.
Many of our test use Newman, and using the below options we can pass the client certs.
```
--ssl-client-cert <path>              Specify the path to a client certificate (PEM)
--ssl-client-key <path>               Specify the path to a client certificate private key
--ssl-client-passphrase <passphrase>  Specify the client certificate passphrase (for protected key)
```

We need a similar feature in testkube, wither via mount or CLI, Please also support this feature on ArgoCD plugin

vLia:
Hello @sharmajee1,

Please use `--args` for anything that is not a file. For files, the `--copy-files` parameter has been released and should be working with all executors on execution and test levels. Please keep in mind that configuring it on tests only configures the path, and the files will be read on execution. 
There's an example command on:
https://kubeshop.github.io/testkube/using-testkube/tests/tests-running/#mapping-local-files

Hope this helps with you use case. In case it is not sufficient, please open a new ticket.

Closing this ticket as the feature was implemented and released.

sharmajee1:
> Hello @sharmajee1,
> 
> Please use `--args` for anything that is not a file. For files, the `--copy-files` parameter has been released and should be working with all executors on execution and test levels. Please keep in mind that configuring it on tests only configures the path, and the files will be read on execution. There's an example command on: https://kubeshop.github.io/testkube/using-testkube/tests/tests-running/#mapping-local-files
> 
> Hope this helps with your use case. In case it is not sufficient, please open a new ticket.
> 
> Closing this ticket as the feature was implemented and released.

Thanks for the response, we are using argoCD along with test-crds and such features still not present in test-crds,

PR #2363, opened 2022-09-25
feat: add file mapping to tests and executions

vLia:
## Pull request description 

Adds option to map files from host specified on Test and Execution levels.
See details: https://github.com/kubeshop/testkube/issues/2216

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

- added new flag called `--copy-files` to create and run test commands

## Fixes

-


pkg/executor/content/config_files.go
```diff
@@ -0,0 +1,16 @@
+package content
+
+import (
+	"fmt"
+	"os"
+)
+
+func PlaceFiles(files map[string][]byte) error {
```
vsukhin, line 8: Will we need to update all executors?

vsukhin (approved):
looks good

pkg/executor/content/config_files.go
vLia: yes, I will work on that today

PR #2502, opened 2022-10-10
feat: add lease checker to test triggers & add graceful shutdown to testkube

dejanzele:
## Pull request description 
Add support for lease checker in test triggers so if multiple API replicas are running, only 1 replica gets lease to watch k8s resources.
Add graceful shutdown of HTTP server (wait for unfinished/currently serving requests, gracefully terminate k8s informers, shut down all threads, use context which finishes on SIGTERM & SIGKILL)

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

- add lease checker in test trigger service
- add graceful shutdown to testkube

## Fixes

-


pkg/utils/utils.go
```diff
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"bufio"
+	"math/rand"
```
vsukhin, line 5: just in case, math/rand is not secure comparing to crypto/rand

dejanzele: good point

exu (approved):

pkg/server/httpserver.go
```diff
@@ -84,6 +86,17 @@ func (s *HTTPServer) getProblemMessage(err error, context ...interface{}) string
...
+func (s *HTTPServer) Run(ctx context.Context) error {
+	// this function listens for finished context and calls graceful shutdown on the API server
+	go func() {
+		select {
+		case <-ctx.Done():
+			// sleep 2 seconds to cover the edge case if SIGTERM or SIGKILL signal occurs before the server is started,
+			// so the application does not get stuck
+			time.Sleep(2 * time.Second)
+			s.Log.Infof("shutting down Testkube API server")
+			_ = s.Mux.Shutdown()
```
vsukhin, line 98: Doesit return error, should we ignore it?

pkg/triggers/lease.go
```diff
@@ -0,0 +1,46 @@
...
+		select {
+		case <-ctx.Done():
+			s.logger.Infof("trigger service: stopping lease checker component")
+			return
+		case <-ticker.C:
+			lease, err := s.leaseBackend.CheckAndSet(ctx, s.identifier)
+			if err != nil {
+				s.logger.Errorf("error checking and setting lease: %v", err)
+			}
+			leaseChan <- lease.Identifier == s.identifier
```
vsukhin, line 27: guess, it will panic in case of error

pkg/triggers/lease_backend.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (b *MongoLeaseBackend) CheckAndSet(ctx context.Context, id string) (*Lease, error) {
+	after := options.After
+	upsert := true
+	opts := options.FindOneAndUpdateOptions{
+		ReturnDocument: &after,
+		Upsert:         &upsert,
+	}
+	res := b.coll.FindOneAndUpdate(ctx, bson.M{"_id": _id, "ds": ""}, bson.M{}, &opts)
```
vsukhin, line 35: And for multiple api replicas, we will update it multiple times?

pkg/server/httpserver.go
dejanzele: It is safe to ignore it as we cannot really do anything about it, the app is shutting down for some reason (sigterm, sigkill...)

pkg/triggers/lease.go
dejanzele: it is updated in the newest commit

pkg/triggers/lease_backend.go
dejanzele: Lease logic does the following:
* Get current lease
* if it doesn't exist, create it, if it fails because of duplicate key, somebody created it in the meantime and got the lease, the current replica will not get the lease
* if it exists, try to do a findAndUpdate which is an atomic operation in Mongo
* check the identifier in the returned document from findAndUpdate operation, if other replica did it in the meantime

pkg/triggers/watcher.go
```diff
@@ -25,34 +43,84 @@ func (s *Service) runWatcher(ctx context.Context) {
...
+	s.informers.daemonsetInformer.Informer().AddEventHandler(s.daemonSetEventHandler(ctx))
+	s.informers.statefulsetInformer.Informer().AddEventHandler(s.statefulSetEventHandler(ctx))
+	s.informers.serviceInformer.Informer().AddEventHandler(s.serviceEventHandler(ctx))
+	s.informers.ingressInformer.Informer().AddEventHandler(s.ingressEventHandler(ctx))
+	s.informers.clusterEventInformer.Informer().AddEventHandler(s.clusterEventEventHandler(ctx))
+	s.informers.testTriggerInformer.Informer().AddEventHandler(s.testTriggerEventHandler())
 
 	s.logger.Debugf("trigger service: starting pod informer")
-	go podInformer.Informer().Run(ctx.Done())
+	go s.informers.podInformer.Informer().Run(stop)
```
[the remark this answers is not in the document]
dejanzele: k8s informers get closed when the stop channel is closed, this is by design in the kubernetes client, when a lease is lost, I close the stop channel, when it is acquired, I recreate them using a new channel

pkg/triggers/lease_backend.go
vsukhin: Yes, but it might fail, if a few replcas try to obtain the lease for all of them. From what I remember, it was suggested to use retry appoach, at least we do it for creating uniqiue numbers for executions

dejanzele: I am not sure I fully understand the idea, can we discuss it internally?

PR #2510, opened 2022-10-12
fix: use secret envs with their name and also with numbers

nicufk:
## Pull request description 
The implementation that we have was specialised for Postman that expects the value of the secret env to be a json string that contains name and secret in it, but it will not be the same for the rest of the executors.
For now two variants will be provided, the env var with the secret key as the var name and the one with the name lwith RUNNER_SECRET_ENV%d format that can be parsed in a loop(this is how Postman is processing them)

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/secret/manager.go
+++ b/pkg/executor/secret/manager.go
@@ -9,10 +9,19 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	// SecretEnvVarPrefix is a prefix for secret env vars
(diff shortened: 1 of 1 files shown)


vLia (approved):
looks good 👍

vsukhin (approved):

PR #2516, opened 2022-10-13
feat: add cloud agent

povilasv:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/api-server/main.go
```diff
@@ -255,6 +256,26 @@ func main() {
 		scheduler,
 	)
 
+	tkcEnabled := os.Getenv("TESTKUBE_CLOUD_ENABLED")
```
exu, line 259: Will it be enough to just check if there is an API key?

exu (approved):
Small comment

pkg/agent/agent.go
```diff
@@ -0,0 +1,129 @@
...
+	handler fasthttp.RequestHandler
+	logger  *zap.SugaredLogger
+	apiKey  string
+}
+
+func NewAgent(logger *zap.SugaredLogger, handler fasthttp.RequestHandler, server, apiKey string, isInsecure bool) (*Agent, error) {
+
+	var security grpc.DialOption
+	security = grpc.WithTransportCredentials(credentials.NewTLS(nil))
+	if isInsecure {
```
vsukhin, line 30: I would use if else, if we defined security var

pkg/agent/agent.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, err
+	}
+
+	client := cloud.NewTestKubeCloudAPIClient(conn)
+	return &Agent{conn: conn, client: client, handler: handler, logger: logger, apiKey: apiKey}, nil
+}
+
+func (ag *Agent) Run(ctx context.Context) error {
+	var opts []grpc.CallOption
+	md := metadata.Pairs("api-key", ag.apiKey)
```
vsukhin, line 44: might be define a const for api-key

pkg/agent/agent.go
```diff
@@ -0,0 +1,129 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			ag.logger.Errorf("agent stream recv: %w", err)
+			return err
+		}
+
+		switch {
+		case cmd.Url == "healthcheck":
```
vsukhin, line 67: should we check ctx.Done()?

vsukhin, line 67: should it be const?

pkg/agent/agent.go
```diff
@@ -0,0 +1,129 @@
...
+					ag.logger.Errorf("stream send: %w", err)
+				}
+				return err
+			}
+			r.SetURI(uri)
+
+			req.Init(r, nil, nil)
+			ag.handler(req)
+
+			fasthttp.ReleaseRequest(r)
```
vsukhin, line 102: no errors in all statements above?

pkg/agent/agent.go
```diff
@@ -0,0 +1,129 @@
...
+
+			headers := make(map[string]*cloud.HeaderValue)
+			req.Response.Header.VisitAll(func(key, value []byte) {
+				_, ok := headers[string(key)]
+				if !ok {
+					headers[string(key)] = &cloud.HeaderValue{Header: []string{string(value)}}
+					return
```
vsukhin, line 109: what do we return there?

vsukhin (approved):
looks good, some minor notes

PR #2528, opened 2022-10-13
feat: add bulk update test triggers endpoint

dejanzele:
## Pull request description 

Add support for bulk update test triggers

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- adds bulk update test triggers endpoint
- generate trigger name from spec if name is not provided
- wipe & clean approach

## Fixes

-

### DIFF
--- a/internal/app/api/v1/testtriggers.go
+++ b/internal/app/api/v1/testtriggers.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	testtriggersv1 "github.com/kubeshop/testkube-operator/apis/testtriggers/v1"
 	"net/http"
 
 	"github.com/pkg/errors"
@@ -78,7 +79,7 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 		testTriggerSpec := testtriggersmapper.MapTestTriggerUpsertRequestToTestTriggerCRD(request)
 		testTrigger.Spec = testTriggerSpec.Spec
 		testTrigger.Labels = request.Labels
-		updated, err := s.TestKubeClientset.TestsV1().TestTriggers(namespace).Update(c.UserContext(), testTrigger, v1.UpdateOptions{})
+		testTrigger, err = s.TestKubeClientset.TestsV1().TestTriggers(namespace).Update(c.UserContext(), testTrigger, v1.UpdateOptions{})
 
 		s.Metrics.IncUpdateTestTrigger(err)
 
@@ -86,9 +87,7 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 			return s.Error(c, http.StatusBadGateway, err)
 		}
 
-		c.Status(http.StatusOK)
-
-		apiTestTrigger := testtriggersmapper.MapCRDToAPI(updated)
+		apiTestTrigger := testtriggersmapper.MapCRDToAPI(testTrigger)
 
 		if c.Accepts(mediaTypeJSON, mediaTypeYAML) == mediaTypeYAML {
 			data, err := crd.GenerateYAML(crd.TemplateTestTrigger, []testkube.TestTrigger{apiTestTrigger})
@@ -99,7 +98,7 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 	}
 }
 
-// BulkUpdateTestTriggersHandler is a handler for buklk updates an existing TestTrigger CRDs based on array of TestTrigger content
+// BulkUpdateTestTriggersHandler is a handler for bukl updates an existing TestTrigger CRDs based on array of TestTrigger content
 func (s *TestkubeAPI) BulkUpdateTestTriggersHandler() fiber.Handler {
 	return func(c *fiber.Ctx) error {
 		var request []testkube.TestTriggerUpsertRequest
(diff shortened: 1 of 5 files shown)


internal/app/api/v1/testtriggers.go
```diff
@@ -96,6 +99,51 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 	}
 }
 
+// BulkUpdateTestTriggersHandler is a handler for buklk updates an existing TestTrigger CRDs based on array of TestTrigger content
+func (s *TestkubeAPI) BulkUpdateTestTriggersHandler() fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		var request []testkube.TestTriggerUpsertRequest
+		err := c.BodyParser(&request)
```
vsukhin, line 106: have you tested it out?

internal/app/api/v1/testtriggers.go
```diff
@@ -96,6 +99,51 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
...
+
+			if err != nil {
+				return s.Error(c, http.StatusBadGateway, err)
+			}
+
+			apiTestTrigger := testtriggersmapper.MapCRDToAPI(updated)
+			response = append(response, apiTestTrigger)
+		}
+
+		c.Status(http.StatusOK)
```
vsukhin, line 141: it's default status

vsukhin (commented):
a couple of questions, but key one, is this just bulk update, or batch processing, like I did for test sources?

exu (approved):
Looks good, assuming the case when all updated resources exists already?

internal/app/api/v1/testtriggers.go
```diff
@@ -96,6 +99,51 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
...
+			testTrigger, err := s.TestKubeClientset.TestsV1().TestTriggers(namespace).Get(c.UserContext(), upsertRequest.Name, v1.GetOptions{})
+			if err != nil {
+				return s.Error(c, http.StatusBadGateway, err)
+			}
+
+			// map TestSuite but load spec only to not override metadata.ResourceVersion
+			testTriggerSpec := testtriggersmapper.MapTestTriggerUpsertRequestToTestTriggerCRD(upsertRequest)
+			testTrigger.Spec = testTriggerSpec.Spec
+			testTrigger.Labels = upsertRequest.Labels
+			updated, err := s.TestKubeClientset.TestsV1().TestTriggers(namespace).Update(c.UserContext(), testTrigger, v1.UpdateOptions{})
```
exu, line 129: We need also handle new ones

exu (changes requested):
We need to handle new triggers too, as array can contain  non existing resources.

internal/app/api/v1/testtriggers.go
dejanzele: yes, worked for my test case

exu (approved):

internal/app/api/v1/testtriggers.go
```diff
@@ -96,6 +99,76 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
...
+
+				s.Metrics.IncUpdateTestTrigger(err)
+
+				if err != nil {
+					err := errors.Wrapf(err, "error updating test trigger %s/%s", testTrigger.Namespace, testTrigger.Name)
+					return s.Error(c, http.StatusBadGateway, err)
+				}
+			}
+			testTriggers = append(testTriggers, testtriggersmapper.MapCRDToAPI(testTrigger))
+		}
```
vsukhin, line 161: what about deleted test triggers?

dejanzele: this endpoint can only be used for bulk update and insert operations, delete trigger is a separate endpoint

vsukhin: I  don't know zn original Pavlo request, but I can think, that he just wants to send an array of test triggers (added, updated or deleted) and api should just sync this array with existing crds

pavloburchak: @vsukhin @dejanzele that is true, I can even send an empty array, then every trigger should be removed

vsukhin: It will work, of course, if we delete all, then create all, but we might use something a bit more clever like I did for
https://github.com/kubeshop/testkube/blob/main/internal/app/api/v1/testsource.go#L153

internal/app/api/v1/testtriggers.go
```diff
@@ -201,3 +273,17 @@ func (s *TestkubeAPI) GetTestTriggerKeyMapHandler() fiber.Handler {
 		return c.JSON(triggerskeymapmapper.MapTestTriggerKeyMapToAPI(triggers.NewKeyMap()))
 	}
 }
+
+// generateName function generates a trigger name from the TestTrigger spec
+// function also takes care of name collisions, not exceeding k8s max object name (63 characters) and not ending with a hyphen '-'
+func generateName(t *testtriggersv1.TestTrigger) string {
+	name := fmt.Sprintf("trigger-%s-%s-%s-%s", t.Spec.Resource, t.Spec.Event, t.Spec.Action, t.Spec.Execution)
+	if len(name) > 57 {
```
vsukhin, line 281: better to use const for 57, 56

internal/app/api/v1/testtriggers.go
```diff
@@ -201,3 +273,17 @@ func (s *TestkubeAPI) GetTestTriggerKeyMapHandler() fiber.Handler {
...
+		name = name[:56]
+	}
+	if !strings.HasSuffix(name, "-") {
+		name += "-"
+	}
+	name = fmt.Sprintf("%s%s", name, utils.RandAlphanum(5))
```
vsukhin, line 287: and const there

vsukhin (commented):
couple of comments

PR #2534, opened 2022-10-16
feat: Executor tests - run, schedule, video disabled for cypress projects to speed up execution

tkonieczny:


test/cypress/executor-smoke/crd/crd.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     repository:
       type: git
       uri: https://github.com/kubeshop/testkube
-      branch: main
+      branch: executor-tests-run
```
vsukhin, line 13: Why do we switch to custom branch?

vsukhin (approved):
looks good, one question

test/cypress/executor-smoke/crd/crd.yaml
tkonieczny: I used it temporary for testing video param - it's already changed in the last commit

PR #2549, opened 2022-10-18
fix: fix issues in openapi

povilasv:
## Pull request description 

Fixing issues in openapi spec.

i used `make openapi-generate-model` did not have any changes

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/api/v1/testkube.yaml
+++ b/api/v1/testkube.yaml
@@ -84,7 +84,7 @@ paths:
           content:
             application/problem+json:
               schema:
-                type: object
+                type: array
                 items:
                   $ref: "#/components/schemas/Problem"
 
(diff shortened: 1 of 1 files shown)


api/v1/testkube.yaml
```diff
@@ -117,15 +117,15 @@ paths:
...
             application/problem+json:
               schema:
-                type: object
+                type: array
                 items:
                   $ref: "#/components/schemas/Problem"
         502:
           description: problem communicating with kubernetes cluster
           content:
             application/problem+json:
```
dejanzele, line 126: I think we don't return `application/problem+json` in case of an error but regular `application/json`.
Should we change the content type for errors here, or should we change the content type in controllers to return `application/problem+json`?

vsukhin (approved):

Issue #2638, opened 2022-10-28
Cannot reference multiple variables from one K8s Secret, when creating a new test via testkube's REST-API


TheLeanMachine:
**Describe the bug**
My use case:

- I want to create a new test in testkube via `POST => https://<TESKUBE_BASE_URL>/v1/tests`
- I need to store passwords for the tests in a secure way using Kubernetes secrets

In general, this works [exactly as documented](https://kubeshop.github.io/testkube/Secrets/#secret-environment-variables) - using the name of the K8s `Secret`'s as key in the JSON, and the name of the env var to set as value, for example...
![Screenshot 2022-10-28 at 12 48 53](https://user-images.githubusercontent.com/3470837/198570146-bb9581e0-161d-45e4-9c16-7649b587152e.png)

The problem arises, when you want to store **multiple** env vars in a `Secret`. For example, this will NOT work (as we are simply _overwriting the same key in a hash-map!_):

```
POST => <TESTKUBE_API_SERVER>/v1/test
{
   "secretEnvs": {
     "name-of-k8s-secret-CRD": "MY_ACTUAL_ENV_VAR",
     "name-of-k8s-secret-CRD": "MY_OTHER_ENV_VAR"
   }
}
```

**To Reproduce**
1. Create test via REST API setting multiple secretEnvs with same key
2. Result: Only last key will be stored in the test

**Expected behavior**
Each variable is considered. This - probably - requires an API-change/extension.

Purely for illustration purposes, here one possible API-sketch...
```
POST => <TESTKUBE_API_SERVER>/v1/test
{
   "secretEnvs": {
     "name-of-k8s-secret-CRD": ["MY_ACTUAL_ENV_VAR", "MY_OTHER_ENV_VAR"]
     "name-of-another-secret": ["YET_ANOTHER_VAR"]
   }
}
```

..and another one...
```
POST => <TESTKUBE_API_SERVER>/v1/test
{
   "secretEnvs": [{
      "name": "MY_ACTUAL_ENV_VAR",
      "type": "k8s-secret"
      "ref": "testkube-secret-name"
    }, {
      "name": "MY_OTHER_ENV_VAR",
      "type": "k8s-secret"
      "ref": "testkube-secret-name"      
    }]
}
```

**Version / Cluster**
- Which testkube version? 1.6.8
- What Kubernetes cluster? AWS EKS
- What Kubernetes version? v1.22.13-eks-15b7512

**Workaround**
[...]

PR #2639, opened 2022-10-28
fix: change order of secret name and secret key

vsukhin:
fix: change order of secret name and secret key to support multiple keys for same secret
#2638 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vsukhin:
In current scheme we don't allow to use k8s secret more than once. And in fact, unique are env vars, not the secrets

exu (approved):

Issue #2635, opened 2022-10-27
--copy-files argument not working for certain files


f-helm:
**Describe the bug**
Using `--copy-files` with `test create` produces error when running `test run`

### Case 1

**To Reproduce**
Steps to reproduce the behavior (file attachted to issue):
1. Run `kubectl testkube create test --file curl.json --name curl-test --copy-files users.csv:users.csv`
2. Run `kubectl testkube run test curl-test`

**Expected behavior**
Test is executed without error

**Actual behavior**
Error message `⨯ illegal base64 data at input byte 110009` when running `testkube get execution <id>`

**Additional context**
The file users.csv has 10002 lines, after removing 3 lines the execution (<10k lines) the execution worked

### Case 2

**To Reproduce**
Steps to reproduce the behavior (file attachted to issue):
1. Run `kubectl testkube create test --file curl.json --name curl-test --copy-files few_users.csv:few_users.csv`
2. Run `kubectl testkube run test curl-test`

**Expected behavior**
Test is executed without error

**Actual behavior**
Error message `⨯ illegal base64 data at input byte 0` when running `testkube get execution <id>`

**Additional context**
The message with input byte 0 appears in the sample file few_users.csv when a line is preceded with a #, if this is not the case then a proper number appears, but still same error

**Version / Cluster**
- Which testkube version? Client Version 1.6.12 Server Version v1.6.12
- What Kubernetes cluster? Rancher Desktop 16.0 (containerd)
- What Kubernetes version? v1.24.4

[users.csv](https://github.com/kubeshop/testkube/files/9880566/users.csv)
[few_users.csv](https://github.com/kubeshop/testkube/files/9880642/few_users.csv)

Issue #2501, opened 2022-10-10
Store files in storage instead of forwarding the contents as string


vLia:
**Describe the enhancement you'd like to see**
In https://github.com/kubeshop/testkube/issues/2408 it came up that files are expected to have a higher size than what the current string-based `--copy-files` implementation supports.

As discussed in the team, we would change the implementation to store these files in minio, and get them on test execution. To make it consistent with other behaviours in Testkube, this would mean that only specifying files on execution would override them, otherwise the test would take the file from storage.

chandana194:
Could you please let me know if any update on this ?

vLia:
Hello @chandana194,
the implementation is in progress, but not released yet. Will let you know when it is ready to use.

chandana194:
@vLia  Any update on this request ?

vLia:
Hello @chandana194 & @f-helm, `--copy-files` for big files was released for test creation with Testkube v1.6.38. Please check and let me know if it works for your use cases. The files will be copied into `/data/uploads/`, for an example check these docs: https://kubeshop.github.io/testkube/using-testkube/tests/tests-creating#mapping-local-files
Thanks in advance!

Issue #2408, opened 2022-09-28
Unable to copy parameter files from bitbucket to Kubernetes cluster through testkube create test open API


chandana194:
Hi 
I am unable copy parameter files and script files from bitbucket to Kubernetes where testkube is running.

Could you please let me know how can I achieve this through create new test API

vLia:
Hello @chandana194,

Could you please provide some more information:
- what version of Testkube are you using?
- what commands did you use to run the test?
- what errors are you getting?

You can use  `kubectl testkube debug info` to get these details

chandana194:
CLIENT VERSION | SERVER VERSION | CLUSTER VERSION  
-----------------+----------------+------------------
  1.5.20         | v1.5.3         | v1.20.15         

I am trying to create a test using /testkube-api-server/v1/tests endpoint by passing bitbucket details and the script path , and trying to copy few parameter files from my local to cluster. 
The test is getting created with 200 response but when I try to run it the testkube job is getting failed on the cluster

{
  "name": "param-test2",
  "namespace": "testkube",
  "type": "k6/script",
  "content": {
    "type": "git-file",
    "repository": {
      "type": "git",
      "uri": "https://tools.lowes.com/path_to_git.git",
      "branch": "main",
      "commit": "822eabd78a1",
      "path": "",
      "username": "my_cred",
      "token": "valid token is passed"   
    },
    "data": "testJSON.js",
    "uri": "/stash/projects/path-to-file/testJSON.js"
  },
  "source": "DACI_ScoreCard_PerfTesting",
  "created": "2022-07-30T06:54:15Z",
  
  "copyFiles": [
    "/Users/cnagara/Documents/K6Scripts/ParamsInK6/data.json:/data/data.json"
  ],
  "executionRequest": {
    "name": "testing parameters option",
    "testSuiteName": "",
    "number": 0,
    "executionLabels": {
      "users": "3",
      "prefix": "some-"
    },
    "namespace": "testkube",
    
    
    "image": "kubeshop/testkube-executor-custom:1.10.11-dev-0a9c91",
    "envs": {
      "record": "true",
      "prefix": "some-"
    },
    
    "sync": true,
    "copyFiles": {
      "property1": [
        null
      ],
      "property2": [
        null
      ]
    }
  }
}

vLia:
Testkube is at version 1.5.38 at the moment, files were introduced at 1.5.34.
Could you upgrade your local installation and share the exact errors that are thrown?

[...]

PR #2687, opened 2022-11-09
feat: use minio for copy files

vLia:
## Pull request description 

Fixes
https://github.com/kubeshop/testkube/issues/2635
https://github.com/kubeshop/testkube/issues/2501
https://github.com/kubeshop/testkube/issues/2408

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/kubectl-testkube/commands/tests/create.go
```diff
@@ -89,6 +89,11 @@ func NewCreateTestsCmd() *cobra.Command {
 
 				err = validateExecutorType(options.Type_, executors)
 				ui.ExitOnError("validating executor type", err)
+
+				if len(copyFiles) > 0 {
+					err := uploadCopyFiles(client, testName, "test", copyFiles)
```
vsukhin, line 94: probably, it's enum const

cmd/kubectl-testkube/commands/tests/run.go
```diff
@@ -88,13 +88,16 @@ func NewRunTestCmd() *cobra.Command {
 					os.Exit(1)
 				}
 
+				if len(copyFiles) > 0 {
+					err := uploadCopyFiles(client, name, "execution", copyFiles)
```
vsukhin, line 92: probably, worth to have typed const

api/v1/testkube.yaml
```diff
@@ -2100,6 +2100,34 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Problem"
+  /copy-files:
```
exu, line 2103: is copy-files a valid resource name?

exu: I would use just `files` and `POST /files`

cmd/kubectl-testkube/commands/tests/common.go
```diff
@@ -386,3 +387,22 @@ func mergeCopyFiles(testFiles []string, executionFiles []string) ([]string, erro
 
 	return result, nil
 }
+
+func uploadCopyFiles(client client.Client, parentID string, parentType string, files []string) error {
```
exu, line 391: upload or copy?

api/v1/testkube.yaml
```diff
@@ -2100,6 +2100,34 @@ paths:
...
+      summary: "Upload file"
+      description: "Upload file to be copied into test suite executions"
+      operationId: uploadTestSuiteCopyFile
+      requestBody:
+        $ref: "#/components/requestBodies/CopyFileBody"
+      responses:
+        200:
+          description: "successful operation"
+          content:
+            text/yaml:
```
vsukhin, line 2119: is this json or yaml response

internal/app/api/v1/copyfiles.go
```diff
@@ -0,0 +1,46 @@
...
+		if filePath == "" {
+			return s.Error(c, fiber.StatusBadRequest, errors.New("filePath cannot be empty"))
+		}
+
+		bucketName := getBucketName(parentType, parentID)
+		file, err := c.FormFile("attachment")
+		if err != nil {
+			return s.Error(c, fiber.StatusBadRequest, fmt.Errorf("unable to upload file: %w", err))
+		}
+		f, err := file.Open()
```
vsukhin, line 31: should we have defer f.Close()

vsukhin (approved):
looks good

cmd/kubectl-testkube/commands/tests/create.go
vLia: updated

cmd/kubectl-testkube/commands/tests/run.go
vLia: updated

api/v1/testkube.yaml
vLia: We have agreed on using `--copy-files` in the CLI. `--file` is already used to give the test files as input. (and we also agreed that `--config-file` is too restrictive, as these files will not always be used as configuration files)

While I do agree that copy is a verb in itself, the closest adjective would be `copied`, but that implies it was already copied over. For consistency's sake I'd keep `/copy-files`, unless there is a better and simpler option.

api/v1/testkube.yaml
vLia: it is `200`, but I'll add the json parts

cmd/kubectl-testkube/commands/tests/common.go
vLia: updated it to uploadFiles(..)

api/v1/testkube.yaml
vLia: After a discussion in a private channel, we decided to go with /uploads for this endpoint. The code is being updated and tested rn

PR #2727, opened 2022-11-15
fix: shorten bucket names

vLia:
## Pull request description 

shorten bucket names

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/storage/minio/minio.go
+++ b/pkg/storage/minio/minio.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


vsukhin (approved):

pkg/storage/minio/minio.go
```diff
@@ -276,3 +277,16 @@ func (c *Client) PlaceFiles(buckets []string, prefix string) error {
...
+func (c *Client) GetValidBucketName(parentType string, parentName string) string {
+	bucketName := fmt.Sprintf("%s-%s", parentType, parentName)
+	if len(bucketName) <= 63 {
+		return bucketName
+	}
+
+	h := fnv.New32a()
+	h.Write([]byte(bucketName))
+
+	return fmt.Sprintf("%s-%d", bucketName[:52], h.Sum32())
```
exu, line 291: Are we sure that string will always be with length above 52? As if not it'll probably panic

vLia: At this point bucketName is above 63, there is a check in line 284

Issue #2133, opened 2022-08-23
Container based executors - Artifacts


exu:
linked: https://github.com/kubeshop/testkube/issues/2061

- Allow users to define an executor by specifying 
         - docker image name
         - credentials (in case of private registries)
         - name of the executor (and other misc data t.b.d)
- Test can be defined to use this executor. The same behaviour is expected from a users's standpoint 
- There should be a clear folder structure documented that specifies
      - Which folder should the artifacts be saved too ?
      - Q: should the logs be saved in artifacts ?

- Pass & fail is solely determined by the exit status of the pods execution. but allow a Junit.xml report to be retrieved for extra info.
- Variables of type basic and secret should be mounted as environment variables on the container.
- Arguments defined by the user should be sent in a similar faction as this example:
    - if the user were to run a nginx image in k8s and send the command `ls –la /` in k8s, he would do it like this -> `kubectl exec nginx -- ls –la /`.  A similar behaviour is expected for the this container based executor and its arguments.
- Are resource limit important to set ? ``resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m" ``

TheBrunoLopes:
@ili-shmidt-wib here's the description of the feature we were discussing last week.
Feel free to add your input 👌

TheLeanMachine:
Hey @TheBrunoLopes ,

hate to ask, but: Is there a rough(!) ETA on this feature?

(I seriously don't want anyone to rush, but I will have to make a decision: Should I wait for this feature ...OR... seek out an try to find alternative for running my tests)

TheLeanMachine:
btw: @exu , @povilasv 

...as soon as this is released - or some kind of beta-version is publicly available -, I would our this "generic container executor" out.

TheBrunoLopes:
Hi @TheLeanMachine it's awesome that you are also wanting to test this feature out!
- The images don't need to be public, you will be able to define the pull secret of this image in case it's not already configured in your k8s cluster.
- We started implementing it yesterday, I can't promise you an exact date yet, but I would venture saying that it will probably take ~2 weeks to make a release out of it. 

- Is there a specific type of test that you want to run ? maybe we can help you with that.

TheLeanMachine:
Hey @TheBrunoLopes ,

(first of all: Thanks 4 the quick and conprehensive reply)

- Cool, will learn about "pull secrets & friends", then
[...]

PR #2748, opened 2022-11-17
feat: container artifacts

vsukhin:
## Pull request description 

Feature #2133 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/scheduler/test_scheduler.go
```diff
@@ -411,3 +414,29 @@ func mapK8sImagePullSecrets(secrets []v1.LocalObjectReference) []string {
...
+		return artifactBase
+	}
+
+	if artifactBase.VolumeName == "" && artifactAdjust.VolumeName != "" {
+		artifactBase.VolumeName = artifactAdjust.VolumeName
+	}
+
+	if artifactBase.VolumeMountPath == "" && artifactAdjust.VolumeMountPath != "" {
+		artifactBase.VolumeMountPath = artifactAdjust.VolumeMountPath
+	}
```
vLia, line 437: all these ifs could be shortened using a switch..case with empty statement in switch, ex:
```
switch {
case base == nil && adjust == nil:
return nil
case base != nil && adjust == nil:
return base
...
}
```

vLia (approved):
Looks great! 👍

pkg/scheduler/test_scheduler.go
vsukhin: hope, it's a bit better now

PR #2823, opened 2022-11-30
feat: set MongoDB SSL

vLia:
## Pull request description 

Set SSL details for MongoDB connection

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


internal/app/api/v1/events.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/websocket/v2"
 
-	"github.com/fluxcd/pkg/runtime/events"
+	events "github.com/fluxcd/pkg/apis/event/v1beta1"
```
vsukhin, line 11: Why did we change it?  Technically, this part is not used. We implemented this requirement using test triggers

cmd/api-server/main.go
```diff
@@ -331,3 +332,38 @@ func main() {
...
+		ui.Warn("Could not find sslClientCertificateKeyFile in secret %s", c.SSLSecretRef)
+	}
+	if caFile, ok = mongoSSLSecret["sslCertificateAuthorityFile"]; !ok {
+		ui.Warn("Could not find sslCertificateAuthorityFile in secret %s", c.SSLSecretRef)
+	}
+	if pass, ok = mongoSSLSecret["sslClientCertificateKeyFilePassword"]; !ok {
+		ui.Warn("Could not find sslClientCertificateKeyFilePassword in secret %s", c.SSLSecretRef)
+	}
+
+	err = os.WriteFile("mongodb.pem", []byte(keyFile), 0644)
```
vsukhin, line 358: Should we place in a server current dir or might be in some temp folder?

vsukhin (approved):
a coupleo of minor questions. good job!

internal/app/api/v1/events.go
vLia: I updated the dependencies, and they moved this part of the code, so the api-server wasn't building anymore. I just updated the path, it's pointing to the same struct.

cmd/api-server/main.go
vLia: good point, will update it

PR #3005, opened 2023-01-10
feat: test trigger conditions models

### LINKED ISSUE #2905: Add trigger that won't trigger until all pods have completed their deployment and are "ready"

I'd like to be able to automatically run the tests but only after a deployment has completed OR all pods within a deployment have completed.

Currently, I deploy via helm with ArgoCD. I have a set trigger that listens to the deployment and runs a test when the image is updated. However, because the pods could take x amount of time to deploy, I want to wait until that has completed before running the tests (I don't want the tests to run against old pods). I have a workaround for this which is add a delay to a test suite starting but that only solves a part of the issue.

Second issue is that if I deploy a trigger and a test together then the test won't run the first time. This is probably due to a race condition where the trigger/test isn't in place yet.

**Additional context**

We have a pull request that creates an environment so triggers/tests not being available at a set time is quite likely

vsukhin:
## Pull request description 

Fix #2905

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vLia (approved):
Looks good 👍

PR #3017, opened 2023-01-11
fix: check for nil pointer

vLia:
## Pull request description 

Fix nil pointer issue

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/process.go
+++ b/pkg/executor/process.go
@@ -17,7 +17,10 @@ import (
 // LoggedExecuteInDir will put wrapped JSON output to stdout AND get RAW output into out var
 // json logs can be processed later on watch of pod logs
 func Run(dir string, command string, envMngr secret.Manager, arguments ...string) (out []byte, err error) {
-	obfuscatedArgs := envMngr.Obfuscate([]byte(strings.Join(arguments, " ")))
+	obfuscatedArgs := []byte{}
+	if envMngr != nil {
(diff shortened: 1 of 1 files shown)


vsukhin (approved):

pkg/executor/process.go
```diff
@@ -17,7 +17,10 @@ import (
 // LoggedExecuteInDir will put wrapped JSON output to stdout AND get RAW output into out var
 // json logs can be processed later on watch of pod logs
 func Run(dir string, command string, envMngr secret.Manager, arguments ...string) (out []byte, err error) {
-	obfuscatedArgs := envMngr.Obfuscate([]byte(strings.Join(arguments, " ")))
+	obfuscatedArgs := []byte{}
+	if envMngr != nil {
+		obfuscatedArgs = envMngr.Obfuscate([]byte(strings.Join(arguments, " ")))
```
vsukhin, line 22: Do we need to read env vars - envMngr.GetVars(envMngr.Variables) ?

nicufk (approved):

pkg/executor/process.go
vLia: That part is done in every executor before Run() is called to execute the test - except for when nil is passed in in the init runner

PR #3075, opened 2023-01-19
feat: parallel testsuite

### LINKED ISSUE #3969: Implement remaining work on Test Suites V3
Current pain points with the v2 and/or v3 yaml are:

- There's fields that are present but don't work. such has the namespace field in the yaml inside a test step. We should allow the tests to run in the according namespace (if feasible) or remove the field entirely if not.
- We introduced the batch & items object that go inside each step of the execution which make the yaml more verbose and complex.
This is the current V2 yaml

Image

instead of adding the objects batch and items in each step (sorry don't have an v3 examples to put here)

Could it instead look something like this:

Image

(yaml for illustration purposes, sintax may not be 100% correct)

note: name is changed to test. there's no support for both.

- Is there any other field we should change ?

**Action points:**

- Review yaml and refactor accordingly.
- Implement UI seen in the mockup and change CLI as needed.
- v2 test suites should still be supported

vsukhin:
## Pull request description 

Fix #3969 #3971 #3977 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vLia (approved):
Looks great 👍 

Just one question/observation, not necessarily part of this issue, but related:
The names get a bit confusing. Could TestSuiteBatchStep be called just Batch? I am starting with the assumption that:
A TestSuite contains a collection of Batches to run (either in order or in parallel)
A Batch contains a collection of Tests to run (either in order or in parallel)
A Test contains a collection of Steps to run (either in order or in parallel)

The last one is not implemented yet, but we did have discussions on this. I think the current implementation using steps = tests gets confusing at some parts in the code.

nicufk (approved):
Looks good, I think we have to expose more logging on info to have more informations when debugging.

vsukhin:
@nicufk Yes, I added more logs to test suite execution. might be, will add a bit more
@vLia Yes, naming becomes a bit confusing, we have a few Batch types in OpenApi, that's why it's not a signle word, but at least corresponding field is always named `batch`

PR #3080, opened 2023-01-21
feat: add backoff and test for conditions timeout

dejanzele:
## Pull request description 

Add backoff when between conditions check iterations and add tests for retries

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/triggers/service.go
+++ b/pkg/triggers/service.go
@@ -25,34 +25,38 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-var (
-	defaultScraperInterval    = 5 * time.Second
-	defaultLeaseCheckInterval = 5 * time.Second
-	defaultMaxLeaseDuration   = 1 * time.Minute
-	defaultClusterID          = "testkube-api"
-	defaultIdentifierFormat   = "testkube-api-%s"
+const (
+	defaultScraperInterval        = 5 * time.Second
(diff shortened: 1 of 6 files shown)


vsukhin (approved):

pkg/triggers/service.go
```diff
@@ -25,34 +25,38 @@ import (
...
-	defaultLeaseCheckInterval = 5 * time.Second
-	defaultMaxLeaseDuration   = 1 * time.Minute
-	defaultClusterID          = "testkube-api"
-	defaultIdentifierFormat   = "testkube-api-%s"
+const (
+	defaultScraperInterval        = 5 * time.Second
+	defaultLeaseCheckInterval     = 5 * time.Second
+	defaultMaxLeaseDuration       = 1 * time.Minute
+	defaultConditionsCheckBackoff = 1 * time.Second
+	defaultConditionsCheckTimeout = 10 * time.Second
```
vsukhin, line 33: do you think 10 second is enough?

dejanzele: I don't know, it seems kinda low, I'd rather put it 60-90s

vsukhin: let's do 90

PR #3091, opened 2023-01-24
feat: lint & test pass

dejanzele:
## Pull request description 

Propagate context object
Fix resource leak in defer in a loop
Propagate error for EndExecution

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


internal/app/api/v1/executions.go
```diff
@@ -365,7 +364,7 @@ func (s TestkubeAPI) ListArtifactsHandler() fiber.Handler {
 // streamLogsFromResult writes logs from the output of executionResult to the writer
 func (s *TestkubeAPI) streamLogsFromResult(executionResult *testkube.ExecutionResult, w *bufio.Writer) error {
 	enc := json.NewEncoder(w)
-	fmt.Fprintf(w, "data: ")
+	_, _ = fmt.Fprintf(w, "data: ")
```
vsukhin, line 367: I would log errors in any case

pkg/executor/common.go
```diff
@@ -181,14 +184,14 @@ func GetJobPods(podsClient tcorev1.PodInterface, jobName string, retryNr, retryC
...
+		return GetJobPods(ctx, podsClient, jobName, retryNr+1, retryCount)
 	}
 	return pods, nil
 }
 
 // GetPodLogs returns pod logs bytes
-func GetPodLogs(c kubernetes.Interface, namespace string, pod corev1.Pod, logLinesCount ...int64) (logs []byte, err error) {
-	count := int64(100)
+func GetPodLogs(ctx context.Context, c kubernetes.Interface, namespace string, pod corev1.Pod, logLinesCount ...int64) (logs []byte, err error) {
+	var count int64 = 100
```
vsukhin, line 194: Is 100  a const?

vsukhin (approved):
looks good, two minor comments

pkg/executor/common.go
dejanzele: I don't know, this was from before, I just changed the semantics. Do you maybe know?

vsukhin: No, I didn't see it before. Put 100 to const at the beginning of the file

PR #3123, opened 2023-01-30
feat: testtrigger condition reason

### LINKED ISSUE #3092: Add "Reason" to trigger condition checks

As a part of #2905, a check is in place to see if a resource is "ready" before triggering a test(suite). This works well, but there is an issue with deployments in that if you have a minimum number of replicas, a pod is created before an old one is terminated. If this is the case then the deployment never leaves the "ready" state so it is triggered immediately.

It would be ideal to have another condition such as "reason" that allows us to determine exactly when a trigger should run. Perhaps even a delay field to ensure it doesn't pick the initial ready state?

vsukhin:
## Pull request description 

Fix #3092 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


nicufk (approved):

jfermi (commented):
Looks good.

PR #3134, opened 2023-02-01
feat: add cloud implementation for testresult repository

dejanzele:
## Pull request description 

Add CloudRepository implementation for testresult.Repository
Refactored code in cloud/data to be a bit cleaner

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/repository/testresult/filter.go
```diff
@@ -6,128 +6,128 @@ import (
...
-	statuses   testkube.TestSuiteExecutionStatuses
-	page       int
-	pageSize   int
-	textSearch string
-	selector   string
-}
-
-func NewExecutionsFilter() *filter {
-	result := filter{page: 0, pageSize: PageDefaultLimit}
+type FilterImpl struct {
```
exu, line 9: Looks like some dotnet or Java 

Do we have any convention for naming? In go there is interface and concrete implementation like repository and mongo repository or sql repository

vsukhin (approved):

pkg/repository/testresult/filter.go
dejanzele: The issue is that Interface type cannot be serialized over wire so I must pass the concrete implementation

PR #3168, opened 2023-02-09
feat: unify var env

### LINKED ISSUE #2540: Make test parametrization uniform
## Context
We realised that for each Executor the Variables, Environment variables and Arguments are behaving differently.
So we need uniformization and to define a standard on what's the expected behaviour for all executors.
## Current Terminology
The terminology we are using today.
### Variables
There's Type basic and or secret and are used to pass variables to be used by the executor during the execution.
Are used in both container and non container executors. In container executors they will be "defined" in the pod as environment variables as well. 

### Arguments
They are arguments or flags used in the executor binary after the command for e.g., for the curl executor the arguments could be
--args "-C -I -some-flag -foo=bar ..."

### Envs
Environment variables that are defined in the pod during test execution. Work the same both on container and non-container executors.

## Proposed Specification
- Variables and Environment variables will be the same thing.
   - All variables will by in the pod as Envs.
   - Non-container executors are still able to use it internally for other purposes.
- Remove any end user behaviour that differs from container executors or non-container executors.
- Deprecrate Envs field. (envs should still work if present, but docs and example won't show envs)
- CLI shows deprecated message when using --env

## Desired Terminology
We don't use parameters as a synonym of arguments. We say "you can parametrize your tests using Variables or Arguments"
### Variables
Will be used as Environment variables and accessible by your Tests during Test Executions.
### Arguments
Extra arguments used by the executors command binary.

vsukhin:
## Pull request description 

Fix #2540 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vLia (approved):
Looks good 👍

nicufk (approved):
Looks good

Issue #3198, opened 2023-02-15
test error: could not start process: chdir /data/repo: no such file or directory


chooco13:
**Describe the bug**
`playwright test error: could not start process: chdir /data/repo: no such file or directory`

**To Reproduce**
I created all resources using the testkube dashboard.
An error occurs when trying to test through a registered TestSource.

Test Definition
```
apiVersion: tests.testkube.io/v3
kind: Test
metadata:
  name: test
  namespace: testkube
spec:
  type: playwright/test
  content:
    repository:
      branch: master
  source: test-project
```

TestSource Definition
```
apiVersion: tests.testkube.io/v1
kind: TestSource
metadata:
  name: test-project
  namespace: testkube
spec:
  repository:
    type: git
    uri: {git_url}
```

Log Output
```
running test [63ec96cc1bc6bbb3804f6d5a]
🚚 Initializing...
🌍 Reading environment variables...
✅ Environment variables read successfully
RUNNER_ENDPOINT="testkube-minio-service-testkube:9000"
RUNNER_ACCESSKEYID="********"
RUNNER_SECRETACCESSKEY="********"
RUNNER_LOCATION=""
RUNNER_TOKEN=""
RUNNER_BUCKET="testkube-artifacts"
RUNNER_SSL=false
RUNNER_SCRAPPERENABLED="true"
RUNNER_GITUSERNAME=""
RUNNER_GITTOKEN=""
RUNNER_DATADIR="/data"
📦 Fetching test content from ...
📂 Fetching uploads from object store testkube-minio-service-testkube:9000...
📂 Placing files from buckets into /data/uploads/ []
📂 Getting the contents of bucket folders [test-test]
📂 Bucket testkube-artifacts does not exist
🚨 Could not place files: bucket testkube-artifacts does not exist
📂 Setting up access to files in /data
🔬 Executing in directory /data: 
 $ chmod 
✅ Execution succeeded
✅ Access to files enabled
✅ Initialization successful
0x400135bc20
running test [63ec96cc1bc6bbb3804f6d5a]
playwright test error: could not start process: chdir /data/repo: no such file or directory
```

**Expected behavior**
However, if I register the git address directly in Test, the test succeeds without any problems.

Test Definition
```
apiVersion: tests.testkube.io/v3
kind: Test
metadata:
  name: test
  namespace: testkube
spec:
  type: playwright/test
  content:
    type: git-dir
    repository:
      type: git-dir
      uri: {git_url}
      branch: master
```

[...]

PR #3210, opened 2023-02-17
fix: check git source when repo type is `git`

vLia:
## Pull request description 

During investigation on https://github.com/kubeshop/testkube/issues/3198, I saw that we do not have the option to specify the git type (git-file or git-dir) when we are using test sources via the UI.

This PR addresses that by adding a function to decide on the go what the type of the input is. Unfortunately it is not possible to get the type of a path without cloning the repo, so this means the repo will be cloned twice on execution: once in the api-server to check the type on the go, and once in the init-container. This has the upside that in case the type changes between the creation of the test and the execution, Testkube will be aware of it.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/content/fetcher.go
+++ b/pkg/executor/content/fetcher.go
@@ -53,17 +53,6 @@ func (f Fetcher) Fetch(content *testkube.TestContent) (path string, err error) {
 	case testkube.TestContentTypeGitDir:
 		return f.FetchGitDir(content.Repository)
 	case testkube.TestContentTypeEmpty:
-		if content.Repository != nil && content.Repository.Type_ == "git" {
-			path, err := f.FetchGitDir(content.Repository)
-			if err != nil {
-				return "", err
-			}
-			content.Type_ = string(testkube.TestContentTypeGitFile)
-
-			if isDir(path) {
-				content.Type_ = string(testkube.TestContentTypeGitDir)
-			}
-		}
 		output.PrintLog(fmt.Sprintf("%s Empty content type", ui.IconCross))
 		return path, nil
 	default:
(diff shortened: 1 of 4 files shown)


pkg/scheduler/test_scheduler.go
```diff
@@ -14,12 +14,15 @@ import (
...
 	"github.com/kubeshop/testkube/pkg/executor/client"
+	"github.com/kubeshop/testkube/pkg/executor/content"
 	testsmapper "github.com/kubeshop/testkube/pkg/mapper/tests"
 	"github.com/kubeshop/testkube/pkg/workerpool"
 )
 
 const (
 	containerType = "container"
+	// tempDir is the temporary directory used to evaluate a type of a 'git' source
+	tempDir = "/tmp"
```
vsukhin, line 25: you can use it like I did for git check

			dir, err := os.MkdirTemp("", "checkout")
			if err != nil {
				return s.Error(c, http.StatusBadGateway, err)
			}
			defer os.RemoveAll(dir) // clean up

pkg/scheduler/test_scheduler.go
```diff
@@ -247,6 +250,19 @@ func (s *Scheduler) getExecuteOptions(namespace, id string, request testkube.Exe
...
 
 		testCR.Spec = mergeContents(testCR.Spec, testSourceCR.Spec)
+
+		if testSourceCR.Spec.Type_ == "" && testSourceCR.Spec.Repository.Type_ == "git" {
+			fetcher := content.NewFetcher(tempDir)
+			content := testsmapper.MapTestContentFromSpec(testCR.Spec.Content)
+			t, err := fetcher.CalculateGitContentType(*content.Repository)
+			if err == nil {
+				testCR.Spec.Content.Type_ = t
+				testCR.Spec.Content.Repository.Type_ = t
```
vsukhin, line 260: guess, we use `git` for this now

pkg/scheduler/test_scheduler.go
vsukhin: Or os.GetTemp()

pkg/executor/content/fetcher.go
```diff
@@ -187,3 +189,37 @@ func (f Fetcher) configureUseOfCertificate() error {
...
 	return nil
 }
+
+// CalculateGitContentType returns the type of the git test source
+func (f Fetcher) CalculateGitContentType(repo testkube.Repository) (string, error) {
+	if repo.Uri == "" || repo.Path == "" {
+		return "", errors.New("repository uri and path should be populated")
+	}
+
+	dir, err := os.MkdirTemp("/tmp", "temp-git-files")
```
vsukhin, line 199: os.GetTemp()

vsukhin (approved):
some miinor comments

pkg/scheduler/test_scheduler.go
vLia: removed this line, tested with git-file and git-dir, still works

pkg/executor/content/fetcher.go
vLia: updated to empty string and let os.MkdirTemp(..) call this function

PR #3309, opened 2023-03-01
feat: add support for cloud artifacts; massive scraper refactor

dejanzele:
## Pull request description 

Added support for sending artifacts to cloud
Massive refactor of Scraper using Extractor -> Loader piping
Added unit and integration tests for Scraper

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/executor/scraper/filesystem_extractor.go
```diff
@@ -0,0 +1,55 @@
...
+
+				if fileInfo.IsDir() {
+					return nil
+				}
+
+				reader, err := e.fs.OpenFileBuffered(path)
+				if err != nil {
+					return errors.Wrapf(err, "error opening buffered %s", path)
+				}
+				object := &Object{
```
exu, line 38: Why not File - Object is quite generic name

pkg/executor/scraper/loader.go
```diff
@@ -0,0 +1,10 @@
+package scraper
+
+import (
+	"context"
+)
+
+//go:generate mockgen -destination=./mock_loader.go -package=scraper "github.com/kubeshop/testkube/pkg/executor/scraper" Loader
+type Loader interface {
```
exu, line 8: Isn't  it `Uploader`? As it's not loading anything :)

pkg/executor/scraper/minio_loader.go
```diff
@@ -0,0 +1,64 @@
...
+	err := client.Connect()
+	if err != nil {
+		return nil, errors.Errorf("error occured creating minio client: %v", err)
+	}
+	l.client = client
+
+	return l, nil
+}
+
+func (l *MinIOLoader) Load(ctx context.Context, object *Object, meta map[string]any) error {
```
exu, line 41: Looks like `Load` just Store/Save/Uploads file with `client.SaveFileDirect(` so Load is confusing

pkg/executor/scraper/scraper.go
```diff
@@ -14,43 +10,22 @@ type Scraper interface {
...
-		Endpoint:        endpoint,
-		AccessKeyID:     accessKeyID,
-		SecretAccessKey: secretAccessKey,
-		Location:        location,
-		Token:           token,
-		Bucket:          bucket,
-		Ssl:             ssl,
-	}
-
+type ScraperV2 struct {
```
exu, line 13: I think we should discuss versioning conventions (V2) usually in go there is subdirectory for V2

And do we need V2 here? As looks like you're overriding existing one

In This example I don't think we need V2 at all
Just from Testkube version 1.10 (some minor bump) we'll allow to handle Cloud + MinIO

exu (approved):

pkg/executor/scraper/scraper.go
vsukhin: If we introduce V2, then we need to have V1 as well. If you just refactor the package, you can keep existing version

pkg/executor/scraper/minio_loader.go
```diff
@@ -0,0 +1,64 @@
...
+	err := client.Connect()
+	if err != nil {
+		return errors.Errorf("error occured creating minio client: %v", err)
+	}
+
+	if err := utils.CheckStringKey(meta, "executionId"); err != nil {
+		return err
+	}
+
+	folder := meta["executionId"].(string)
```
vsukhin, line 52: possible panic

vsukhin: sounds like, it should be GetStringKey in order to not have this statement

pkg/cloud/data/scraper/loader.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+	if err := utils.CheckStringKey(info, "executionId"); err != nil {
+		return nil, err
+	}
+	if err := utils.CheckStringKey(info, "testName"); err != nil {
+		return nil, err
+	}
+	req := &PutObjectSignedURLRequest{
+		Object:      info["object"].(string),
```
vsukhin, line 55: possible panic, if we change CheckStringKey method

pkg/cloud/data/scraper/loader.go
```diff
@@ -0,0 +1,103 @@
...
+		ExecutionID: info["executionId"].(string),
+		TestName:    info["testName"].(string),
+	}
+
+	if info["testSuiteName"] != nil {
+		if s, ok := info["testSuiteName"].(string); ok {
+			req.TestSuiteName = s
+		}
+	}
```
vsukhin, line 64: guess, can be reused above method

pkg/executor/scraper/filesystem_extractor.go
dejanzele: I wanted to build generic Extractor interface which should in theory be able to extract data from anywhere.
In our case, we only walk directories so we only have 1 implementation and we deal with files there.

pkg/executor/scraper/loader.go
dejanzele: Or that, yeah, it extracts then uploads

pkg/executor/scraper/scraper.go
dejanzele: keep in mind I need to refactor every single executoro

PR #3327, opened 2023-03-03
fix: update default body size limit for http client

vLia:
## Pull request description 

Complaint coming from user on Discord that uploading a file of size 250MB, the api server is returning: 413

image

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/internal/app/api/v1/server.go
+++ b/internal/app/api/v1/server.go
@@ -45,7 +45,10 @@ import (
 	"github.com/kubeshop/testkube/pkg/utils/text"
 )
 
-const HeartbeatInterval = time.Hour
+const (
+	HeartbeatInterval = time.Hour
+	HttpBodyLimit     = 1 * 1024 * 1024 * 1024 // 1GB - needed for file uploads
+)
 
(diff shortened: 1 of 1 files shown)


internal/app/api/v1/server.go
```diff
@@ -45,7 +45,10 @@ import (
 	"github.com/kubeshop/testkube/pkg/utils/text"
 )
 
-const HeartbeatInterval = time.Hour
+const (
+	HeartbeatInterval = time.Hour
+	HttpBodyLimit     = 1 * 1024 * 1024 * 1024 // 1GB - needed for file uploads
```
vsukhin, line 50: need to have helm value for this

vsukhin (approved):

internal/app/api/v1/server.go
vLia: I will create a follow-up ticket, just want to fix the issue of the user for now

PR #3378, opened 2023-03-08
feature: refactor main to support loading templates from files

dejanzele:
## Pull request description 

Refactor main to support loading config from files.
If a config file does not exists, fallback is to load from env var.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -42,8 +42,6 @@ import (
 	"github.com/kubeshop/testkube/pkg/k8sclient"
 	"github.com/kubeshop/testkube/pkg/triggers"
 
-	"github.com/kelseyhightower/envconfig"
-
 	kubeclient "github.com/kubeshop/testkube-operator/client"
 	executorsclientv1 "github.com/kubeshop/testkube-operator/client/executors/v1"
 	scriptsclient "github.com/kubeshop/testkube-operator/client/scripts/v2"
@@ -60,22 +58,11 @@ import (
 	"github.com/kubeshop/testkube/pkg/ui"
 )
 
-type MongoConfig struct {
-	DSN          string `envconfig:"API_MONGO_DSN" default:"mongodb://localhost:27017"`
-	DB           string `envconfig:"API_MONGO_DB" default:"testkube"`
-	SSLSecretRef string `envconfig:"API_MONGO_SSL_CERT"`
-	AllowDiskUse bool   `envconfig:"API_MONGO_ALLOW_DISK_USE"`
-}
-
-var Config MongoConfig
-
 var verbose = flag.Bool("v", false, "enable verbosity level")
 
 func init() {
 	flag.Parse()
 	ui.Verbose = *verbose
-	err := envconfig.Process("mongo", &Config)
-	ui.PrintOnError("Processing mongo environment config", err)
 }
 
 func runMigrations() (err error) {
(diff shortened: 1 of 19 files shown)


cmd/api-server/main.go
```diff
@@ -192,10 +193,10 @@ func main() {
 	}
 
 	if clusterId == "" {
-		cmConfig, err = configRepository.Get(ctx)
+		cmConfig, _ = configRepository.Get(ctx)
```
vsukhin, line 196: why do hide error there?

pkg/executor/common.go
```diff
@@ -306,6 +309,78 @@ func AbortJob(ctx context.Context, c kubernetes.Interface, namespace string, job
...
+	} else if cfg.TestkubeContainerTemplatePVC != "" {
+		decoded, err = base64.StdEncoding.DecodeString(cfg.TestkubeContainerTemplatePVC)
+		if err != nil {
+			return t, err
+		}
+		t.PVC = string(decoded)
+	}
+
+	return t, nil
+}
```
vsukhin, line 382: Guess, you can do it similar to what I did for crd templates. They are embedded into binary, no need to have a separted dir with templates and copy it

cmd/api-server/main.go
dejanzele: it wasn't checked before, I just hide it because staticcheck was linting it for not being checked.

SHould it be checked here?

pkg/executor/common.go
```diff
@@ -327,22 +402,38 @@ func NewTemplatesFromEnv(env string) (t Templates, err error) {
...
-// SyncDefaultExecutors creates or updates default executors
-func SyncDefaultExecutors(executorsClient executorsclientv1.Interface, namespace, data string, readOnlyExecutors bool) (
-	images Images, err error) {
-	var executors []testkube.ExecutorDetails
+func ParseExecutors(cfg *config.Config) (executors []testkube.ExecutorDetails, err error) {
+	var data []byte
 
-	if data == "" {
-		return images, nil
+	f, err := os.Open(filepath.Join(cfg.TestkubeConfigDir, "executors.json"))
```
vsukhin, line 408: executors.json is dynamically updated in helm-chart for new executor relase.how are going to sync them?

dejanzele: the `config` folder in the testkube serves as example config and for localdev.

idea is to use a ConfigMap and define executors.json, templates, etc etc... in the ConfigMap and mount it in the Testkube API Server pod.

when there is a change we will update it in the Helm chart, we could probably just use the Helm file function to add it in the ConfigMap, so we keep the already existing files

cmd/api-server/main.go
vsukhin: Might be log for debug mode

pkg/executor/common.go
dejanzele: I don't fully know what is dynamic and what not.

Are the templates something that doesn't change often?

//go:embed can also be used to embed a file into a string

pkg/executor/common.go
vsukhin: ok, then makes sense

pkg/executor/common.go
vsukhin: executors.json is dynamic, other ones are not changes often. but if we want to mount config map,then we don't need embedded

nicufk (approved):

PR #3438, opened 2023-03-17
fix: multinamespace watchers

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


nicufk (approved):

internal/config/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	TestkubeCloudAPIKey              string `envconfig:"TESTKUBE_CLOUD_API_KEY" default:""`
 	TestkubeCloudURL                 string `envconfig:"TESTKUBE_CLOUD_URL" default:""`
 	TestkubeCloudTLSInsecure         bool   `envconfig:"TESTKUBE_CLOUD_TLS_INSECURE" default:"false"`
+	TestkubeWatcherNamespaces        string `envconfig:"TESTKUBE_WATCHER_NAMESPACES" default:""`
```
dejanzele, line 45: envconfig supports []string if TESTKUBE_WATCHER_NAMESPACES is provided as a comma-separated string (default should be changed to null in that case also)

PR #3481, opened 2023-03-21
add bool flag to modify git clone args for bitbucket server auth

zaljic:
## Pull request description 
By merging this PR, the API gets the functionality to inject a `git clone` command with the `http.extraHeader` arg set. This allows authentication with HTTP Access Tokesn on Bitbucket Server repositories. 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- Updated OpenAPI definition
- Modified `Checkout` and `Fetch` functions to check if `isBitbucketServer` is set to `true`
- Added `isBitbucketServer` flag to CLI

## Fixes

-

### DIFF
--- a/api/v1/testkube.yaml
+++ b/api/v1/testkube.yaml
@@ -3474,6 +3474,10 @@ components:
         token:
           type: string
           description: git auth token for private repositories
+        isBitbucketServerToken:
+          type: boolean
+          description: if true git clone will be executed with the http.extraHeader arg set to the value specified in token
+          example: "false"
         usernameSecret:
           $ref: "#/components/schemas/SecretRef"
         tokenSecret:
--- a/pkg/executor/content/fetcher.go
+++ b/pkg/executor/content/fetcher.go
@@ -91,7 +91,7 @@ func (f Fetcher) FetchGitDir(repo *testkube.Repository) (path string, err error)
 
 	// if path not set make full repo checkout
 	if repo.Path == "" || repo.WorkingDir != "" {
-		path, err := git.Checkout(uri, repo.Branch, repo.Commit, f.path)
+		path, err := git.Checkout(uri, repo.Branch, repo.Commit, repo.Token, f.path, repo.IsBitbucketServerToken)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


api/v1/testkube.yaml
```diff
@@ -3474,6 +3474,10 @@ components:
         token:
           type: string
           description: git auth token for private repositories
+        isBitbucketServerToken:
```
vsukhin, line 3477: let's call isHeaderToken, because it might be useful for other git providetrs

vsukhin (commented):
in general looks,left a couple of comments,but the changes are more complex:
1. We need to store flag value in Test CRD, so check testkube repo for `Branch` field and you will find,that it's used in different places
2. We will also need to update testkube-operator and helm-chart repos, because they define the CRD model for storing data
3. We usually test everyting in a local cluster for changed version

pkg/executor/content/fetcher.go
```diff
@@ -91,7 +91,7 @@ func (f Fetcher) FetchGitDir(repo *testkube.Repository) (path string, err error)
 
 	// if path not set make full repo checkout
 	if repo.Path == "" || repo.WorkingDir != "" {
-		path, err := git.Checkout(uri, repo.Branch, repo.Commit, f.path)
+		path, err := git.Checkout(uri, repo.Branch, repo.Commit, repo.Token, f.path, repo.IsBitbucketServerToken)
```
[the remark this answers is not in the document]
zaljic: Done. `repo.IsBitbucketServertoken` (now `repo.IsHeaderToken`) is a bool to determine whether a modification of `git clone` is necessary or not. My idea is that there will be a button in the UI, which sends the `isHeaderToken: true` value to the testkube backend. Do you think the way it is implemented now is the way to go?

api/v1/testkube.yaml
zaljic: Done

zaljic:
Created PR in testkube-operator:
https://github.com/kubeshop/testkube-operator/pull/125

PR #3513, opened 2023-03-25
feat: basic graphql subscription for executor

exu:
## Pull request description 

First implementation of GraphQL subscriptions API 

list will be reloaded automatically on create/delete/update event from NATS 

Still (probably) a lot to do 

https://user-images.githubusercontent.com/30776/227668733-4229f26e-9e7e-420e-bb01-43f294100985.mov

Limits: 
* GraphQL API can't be attached to the Fiber router, as there are issues with adaptor not handling websockets correctly. (look at TODO comments in code - there is pending issue which may be resolved in future) 

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vsukhin:
Jacek is on fire at nights :)

internal/graphql/server.go
```diff
@@ -0,0 +1,57 @@
...
+// 	executorsClient := executorsclientv1.NewClient(kubeClient, "testkube")
+
+// 	srv := GetGraphQLServer(eventBus, executorsClient)
+
+// 	http.Handle("/", playground.Handler("GraphQL playground", "/query"))
+// 	http.Handle("/query", srv)
+
+// 	log.DefaultLogger.Infof("connect to http://localhost:%s/ for GraphQL playground", port)
+// 	log.DefaultLogger.Panic(http.ListenAndServe(":"+port, nil))
+// }
```
vsukhin, line 43: do we need this commented code?

internal/graphql/graph/schema.resolvers.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// TODO You can (and probably should) handle your channels in a central place outside of `schema.resolvers.go`.
+	// For this example we'll simply use a Goroutine with a simple loop.
+	go func() {
+		r.Log.Infof("%+v\n", "subscribed to events.executor.>")
+
+		r.Bus.SubscribeTopic("events.executor.>", rand.String(30), func(e testkube.Event) error {
+			r.Log.Infof("%s %s %s\n", e.Type_, *e.Resource, e.ResourceId)
+
+			execs, err := r.Client.List("")
```
vsukhin, line 64: will we send there always all items regardless of filtering?

vsukhin (approved):
looks good, some minor questions

internal/app/api/v1/graphql.go
```diff
@@ -0,0 +1,28 @@
...
+
+// RunGraphQLServer runs GraphQL server on go net/http server
+// There is an issue with gqlgen and fasthttp server
+func (a *TestkubeAPI) RunGraphQLServer(
+	ctx context.Context,
+	cfg *config.Config,
+) error {
+	srv := graphql.GetServer(a.Events.Bus, a.ExecutorsClient)
+
+	http.Handle("/", playground.Handler("GraphQL playground", "/query"))
```
rangoo94, line 22: `/graphql` endpoints are rather more standard

internal/app/api/v1/graphql.go
```diff
@@ -0,0 +1,28 @@
...
+	http.Handle("/query", srv)
+
+	log.DefaultLogger.Infow("running GraphQL server", "port", cfg.GraphqlPort)
+
+	return http.ListenAndServe(":"+cfg.GraphqlPort, nil)
```
rangoo94, line 27: If we want to use it on the front end, the GraphQL server should be exposed under the same port - either directly or proxied.

Otherwise, we would need to pass two URLs to the front-end - `apiEndpoint` and `gqlEndpoint`.

exu: @rangoo94 it's not possible on this level, as current GraphQL implementation in go doesn't handle FastHTTP well .

internal/graphql/graph/schema.resolvers.go
exu: probably not? don't know yet, we should define what params are needed with @rangoo94 and @pavloburchak for now whole list (for executors there is no filtering??)
image

testCoverage.txt
```diff
@@ -0,0 +1,5644 @@
+mode: set
```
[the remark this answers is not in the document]
exu: nope some accidental commit and push - removed

internal/graphql/graph/schema.resolvers.go
```diff
@@ -0,0 +1,118 @@
...
+// Executors is the resolver for the executors field.
+func (r *subscriptionResolver) Executors(ctx context.Context) (<-chan []testkube.ExecutorDetails, error) {
+	ch := make(chan []testkube.ExecutorDetails)
+
+	// TODO You can (and probably should) handle your channels in a central place outside of `schema.resolvers.go`.
+	// For this example we'll simply use a Goroutine with a simple loop.
+	go func() {
+		r.Log.Infof("%+v\n", "subscribed to events.executor.>")
+
+		r.Bus.SubscribeTopic("events.executor.>", rand.String(30), func(e testkube.Event) error {
```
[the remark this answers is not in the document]
exu: Is this how it should be used?

internal/app/api/v1/graphql.go
exu: So we need to split it, but I think it should be possible to merge it on ingress level (ar least I hope so ;)

internal/graphql/graph/schema.resolvers.go
rangoo94: We do not have any filtering yet, and I don't think that someone will have so many custom executors that it will be a problem. I would say that it may be easily postponed unless you would like to experiment with GraphQL variables here.

internal/graphql/graph/schema.resolvers.go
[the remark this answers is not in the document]
rangoo94: **TLDR:** it is often used this way and it's rather perfectly suited for our needs.

The most natural choice for subscriptions is to send events (like `onExecutorsListChange` in this case). On the opposite, live queries are just regular queries, but with real-time (more/less) updates.

Live queries are not a part of GraphQL specification and are implemented in different ways. The most distinctive way is to add `@live` directive to the query, but it's not necessarily supported (i.e. `gqlgen` [**doesn't support it**](https://github.com/99designs/gqlgen/issues/2552)). Because of all the problems, a lot of tools, like [**Hasura**](https://hasura.io/docs/latest/subscriptions/postgres/livequery/index/) are using subscriptions this way (so basically replace `query` with `subscription` to obtain data).

There is [**nice article**](https://the-guild.dev/blog/subscriptions-and-live-queries-real-time-with-graphql) about this topic.

As an alternative solution, you can `query` the data before, and add `subscription` later for changes. This solution has multiple problems though:
* it's more complex
* you need to handle data from the time between the query and subscription started (i.e. querying again or providing the last update date for the subscription as param)

PR #3626, opened 2023-04-06
fix: improve execution logs

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/executor/client/job.go
```diff
@@ -153,7 +153,7 @@ func (c *JobExecutor) Logs(ctx context.Context, id string) (out chan output.Outp
 
 		for l := range logs {
 			entry, err := output.GetLogEntry(l)
-			if err != nil {
+			if err != nil && entry.Type_ != output.TypeUnknown {
```
vsukhin, line 156: will we have not nil entry there always?

pkg/process/exec.go
```diff
@@ -25,12 +25,12 @@ func ExecuteInDir(dir string, command string, arguments ...string) (out []byte,
 	cmd.Stderr = buffer
 
 	if err = cmd.Start(); err != nil {
-		return buffer.Bytes(), fmt.Errorf("could not start process: %w", err)
+		return buffer.Bytes(), fmt.Errorf("could not start process with command: %s, exited with code:%d, error: %w\noutput: %s", command, cmd.ProcessState.ExitCode(), err, buffer.String())
```
vsukhin, line 28: is ProcessState always not nill?

vsukhin (approved):
looks good have two minor questions

pkg/executor/client/job.go
nicufk: GetLogEntry is returning a testkube.ExecutorOutput type not a pointer.

Issue #3636, opened 2023-04-07
API Server is not closing after the SIGTERM/SIGINT signals


rangoo94:
**Describe the bug**

When sending the signal, the application is not closed, because the GraphQL server is not closing.

**To Reproduce**

1. Start the API in terminal
2. Type Ctrl+C
3. The application is still running

**Expected behavior**

The application should gracefully close.

PR #3637, opened 2023-04-07
fix: Close API server gracefully when signal appeared

rangoo94:
## Pull request description 

* Close the GraphQL server gracefully when the app should be terminated
* Don't wait 2 seconds for closing the API servers, instead separate the listener and serve to avoid problems

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

- https://github.com/kubeshop/testkube/issues/3636

### DIFF
--- a/pkg/server/httpserver.go
+++ b/pkg/server/httpserver.go
@@ -3,13 +3,12 @@ package server
 import (
 	"context"
 	"encoding/json"
-	"time"
-
 	"github.com/gofiber/adaptor/v2"
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/pprof"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


pkg/server/httpserver.go
```diff
@@ -3,13 +3,12 @@ package server
...
 	"context"
 	"encoding/json"
-	"time"
-
 	"github.com/gofiber/adaptor/v2"
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/pprof"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.uber.org/zap"
+	"net"
```
vsukhin, line 11: move it to std lib group

vsukhin (approved):
linting comments

Issue #3116, opened 2023-01-30
MongoDB SSL connection to AWS DocDB


vLia:
**Is your feature request related to a problem? Please describe.**
[Discord discussion](https://discord.com/channels/884464549347074049/1068317088147574864)
AWS provides a bundle of CA certificates. The current implementation does not support this.

**Describe the solution you'd like**
Certificate setting should be updated to support this use case.
See details: https://docs.aws.amazon.com/documentdb/latest/developerguide/connect_programmatically.html#connect_programmatically-tls_enabled

vLia:
Implementation released in Testkube v1.10.43

PR #3657, opened 2023-04-11
feat: AWS DocDB connection

vLia:
## Pull request description 

Fix without docs to https://github.com/kubeshop/testkube/issues/3116

This PR adds code to automatically download the pem files for aws docdb. There is no way to configure client certificates, is uses user/pass only.

- [ ] tested locally
- [x] tested on cluster - created cluster and docdb on aws and connected them
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/repository/storage/mongo.go
+++ b/pkg/repository/storage/mongo.go
@@ -3,7 +3,13 @@ package storage
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
+	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
 	"time"
 
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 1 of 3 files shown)


pkg/repository/storage/mongo.go
```diff
@@ -20,19 +26,37 @@ type MongoSSLConfig struct {
...
+				"sslClientCertificateKeyFile":     certConfig.SSLClientCertificateKeyFile,
+				"sslClientCertificateKeyPassword": certConfig.SSLClientCertificateKeyFilePassword,
+				"sslCertificateAuthorityFile":     certConfig.SSLCertificateAuthoritiyFile,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("could not build SSL config: %w", err)
+			}
+		}
+	case TypeDocDB:
+		mongoOptions, err = getDocDBTLSConfig()
```
vsukhin, line 52: is this mandatory option? because some users already use DocDB without tls?

pkg/repository/storage/mongo.go
```diff
@@ -42,3 +66,50 @@ func GetMongoDatabase(dsn, name string, certConfig *MongoSSLConfig) (db *mongo.D
...
 }
+
+func getDocDBTLSConfig() (*tls.Config, error) {
+	caFilePath, err := GetDocDBcaFile()
+	if err != nil {
+		return nil, fmt.Errorf("could not get CA file: %w", err)
+	}
+
+	tlsConfig := new(tls.Config)
+	certs, err := ioutil.ReadFile(caFilePath)
```
vsukhin, line 77: do we clean this file after usage?

pkg/repository/storage/mongo.go
vsukhin: Or they should use MongoDb in this case?

pkg/repository/storage/mongo.go
```diff
@@ -42,3 +66,50 @@ func GetMongoDatabase(dsn, name string, certConfig *MongoSSLConfig) (db *mongo.D
...
+// Due to size limitations we cannot use Kubernetes secrets like we use for MongoDB TLS configs
+func GetDocDBcaFile() (string, error) {
+	// Get the data
+	resp, err := http.Get(DocDBcaFileURI)
+	if err != nil {
+		return "", fmt.Errorf("could not fetch file from %s: %w", DocDBcaFileURI, err)
+	}
+	defer resp.Body.Close()
+
+	docDBcaPath := "/tmp/rds-combined-ca-bundle.pem"
```
vsukhin, line 103: shouldn't it be in some random temp subdir?

vsukhin (approved):
looks good, a few comments

pkg/repository/storage/mongo.go
vLia: it would work if they would not set a db type, but I'll add a field to make this clear

PR #3669, opened 2023-04-12
feat: Clean up GraphQL Server

rangoo94:
## Pull request description 

This PR is meant to prepare GraphQL code architecture, that will be:

* Easy to maintain (because the structure is obvious, the code is split nicely, and less custom code is needed)
* Testable (as it's easy to write unit tests for every layer)
* Simple to extend (using the available helpers, it should be easy to create a new queries and subscriptions upon that)

> Don't worry that much about the PR size:
> 
> * `+267, -90` is `generated.go`
> * `+66, -0` is generated `mock_executors.go`
> * `+328, -0` are unit tests
> * `+80, -0` are test utilities
> * `+91, -82` are schema
> * `+10, 0` is config
>
> Summing up, there are 257 lines of actual code added.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

- None

## Changes

- Split GraphQL schema files per functionality (root & executors)
- Adjust GraphQL schema to match actual objects
- Add `StringMap` scalar (`map[string]string`), so there is less resolvers needed
- Extract business logic from GraphQL resolvers to "Services"
- Add example unit tests for both Executors resolvers and service
  - Includes also helpers for testing
- chore: Ignore false-positive errors in proxy connection

## Fixes

- https://github.com/kubeshop/testkube/issues/3676


internal/graphql/scalars/maps.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/99designs/gqlgen/graphql"
+)
+
+// Generic implementation
+
+func MarshalAnyMapScalar[T interface{}](val map[string]T) graphql.Marshaler {
+	return graphql.WriterFunc(func(w io.Writer) {
+		err := json.NewEncoder(w).Encode(val)
+		if err != nil {
+			panic(err)
```
vsukhin, line 17: Please remove panic, log error and return

internal/graphql/services/common.go
```diff
@@ -0,0 +1,61 @@
...
+) (<-chan U, error) {
+	ch := make(chan U, 1)
+
+	// Load initial data
+	initial, err := get()
+	if err == nil {
+		ch <- initial
+	} else {
+		s.Logger().Errorw("failed to get initial data for "+topic, err)
+		return nil, err
```
vsukhin, line 32: should we close ch there?

internal/graphql/services/common.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if err == nil {
+		s.Logger().Debug("graphql subscription: subscribed to " + topic)
+		go func() {
+			<-ctx.Done()
+			_ = s.Bus().Unsubscribe(queue)
+			close(ch)
+		}()
+	} else {
+		s.Logger().Errorw("graphql subscription: failed to subscribe to "+topic, err)
```
vsukhin, line 56: should we close ch?

vsukhin (approved):
looks good, remove panic, couple of minor issues

internal/graphql/services/common.go
rangoo94: The channel is not returned, so it will be garbage collected.

internal/graphql/services/common.go
```diff
@@ -0,0 +1,61 @@
...
+	initial, err := get()
+	if err == nil {
+		ch <- initial
+	} else {
+		s.Logger().Errorw("failed to get initial data for "+topic, err)
+		return nil, err
+	}
+
+	// Setup queue
+	queue := rand.String(30)
```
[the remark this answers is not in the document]
rangoo94: It's not, for every subscription there is created exclusive queue.

internal/graphql/services/common.go
rangoo94: The channel is not returned, so it will be garbage collected.

internal/graphql/schemas/executors.graphqls
```diff
@@ -0,0 +1,82 @@
...
+    executors(selector: String! = ""): [ExecutorDetails!]!
+}
+
+type ExecutorDetails {
+    name: String!
+    executor: Executor!
+}
+
+"""CRD based executor data"""
+type Executor {
```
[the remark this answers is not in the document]
rangoo94: We could consider that for the future for the subset of types, but we will have differences in some types.

As an example, I believe that in the GraphQL query/subscription for tests/test suites, we will have nested resolvers for metrics - that's not happening in the REST API, and probably we will not have it here.

PR #3890, opened 2023-05-24
feat: condition ttl

### LINKED ISSUE #3740: Add trigger that waits until all pods are new and ready
I'd like my test to be triggered after I have changed e.g. an image of a deployment and all pods of that deployment are new and ready (so all pods are replaced with the new version -> the deployment it done and finished successfully).

I've seen #2905 and #3092 and setup a trigger as follows:
```yaml
apiVersion: tests.testkube.io/v1
kind: TestTrigger
metadata:
  name: test-trigger-1
  namespace: testkube
spec:
  action: run
  event: deployment-image-update
  execution: test
  resource: deployment
  resourceSelector:
    labelSelector:
      matchLabels:
        app: exampleapp
  testSelector:
    name: curl-test-1
  delay: 10s
  conditionSpec:
    timeout: 60
    conditions:
      - type: Progressing
        status: "True"
        reason: "NewReplicaSetAvailable"
      - type: Available
        status: "True"
```
I've also seen the possibility to add a delay. 
But when I understand it correctly (and it also seems that way from testing) the delay is applied after the trigger has been matched and just delays the test execution itself.

I guess my observation is that the status of my deployment starts with this status (when updating the image)
```
  "reason": "NewReplicaSetAvailable",
  "status": "True",
  "type": "Progressing"
```
and thus directly triggers the (delayed, when setup with a `delay`) test.

My example application which is tested by the curl test starts slower than the 10 seconds delay, so that the test currently tests the old version. I would not like to rely on some hardcoded delay if possible (just increasing the delay to for example 60).

Am I missing something to get the desired behavior by changing configuration?
If not: Could that behavior be implemented / archived in any way?

**Additional context**

*(Just became aware of this awesome project at kubecon :) 👍 )*

vsukhin:
## Pull request description 

Fix #3740

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vLia (approved):
Looks good 👍

PR #4033, opened 2023-06-14
feat: validate cloud context on root cmd

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/root.go
+++ b/cmd/kubectl-testkube/commands/root.go
@@ -64,12 +64,15 @@ var RootCmd = &cobra.Command{
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		ui.SetVerbose(verbose)
 
+		cfg, err := config.Load()
+		ui.ExitOnError("loading config", err)
+
+		common.UiContextHeader(cmd, cfg)
+
+		// don't validate context before set
 		if cmd.Name() == "context" {
 			return
 		}
-
-		cfg, err := config.Load()
-		ui.ExitOnError("loading config", err)
 
 		if err = validator.ValidateCloudContext(cfg); err != nil {
 			common.UiCloudContextValidationError(err)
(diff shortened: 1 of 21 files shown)


cmd/kubectl-testkube/commands/root.go
```diff
@@ -62,6 +63,17 @@ var RootCmd = &cobra.Command{
...
+
+		if cmd.Name() == "context" {
+			return
+		}
+
+		cfg, err := config.Load()
+		ui.ExitOnError("loading config", err)
+
+		if err = validator.ValidateCloudContext(cfg); err != nil {
+			common.UiCloudContextValidationError(err)
```
vsukhin, line 75: cli should work without os error code for missed config, like PostRun

vsukhin (commented):
one important comment

cmd/kubectl-testkube/commands/root.go
exu: ok removed it

cmd/kubectl-testkube/commands/root.go
```diff
@@ -62,6 +63,20 @@ var RootCmd = &cobra.Command{
 	Short: "Testkube entrypoint for kubectl plugin",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		ui.SetVerbose(verbose)
+
+		cfg, err := config.Load()
+		ui.ExitOnError("loading config", err)
```
vsukhin, line 68: I would use WarnOnError instead

vsukhin, line 68: will it run without kubernetes cluster and ability to load/store testkube config?

vsukhin (approved):

PR #4119, opened 2023-07-03
fix: escape and unescape dots

### LINKED ISSUE #4116: Error starting tests managed by Flux
**Describe the bug**
I see the below error on the Dashboard when running a test managed by Flux. I have a Git Repo hosting the Tests, Test Suites, Sources, and Executors CRDs and have Flux reconcile and apply them. 
```
failed to execute test: execution failed: can't create new test execution, can't insert into storage: write exception: write errors: [Document can't have '.' in field names: kustomize.toolkit.fluxcd.io/name], context: null
```

**Version / Cluster**
- Which testkube version? 1.12.11
- What Kubernetes cluster? AWS EKS
- What Kubernetes version? 1.24
 
**Additional context**
The same tests run without any errors if they are not managed by Flux and manually applied to the cluster. I use AWS DocumentDB 5.0 engine version.

vsukhin:
## Pull request description 

Fix #4116 

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/api/v1/testkube/model_test_suite_execution_extended.go
+++ b/pkg/api/v1/testkube/model_test_suite_execution_extended.go
@@ -213,3 +213,32 @@ func (e *TestSuiteExecution) IsAborted() bool {
 func (e *TestSuiteExecution) IsTimeout() bool {
 	return *e.Status == TIMEOUT_TestSuiteExecutionStatus
 }
+
+func (e *TestSuiteExecution) convertDots(fn func(string) string) TestSuiteExecution {
+	labels := make(map[string]string, len(e.Labels))
+	for key, value := range e.Labels {
+		labels[fn(key)] = value
+	}
+	e.Labels = labels
+
+	envs := make(map[string]string, len(e.Envs))
+	for key, value := range e.Envs {
+		envs[fn(key)] = value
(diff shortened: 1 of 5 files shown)


pkg/api/v1/testkube/model_test_suite_execution_extended.go
```diff
@@ -213,3 +213,32 @@ func (e *TestSuiteExecution) IsAborted() bool {
 func (e *TestSuiteExecution) IsTimeout() bool {
 	return *e.Status == TIMEOUT_TestSuiteExecutionStatus
 }
+
+func (e *TestSuiteExecution) convertDots(fn func(string) string) TestSuiteExecution {
```
rangoo94, line 217: Shouldn't it either:

* Modify the existing execution (as it does) and return its pointer (as it doesn't), or
* Don't modify the existing execution (as it doesn't) and return new object (as it does)
?

vsukhin: yes, good point, should be first

PR #4184, opened 2023-07-17
feat: dashboard-e2e integration CRD

tkonieczny:
CRD for integration cluster (namespace)


tkonieczny:
Closing - separate tk run test with variable overwrite will be used instead

Issue #3801, opened 2023-05-09
Websocket no longer streaming log lines


donaldoakes:
**Describe the bug**
Previous api server version (specifically 1.10.16) streamed individual log lines at e.g.: `ws://localhost:8088/v1/executions/645a806aa995da0b1de187c2/logs/stream`. After upgrade to 1.11.18, individual log lines are no longer streamed; only `result` type entries. Before and after screenshots attached for [postman-test](https://docs.testkube.io/test-types/executor-postman#b-creating-a-test-with-git-file).

**To Reproduce**
1. Run `kubectl testkube dashboard`
2. Add postman-test as per: https://docs.testkube.io/test-types/executor-postman#b-creating-a-test-with-git-file
3. Run postman-test
4. Observe websocket streamed messages no longer include log lines.

**Expected behavior**
Log line should be included in /v1/executions/{executionId}/logs/stream.

**Version / Cluster**
- Which testkube version?  1.11.18
- What Kubernetes cluster? Docker Desktop MacOS
- What Kubernetes version? v1.25.4

**Screenshots**
**1.10.16**  
<img width="2032" alt="testkube-1 10 16-postman" src="https://github.com/kubeshop/testkube/assets/19718743/7a9397d6-408e-4245-8f3d-d45f0c2954c2">
**1.11.18**  
<img width="2032" alt="testkube-1 11 18-postman" src="https://github.com/kubeshop/testkube/assets/19718743/6ee593dc-82e2-4d56-9704-43a04f1ecdd2">

donaldoakes:
This seems to be an issue with the server, and not with the UI. Logs are not being streamed to the websocket listener. Tested with postman-test and with curl-test.

nicufk:
Hey @donaldoakes I tried to reproduce it on my side and it is working fine with the latest changes, I checked in our demo cluster as well and it is working as you can see [here](https://demo.testkube.io/tests/executions/postman-executor-smoke-testsource-git-file/execution/645b61907ccc9872354e59de) and the version of the testkube is https://demo.testkube.io/results/v1/info.

There might be something else that is affecting logs on your side

nicufk:
@donaldoakes, yes I see, I will investigate it

TheBrunoLopes:
@nicufk seems like the PR is open https://github.com/kubeshop/testkube/pull/4212
should we merge it ?

nicufk:
> @nicufk seems like the PR is open #4212 should we merge it ?

Merged!

[...]

PR #4212, opened 2023-07-27
fix: streaming logs not working

nicufk:
## Pull request description 
Closes https://github.com/kubeshop/testkube/issues/3801

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/client/job.go
+++ b/pkg/executor/client/job.go
@@ -542,9 +542,6 @@ func (c *JobExecutor) TailJobLogs(ctx context.Context, id string, logs chan []by
 				if err = wait.PollImmediate(pollInterval, c.podStartTimeout, executor.IsPodLoggable(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
 					l.Errorw("poll immediate error when tailing logs", "error", err)
 					return err
-				} else if err = wait.PollImmediate(pollInterval, pollTimeout, executor.IsPodReady(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
-					l.Errorw("poll immediate error when tailing logs", "error", err)
-					return c.GetLastLogLineError(ctx, pod)
 				}
 
 				l.Debug("tailing pod logs")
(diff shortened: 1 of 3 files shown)


pkg/executor/client/job.go
```diff
@@ -542,9 +542,6 @@ func (c *JobExecutor) TailJobLogs(ctx context.Context, id string, logs chan []by
 				if err = wait.PollImmediate(pollInterval, c.podStartTimeout, executor.IsPodLoggable(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
 					l.Errorw("poll immediate error when tailing logs", "error", err)
 					return err
-				} else if err = wait.PollImmediate(pollInterval, pollTimeout, executor.IsPodReady(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
-					l.Errorw("poll immediate error when tailing logs", "error", err)
```
vsukhin, line 546: Why are we removing it?

nicufk: the logic is tricky PollImediate will return true when the func IsPodLoggable returns true but the error is nil so it will go to the next PollImmediate, IsPodReady is waiting for corev1.PodSucceeded and it will not let the streaming start.

vsukhin: 1. We have this logic in 2 places, if we want to change it
2. Second check was added to support cases of wrong images or not reachable at the moment, because in this case we simply doesn't have any errors and logs. So need to find a way to work properly for both things

pkg/executor/client/job.go
```diff
@@ -542,9 +540,6 @@ func (c *JobExecutor) TailJobLogs(ctx context.Context, id string, logs chan []by
 				if err = wait.PollImmediate(pollInterval, c.podStartTimeout, executor.IsPodLoggable(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
 					l.Errorw("poll immediate error when tailing logs", "error", err)
 					return err
-				} else if err = wait.PollImmediate(pollInterval, pollTimeout, executor.IsPodReady(ctx, c.ClientSet, pod.Name, c.Namespace)); err != nil {
```
vsukhin, line 545: shouldn't we return `return c.GetLastLogLineError(ctx, pod)`?

vsukhin (approved):

PR #4213, opened 2023-07-27
fix: Playwright custom container executor - CI=1

tkonieczny:


vsukhin (approved):

rangoo94:
As per comment in https://github.com/kubeshop/testkube-dashboard/pull/797:

> We don't use it in other executors, therefore we should rather either consistently use it everywhere, or don't use it anywhere.
> I would suggest adding `CI=1` automatically on the job level (for all jobs), not the executor or test level.

tkonieczny:
I created a separate issue for that: https://github.com/kubeshop/testkube/issues/4214

nicufk (approved):

PR #4230, opened 2023-08-01
fix: install script

### LINKED ISSUE #4229: Install script is broken: due to invalid parameter substitution syntax
**Describe the bug**

* PR #4071 introduced a [parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) that contains a typo, making the syntax invalid in POSIX shell (missing the `-` dash after the `:` column)

```diff
-  echo "https://github.com/kubeshop/testkube/releases/download/${version}/testkube_${version:1}_${os}_$arch.tar.gz"
+  echo "https://github.com/kubeshop/testkube/releases/download/${version}/testkube_${version:-1}_${os}_$arch.tar.gz"
```

* Download URL need both tag and version which differ by a prefix.

```
❯ sh ./install.sh
…
./install.sh: 46: Bad substitution
```

**To Reproduce**

```
❯ sh
$ foo="123"
$ echo "bar=${foo}"
bar=123
$ echo "bar=${foo:1}"  
sh: 3: Bad substitution
```

**Expected behavior**

The parameter subsitution should happen

**Version / Cluster**
- Which testkube version?  699e66cb9
- ~What Kubernetes cluster?~ irrelevant
- ~What Kubernetes version?~ irrelevant

**Screenshots**
If applicable, add CLI commands/output to help explain your problem.
 
**Additional context**

edouard-lopez:
related: #4071, fixes #4229

### Usage

```
❯ sudo bash ./install.sh
```

## Details

- fix: switch shebang to reflect existing code and move to the top
- chore: declare and assign separately to avoid masking return values.
- refactor: split to multiline so it simpler to read and maintain
- fix(install): correct typo in parameter expansion for default value
- fix: add a `tag` variable as download URL require `tag` and `version` (which differ)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,5 @@
+#!/bin/bash 
 echo "Getting kubectl-testkube plugin"
-#!/bin/sh 
 
 if [ ! -z "${DEBUG}" ]; 
 then set -x 
--- a/install.sh
+++ b/install.sh
@@ -56,7 +56,7 @@ _download_url() {
     version="$TESTKUBE_VERSION"
   fi
 
-  echo "https://github.com/kubeshop/testkube/releases/download/${version}/testkube_${version:1}_${os}_$arch.tar.gz"
+  echo "https://github.com/kubeshop/testkube/releases/download/${version}/testkube_${version:-1}_${os}_$arch.tar.gz"
 }
 
 if [ "$1" = "beta" ]; then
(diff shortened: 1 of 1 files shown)


vsukhin (approved):
thank you @edouard-lopez looks good to me. @ypoplavs please check it out

ypoplavs:
Tested on MacOs and Linux, works well.

PR #4368, opened 2023-09-14
feat: concurrency policy support

### LINKED ISSUE #4361: Multiple event not properly handled by testtrigger
**Describe the bug**
A test trigger cannot correctly handle matching events:
- When the first matching event appears, tests associated with the trigger are executed as expected
- If the first event is deleted and an identical one is created, no new tests are triggered after the event appearance
- Only after each test associated with the trigger end, the events are correctly handled again
Once a matching event for a trigger is created

**To Reproduce**
1. Create a trigger matching a specific event creation
2. Label some tests to match the trigger labels
3. Create the matching in k8s (for example with kubectl apply -f)
4. Observe the labeled test starting
5. Delete the event
6. While the tests are still running, create again the event
7. Observe that no new test is scheduled in the queue

**Expected behavior**
I would expect that each matching event would trigger the labeled tests and they would be scheduled for execution

**Version / Cluster**
- Which testkube version?  1.13.2:de8
- What Kubernetes cluster? local
- What Kubernetes version? v1.24.2

vsukhin:
## Pull request description 

Fix #4361 

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vLia (approved):
Looks good!

nicufk (approved):

PR #4372, opened 2023-09-17
fix: change jmeterd docker versions

dejanzele:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/contrib/executor/jmeterd/pkg/slaves/client.go
+++ b/contrib/executor/jmeterd/pkg/slaves/client.go
@@ -141,7 +141,7 @@ func (client *Client) getSlavePodConfiguration(currentSlavesCount int) (*v1.Pod,
 			InitContainers: []v1.Container{
 				{
 					Name:            "init",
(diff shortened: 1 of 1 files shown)


contrib/executor/jmeterd/pkg/slaves/client.go
```diff
@@ -141,7 +141,7 @@ func (client *Client) getSlavePodConfiguration(currentSlavesCount int) (*v1.Pod,
 			InitContainers: []v1.Container{
 				{
 					Name:            "init",
-					Image:           "kubeshop/testkube-init-executor:1.14.3",
+					Image:           "kubeshop/testkube-init-executor:1.14.3", // TODO: make it configurable
```
vsukhin, line 144: Read from env var, if empty you can use hardcoded value

Issue #4456, opened 2023-10-11
Bug Gradle executor: environment variables in system properties are not evaluated


vLia:
**Describe the bug**
See conversation in Discord: https://discord.com/channels/884464549347074049/1160975216458072134

**To Reproduce**
Create test:
```bash
apiVersion: tests.testkube.io/v3
kind: Test
metadata:
  labels:
    executor: gradle-executor
    test-type: gradle-test
  name: gradle-test
  namespace: testkube
spec:
  content:
    repository:
      branch: gradle-tests
      path: gradle-example/hello-gradle
      type: git
      uri: https://github.com/vlia/testkube-tests.git
    type: git
  executionRequest:
    args:
    - -Dtestkube=$TESTKUBE
    - -i
    variables:
      TESTKUBE:
        name: TESTKUBE
        type: basic
        value: "true"
        valueFrom: {}
  type: gradle/test
```
Run test:
```
tk run test gradle-test
```

**Expected behavior**
Test is successful.

**Actual behavior**
Test fails due to the variable not being evaluated.

**Version / Cluster**
- Which testkube version? 
Client Version 1.14.6
Server Version 1.14.6
- What Kubernetes cluster? (e.g. GKE, EKS, Openshift etc, local KinD, local Minikube)
kind
- What Kubernetes version?
Client Version: v1.28.2
Server Version: v1.26.3

vsukhin:
this is more an enhancement

vsukhin:
yes, I agree, it's a nice improvement and shold work for all executors

vLia:
Created follow-up: https://github.com/kubeshop/testkube/issues/4458

PR #4457, opened 2023-10-11
feat: evaluate environment variables in the args of gradle

vLia:
## Pull request description 

https://github.com/kubeshop/testkube/issues/4456

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/contrib/executor/gradle/pkg/runner/runner.go
+++ b/contrib/executor/gradle/pkg/runner/runner.go
@@ -131,9 +131,9 @@ func (r *GradleRunner) Run(ctx context.Context, execution testkube.Execution) (r
 		if args[i] == "<projectDir>" {
 			args[i] = project
 		}
+		args[i] = os.ExpandEnv(args[i])
 	}
 
-	output.PrintEvent("Running task: "+task, directory, gradleCommand, args)
 	out, err := executor.Run(runPath, gradleCommand, envManager, args...)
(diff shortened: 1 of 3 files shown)


contrib/executor/gradle/pkg/runner/runner.go
```diff
@@ -131,9 +131,9 @@ func (r *GradleRunner) Run(ctx context.Context, execution testkube.Execution) (r
 		if args[i] == "<projectDir>" {
 			args[i] = project
 		}
+		args[i] = os.ExpandEnv(args[i])
 	}
 
-	output.PrintEvent("Running task: "+task, directory, gradleCommand, args)
```
vsukhin, line 136: why do we remove this - duplication?

vsukhin (approved):
one question

contrib/executor/gradle/pkg/runner/runner.go
vLia: Both duplication and it prints out the secret variables. The command with arguments will be printed out in the following function with obfuscated secrets, and I also checked that there are no errors before it.

vsukhin:
ok, also worth to create a ticket with similar improvements for all our prebuilt executors

PR #4500, opened 2023-10-23
fix: mask slack token in log message

skoef:
## Pull request description 
When configured, the Slack token is shown in plain text in the logs. This is not desired and this PR masks all but the first part of the token.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Fixes

- Mask slack token in log message

### DIFF
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kubeshop/testkube/pkg/api/v1/testkube"
 	"github.com/kubeshop/testkube/pkg/log"
 	"github.com/kubeshop/testkube/pkg/utils"
+	"github.com/kubeshop/testkube/pkg/utils/text"
 )
 
 type MessageArgs struct {
@@ -47,7 +48,7 @@ func NewNotifier(template, clusterName, dashboardURI string, config []Notificati
 		config: NewConfig(config), envs: envs}
 	notifier.timestamps = make(map[string]string)
 	if token, ok := os.LookupEnv("SLACK_TOKEN"); ok {
-		log.DefaultLogger.Infow("initializing slack client", "SLACK_TOKEN", token)
+		log.DefaultLogger.Infow("initializing slack client", "SLACK_TOKEN", text.Obfuscate(token))
 		notifier.client = slack.New(token, slack.OptionDebug(true))
 		notifier.Ready = true
 	} else {


vsukhin (approved):
lgtm. thank you @skoef!  please confirm @nicufk

nicufk (changes requested):
Looks good I think it is good to display at least the last/first 3 or 4 characters

pkg/slack/slack.go
```diff
@@ -241,3 +242,19 @@ func (s *Notifier) composeTestMessage(execution *testkube.Execution, eventType t
 	}
 	return message.Bytes(), nil
 }
+
+// maskToken only shows the 4 characters of a slack token and masks the rest
+func maskToken(s string) string {
+	fields := strings.Split(s, "-")
+	// only show first 4 characters of the first field
```
nicufk, line 249:
```suggestion
	// only show first n characters of the second field
```

pkg/slack/slack.go
```diff
@@ -241,3 +242,19 @@ func (s *Notifier) composeTestMessage(execution *testkube.Execution, eventType t
...
+	if len(fields[0]) > 4 {
```
nicufk, line 250:
```suggestion
        n := 4
	if len(fields[1]) > n {
```

pkg/slack/slack.go
```diff
@@ -241,3 +242,19 @@ func (s *Notifier) composeTestMessage(execution *testkube.Execution, eventType t
...
+		fields[0] = fields[0][:4] + strings.Repeat("X", len(fields[0])-4)
```
nicufk, line 251:
```suggestion
		fields[1] = fields[1][:n] + strings.Repeat("X", len(fields[1])-n)
```

pkg/slack/slack_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+func TestMaskToken(t *testing.T) {
+	assert.Equal(t, "", maskToken(""))
+	assert.Equal(t, "xoxb", maskToken("xoxb"))
+	assert.Equal(t, "xoxbXXXX", maskToken("xoxbkjh3"))
+	assert.Equal(t, "xoxb-", maskToken("xoxb-"))
+	assert.Equal(t, "xoxbXXXX-", maskToken("xoxbkjh3-"))
+	assert.Equal(t, "xoxb-XXXXX", maskToken("xoxb-sdfff"))
+	assert.Equal(t, "xoxb-XXXXX-XXX", maskToken("xoxb-sdfff-sdf"))
+	assert.Equal(t, "xoxb-XXXXX-XXX-XXXXXXXXXXXX", maskToken("xoxb-sdfff-sdf-341234234asd"))
```
nicufk, line 17:
```suggestion
	assert.Equal(t, "xoxb-XXXXX-XXX-XXXXXXXXXXXX", maskToken("xoxb-sdfff-sdf-341234234asd"))
	assert.Equal(t, len("xoxb-XXXXX-XXX-XXXXXXXXXXXX"), len(maskToken("xoxb-sdfff-sdf-341234234asd")))
```

nicufk (changes requested):
please check the suggestions

pkg/slack/slack.go
```diff
@@ -241,3 +242,20 @@ func (s *Notifier) composeTestMessage(execution *testkube.Execution, eventType t
 	}
 	return message.Bytes(), nil
 }
+
+// maskToken only shows the 4 characters of a slack token and masks the rest
+func maskToken(s string) string {
```
exu, line 247: It's quite generic function which should be in text package

skoef: @exu I think `Obfuscate` is even better, makes for a much smaller change in this MR and you could recognise one slack token from another from the leading and trailing non-obfuscated characters imho.

exu (approved):

nicufk (approved):

PR #4518, opened 2023-10-26
feat: improve performance of listing tests/test suites/executions

rangoo94:
## Pull request description 

Redesigned MongoDB queries for loading test/test suite executions.

* **Earlier:** there were 4 queries to MongoDB: `aggregate()` to get latest executions ID per `starttime` and `find()` to get these executions, and both the same for `endtime`. Later these were compared based on latest date.
* **Now:** all of that is resolved in a single `aggregate()`

Thanks to that, it works faster. To make it even faster, we would need to persist something like `updatetime` (latest of `starttime`/`endtime`) in the execution, but it can be a further improvement in future.

> **Cloud implementation**
> As for now Cloud API is using similar strategy as OSS before, with loading two sets on both `starttime`/`endtime` and comparing them, I kept the same behavior as it was for the `CloudRepository`.
>
> When the Cloud API will have it reinvented (for backwards compatibility, probably: will have a new Command that will use similar strategy), we may replace the `CloudRepository` to call it directly. Cc: @exu 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] updated tests

### DIFF
--- a/internal/app/api/v1/executions_test.go
+++ b/internal/app/api/v1/executions_test.go
@@ -135,11 +135,11 @@ func (r MockExecutionResultsRepository) GetByNameAndTest(ctx context.Context, na
 	panic("not implemented")
 }
 
-func (r MockExecutionResultsRepository) GetLatestByTest(ctx context.Context, testName, sortField string) (testkube.Execution, error) {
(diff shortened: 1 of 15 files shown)


vsukhin (approved):
lgtm

PR #4534, opened 2023-10-27
feat: add MongoDB migrations, run it on API start

rangoo94:
## Pull request description 

* Add MongoDB migrations mechanism
   * Inspired by https://github.com/golang-migrate/migrate, but we need to automatically detect migrations to run or rollback
   * Create new migrations by adding `.up.json` and `.down.json` files in the `config/db-migrations` directory
   * It will automatically detect changes in previous files (rolling back) and run all migrations from top
* There are some downsides
   * There is no lock mechanism - in OSS we are using single API server though
   * We don't use transactions - unfortunately, we don't use replicaset in OSS, which is required for transactions

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94:
@vsukhin, I've not been able to use `interface`s to shadow the `*mongo.Database` and write unit tests. Do you have some nice patterns for such unit tests? I'm not sure if we should:
* Use fake MongoDB, or
* Use real MongoDB, or
* Write wrapper around the Mongo (MongoRepository and Fake/MockRepository) that will be passed there

What's your opinion on that? Do we have some patterns?

cmd/api-server/main.go
```diff
@@ -242,6 +243,17 @@ func main() {
 				log.DefaultLogger.Infow("minio is not available, using default logs storage", "error", err)
 			}
 		}
+
+		// Run DB migrations
+		dbMigrator, err := dbmigrator.NewDbMigrator(db, "__migrations", cfg.TestkubeConfigDir)
+		ui.ExitOnError("Loading MongoDB migrations", err)
+		plan, _ := dbMigrator.Plan(ctx)
```
vsukhin, line 250: we need to log error

cmd/api-server/main.go
```diff
@@ -242,6 +243,17 @@ func main() {
 				log.DefaultLogger.Infow("minio is not available, using default logs storage", "error", err)
 			}
 		}
+
+		// Run DB migrations
+		dbMigrator, err := dbmigrator.NewDbMigrator(db, "__migrations", cfg.TestkubeConfigDir)
```
vsukhin, line 248: might be it will be good to enable/disable migrations in helm chart

cmd/api-server/main.go
```diff
@@ -242,6 +243,17 @@ func main() {
...
+		ui.ExitOnError("Loading MongoDB migrations", err)
```
vsukhin, line 249: Should it be error or warning? Is it critical?

cmd/api-server/main.go
```diff
@@ -242,6 +243,17 @@ func main() {
...
+		plan, _ := dbMigrator.Plan(ctx)
+		if plan.Total == 0 {
+			log.DefaultLogger.Info("No MongoDB migrations to apply.")
+		} else {
+			log.DefaultLogger.Info(fmt.Sprintf("Applying MongoDB migrations: %d rollbacks and %d ups.", len(plan.Downs), len(plan.Ups)))
+		}
+		ui.ExitOnError("Executing MongoDB migrations", dbMigrator.Apply(ctx))
```
vsukhin, line 256: Should it be error or warning? Is it critical?

vsukhin (commented):
a few comments

cmd/api-server/main.go
rangoo94: I think that when MongoDB migrations will be a part of code, we should rather consider them as critical error. Each new part of code may depend on some migrations (i.e. if migration will rename `status` property name to `executionStatus` and that migration won't be ran, the API will not be working correctly)

cmd/api-server/main.go
rangoo94: Same as above

cmd/api-server/main.go
rangoo94: We can take the "warning" approach too (as MongoDB is rather flexible, so often lack of migrations will not break anything), but it's just less safe

vsukhin: great

vsukhin (approved):

vsukhin:
I would follow last one `Write wrapper around the Mongo (MongoRepository and Fake/MockRepository) that will be passed there`

cmd/api-server/main.go
rangoo94: Added `DISABLE_MONGO_MIGRATIONS` environment variable, and configured it via `disableMongoMigrations` in Helm chart (https://github.com/kubeshop/helm-charts/pull/662)

vsukhin: cool

rangoo94:
Thanks @vsukhin, added unit tests, environment variable / helm chart variable to disable it, and made MongoDB failure just a warning 👍

PR #4536, opened 2023-10-27
feat: speed up - use covered queries for tests/test suites list

rangoo94:
## Pull request description 

* This is making use of the queries covered by indexes. It's achieving constant low-time results for big sets of data.
* It requires indexes though
   * without indexes is heavily slow, because its `$lookup`s performance depends on them. Without indexes, the solution is 30x slower than existing one
   * ~making as draft until any mechanism to create them automatically will be there (i.e.) https://github.com/kubeshop/testkube/pull/4534~

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/internal/db-migrations/02_execution_search_indexes.down.json
+++ b/internal/db-migrations/02_execution_search_indexes.down.json
@@ -1,0 +1,10 @@
+[
+  {
+    "dropIndexes": "testresults",
+    "index": ["testsuite.name_hashed_starttime-1", "testsuite.name_hashed_endtime-1"]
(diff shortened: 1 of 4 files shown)


rangoo94:
@vsukhin, rebased on `develop` and tested, works well with the indexes that are now created automatically 👍

vsukhin:
thank you @rangoo94 !

vsukhin (approved):

PR #4540, opened 2023-10-28
feat: embed MongoDB migrations in the binary

rangoo94:
## Pull request description 

* Avoid using external directory, as it needs to be mounted to the container - the migrations are integral part of the application

- [x] tested locally
- [x] tested on cluster
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


vsukhin (approved):
good, pk/crd is also use embedded fs

PR #4576, opened 2023-11-03
fix: result line for websockets

exu:
## Pull request description 

should fix this first line:

image

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/agent/agent.go
+++ b/pkg/executor/agent/agent.go
@@ -95,9 +95,9 @@ func Run(ctx context.Context, r runner.Runner, args []string) {
 
 	if r.GetType().IsMain() {
 		output.PrintEvent("test execution finished", e.Id)
+		output.PrintResult(result)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/executor/agent/agent.go
```diff
@@ -95,9 +95,9 @@ func Run(ctx context.Context, r runner.Runner, args []string) {
 
 	if r.GetType().IsMain() {
 		output.PrintEvent("test execution finished", e.Id)
+		output.PrintResult(result)
 	}
 
-	output.PrintResult(result)
```
vsukhin, line 100: please check, that it will not break checking for errors in init and scraper executors

exu: there is no error passed in this line - they are above.

PR #4594, opened 2023-11-08
feat: add minio log consumer

nicufk:
## Pull request description 
When processing 100 log providers with sizes between 1 MB and 105 MB, in use memory of the minio consumer doesn't exceed 400 MB, the pod Memory can reach 3500 Mi and CPU 800m
Screenshot 2023-11-27 at 16 54 07
Screenshot 2023-11-27 at 16 17 26
Screenshot 2023-11-27 at 16 16 33
Screenshot 2023-11-27 at 14 14 46

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,102 @@
...
+type MinioConsumer struct {
+	minioConnecter *minioconnecter.Connecter
+	bucket         string
+	region         string
+	Log            *zap.SugaredLogger
+	disconnected   bool
+}
+
+func (s *MinioConsumer) Notify(id string, e events.LogChunk) error {
+	if s.disconnected {
```
exu, line 41: we'll lose data - I would return error here

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,102 @@
+package consumer
```
exu, line 1: Can't we init minio Writer on the beginning and just push chunk of data to it. 

This implementation be quite slow - you're loading appending removing object in one go - it'll be impossible to write even 100 lines per second

pkg/storage/minio/minio_connecter.go
```diff
@@ -0,0 +1,80 @@
...
+		c.minioClient, err = c.Connect()
+		if err != nil {
+			return nil, err
+		}
+	}
+	return c.minioClient, nil
+}
+
+// Connect connects to MinIO server
+func (c *Connecter) Connect() (*minio.Client, error) {
```
exu, line 50: Why we're duplicating code from minio.go file here?

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/kubeshop/testkube/pkg/log"
+	"github.com/kubeshop/testkube/pkg/logs/events"
+	minioconnecter "github.com/kubeshop/testkube/pkg/storage/minio"
+)
+
+var _ Consumer = &MinioConsumer{}
+
+// MinioConsumer creates new MinioSubscriber which will send data to local MinIO bucket
+func NewMinioConsumer(endpoint, accessKeyID, secretAccessKey, region, token, bucket string, ssl bool) *MinioConsumer {
+	c := &MinioConsumer{
```
exu, line 21: init bucket must be done here

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,102 @@
...
+
+type MinioConsumer struct {
+	minioConnecter *minioconnecter.Connecter
+	bucket         string
+	region         string
+	Log            *zap.SugaredLogger
+	disconnected   bool
+}
+
+func (s *MinioConsumer) Notify(id string, e events.LogChunk) error {
```
exu, line 40: and here just thin append to the bucket file

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	file, err := minioClient.GetObject(context.TODO(), s.bucket, id, minio.GetObjectOptions{})
+	if err != nil {
+		return err
+	}
+
+	fileContent, err := io.ReadAll(file)
```
exu, line 67: we must stream content to the file 

e.g. for GCP we have option to get 
```
func (s *GCPStorage) ObjectWriter(ctx context.Context, key string) (io.WriteCloser, error) {
	obj := s.client.Bucket(s.bucket).Object(key)
	w := obj.NewWriter(ctx)
	w.ContentType = "text/plain"
	return w, nil
}
```

exu: and you can stream to it we would need to figure out how to do the same for MinIO

pkg/storage/minio/minio_connecter.go
nicufk: actually I got it here to not duplicate it and to be able to use it in the log minio consumer

pkg/logs/consumer/minio.go
nicufk: minio doesn't support stream writer, I implemented a buffer version

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,170 @@
...
+func NewMinioConsumer(endpoint, accessKeyID, secretAccessKey, region, token, bucket string, ssl bool) *MinioConsumer {
+	c := &MinioConsumer{
+		minioConnecter: minioconnecter.NewConnecter(endpoint, accessKeyID, secretAccessKey, region, token, bucket, ssl, log.DefaultLogger),
+		Log:            log.DefaultLogger,
+		bucket:         bucket,
+		region:         region,
+		disconnected:   false,
+		buffInfos:      make(map[string]BufferInfo),
+	}
+	minioClient, err := c.minioConnecter.GetClient()
```
exu, line 46: I don't get why you need to overcomplicate it with minioConnecter? - looks like you need minio client only here? This one adds additional level of object hierarchy.

pkg/logs/consumer/minio.go
```diff
@@ -0,0 +1,170 @@
+package consumer
+
+import (
```
exu, line 3: I would go with writing stream in chunks. As we can't implement writing on the end of stream. 

So each second logs should be flushed to the minio.

nicufk: Added saving on a 80 kB batch

PR #4673, opened 2023-11-23
fix: permission issue fix in .testkube directory

emamihe:
## Pull request description 

Permission issue in /.testkube for updating the config file fixed. since the user for the container was 1001 and it hadn't the permission to update the config file

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/build/kubectl-testkube/Dockerfile
+++ b/build/kubectl-testkube/Dockerfile
@@ -2,6 +2,6 @@
 ARG ALPINE_IMAGE
 FROM ${ALPINE_IMAGE}
 COPY kubectl-testkube /bin/kubectl-testkube
-RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube && chown -R 1001. /.testkube && chmod 660 /.testkube/config.json
+RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube && chown -R 1001:1001 /.testkube && chmod 660 /.testkube/config.json
 USER 1001
 ENTRYPOINT ["/bin/kubectl-testkube"]


vsukhin (approved):

rangoo94 (approved):

build/kubectl-testkube/Dockerfile
```diff
@@ -2,6 +2,6 @@
 ARG ALPINE_IMAGE
 FROM ${ALPINE_IMAGE}
 COPY kubectl-testkube /bin/kubectl-testkube
-RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube
+RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube && chown -R 1001. /.testkube && chmod 660 /.testkube/config.json
```
rangoo94, line 5:
```suggestion
RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube && chown -R 1001 /.testkube && chmod 660 /.testkube/config.json
```

rangoo94: Is it expected? I didn't knew about syntax for `chown` user ID with a dot

rangoo94: Looking at `man`:

```
Some older scripts may still use ‘.’ in place of the ‘:’ separator. POSIX 1003.1-2001 (see [Standards conformance](https://www.gnu.org/software/coreutils/manual/html_node/Standards-conformance.html)) does not require support for that, but for backward compatibility GNU chown supports ‘.’ so long as no ambiguity results, although it issues a warning and support may be removed in future versions. New scripts should avoid the use of ‘.’ because it is not portable, and because it has undesirable results if the entire owner‘.’group happens to identify a user whose name contains ‘.’.
```

rangoo94: If that's working, good enough 👍 Although, based on the `man` this is deprecated, so probably `1001:1001` specifically could be better:
```suggestion
RUN mkdir /.testkube && echo "{}" > /.testkube/config.json && chmod -R 755 /.testkube && chown -R 1001:1001 /.testkube && chmod 660 /.testkube/config.json
```

Issue #4587, opened 2023-11-07
Distributed Jmeter - Run load test from master pod if SLAVE_COUNT <= 1


jmorante-ks:
**Describe the enhancement you'd like to see**
Currently when SLAVES_COUNT is 1, the Distributed Jmeter is starting a slave pod and also the master node, which is a waste of resources. 

The test could be executed from master node when just one slave is requested

**Additional context**
Ideally we could combine Distributed Jmeter and Jmeter executors in one, and depending on SLAVES_COUNT value it could behave as standard jmeter (when SLAVES_COUNT <= 1 or undefined) or as distributed jmeter if SLAVES_COUNT > 1.

Issue #4589, opened 2023-11-07
Distributed Jmeter - Invalid test result when Jmeter fails to start


jmorante-ks:
**Describe the bug**
Testkube uses the exit code of the test process as the way to know if test failed or not.

Jmeter process returns exit 0 (succeed) when it fails to start. As a consequence the test is marked as "passed" while the test didn't run properly.

That could be considered as a limitation or a bug of Jmeter, but we need to find a way to resolve it for Testkube, otherwise it won't be possible to detect this kind of failures in Testkube.

**To Reproduce**
Configure a Jmeter test with an intentional bug that makes it crash, and you will be able to reproduce the issue.

**Expected behavior**
In case of startup or Jmeter internal error, the test should be marked as failed in Testkube.

**Screenshots**
Example:
![image](https://github.com/kubeshop/testkube/assets/142315352/d571acb5-1d59-444f-840f-6d67c1682a47)

 
**Additional context**
Example of this Jmeter known limitation in stackoverflow: https://stackoverflow.com/questions/49652379/jmeter-exit-code-in-bash-script-always-returns-0

PR #4704, opened 2023-11-29
feat: refactor jmeterd executor to add better testing and support for better exit status handling

dejanzele:
## Pull request description 

Refactor executor so components are easily testable.
Add more unit tests and integration tests.
Add support for handling the scenario when test has an error and JMeter returns 0 exit code
Support for running regular JMeter if SLAVES_COUNT is omitted or set to 0 or 1.

This should close https://github.com/kubeshop/testkube/issues/4587 and https://github.com/kubeshop/testkube/issues/4589

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


contrib/executor/jmeterd/pkg/parser/csvjtl.go
```diff
@@ -0,0 +1,80 @@
...
+				dict[header[i]] = record[i]
+			}
+			rows = append(rows, dict)
+		}
+	}
+	return rows, nil
+}
+
+func MapElementToResult(in map[string]string) CSVResult {
+	elapsed, _ := strconv.Atoi(in["elapsed"])
```
vsukhin, line 71: do we need to log error?

vsukhin (approved):
looks good, fix  unit tests

config/slave-pod-template.yml
```diff
@@ -3,11 +3,13 @@ kind: Pod
 metadata:
   name: "{{ .Name }}"
   namespace: {{ .Namespace }}
+  {{- if .JobUID }}
```
vsukhin, line 6: don't forget tto change it in helm-chart

contrib/executor/jmeterd/pkg/parser/csvjtl.go
dejanzele: Good question, we parse the JTL file here, it shouldn't ever error, unless JMeter changes it to something else in future versions which doesn't seem likely.
Even if it failed, I don't know should we fail the test, but might make sense to log the error at least

contrib/executor/jmeterd/pkg/runner/runner.go
```diff
@@ -148,146 +129,116 @@ func (r *JMeterDRunner) Run(ctx context.Context, execution testkube.Execution) (
...
 	entryPoint := getEntryPoint()
 	for i := range execution.Command {
 		if execution.Command[i] == "<entryPoint>" {
 			execution.Command[i] = entryPoint
+			break
 		}
 	}
 
+	d, _ := exec.Command("pwd").Output()
+	fmt.Println(string(d))
```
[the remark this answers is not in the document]
dejanzele: yep, I removed it

PR #4823, opened 2023-12-21
fix: put pre-run and post-run script files in the data volume

rangoo94:
## Pull request description 

* The pre-run and post-run scripts were put in the working directory, while it not necessarily was available as a volume.
  * Moved all of shell files into `DataDir`
  * Move command to separate file so initial `workingDir` will be persisted for post-run script
* Additionally, the failure of pre-run script is not blocking the rest of executing.
  * Added `set -e` to fail fast

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

- testkube#4719

### DIFF
--- a/contrib/executor/init/pkg/runner/runner.go
+++ b/contrib/executor/init/pkg/runner/runner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


nicufk (approved):

vLia (approved):

vLia:
For the failures just run `make docs`. The cloud commands with the missing files are marked as deprecated in the docs, and we will delete them together with the commands.

Issue #4844, opened 2023-12-25
Documentation on running tests uses an invalid flag in the example


dejanzele:
**Describe the bug**
In the [Passing Parameters](https://docs.testkube.io/articles/running-tests/#passing-parameters) section of the [Running tests](https://docs.testkube.io/articles/running-tests) page it is mentioned that parameters can be passed using `-p`.

Example:
```
kubectl testkube run test kubeshop-cypress -p testparam=testvalue -f --secret secret-name=secret-key
```

But when this command is executed, it errors with `unknown shorthand flag: 'p' in -p`

**To Reproduce**
1. Try to run any test with `-p` parameter and it will fail with the same error

**Expected behavior**
Test is run with the provided parameter(s).

**Version / Cluster**
- Which testkube version?  Any
- What Kubernetes cluster? (e.g. GKE, EKS, Openshift etc, local KinD, local Minikube) Any
- What Kubernetes version? Any
 
**Additional context**
This was reported by users on Discord [here](https://discord.com/channels/884464549347074049/885185707419770930/1187835143667335219)

windowsrefund:
FYI, This may relate in one way or another to #787 which was also recently discussed in this Discord [thread](https://discord.com/channels/884464549347074049/885185660808474664/1188663142918656123).

PR #4845, opened 2023-12-26
docs: fix invalid flag in running-tests.md

dejanzele:
## Pull request description 

Fix an invalid flag (`--parameter|-p` instead of `--variable|-v`) in the [Passing parameters](https://docs.testkube.io/articles/running-tests/#passing-parameters) section of Testkube docs.

This closes https://github.com/kubeshop/testkube/issues/4844

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/docs/docs/articles/running-tests.md
+++ b/docs/docs/articles/running-tests.md
@@ -107,7 +107,7 @@ For some 'real world' tests, configuration variables are passed in order to run 
 
 Let's assume that our example Cypress test needs the `testparam` parameter with the value `testvalue`.
 
-This is done by using the `--variable` flag (or just `-v`). If you need to pass more parameters, either pass the values as a comma-separated string or simply pass multiple `-p` flags.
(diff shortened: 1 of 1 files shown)


vLia (approved):

jfermi (approved):

windowsrefund:
Not sure if this is the right place but there seems to be some misalignment around the terms "parameters", "variables", and "environment variables". I'm pretty sure this `-v` flag essentially results in the creation of an environment variable which is odd considering we're discussing "passing parameters". Also, it's not clear how `-v` is any different than the `--env` flag?

TheBrunoLopes:
Hey @windowsrefund .
you are right, some time ago we were using the word parameters but we found it confusing to explain the different between that and "arguments". So now we have variables and arguments.

@dejanzele perhaps it's better to change the name of the section from "Passing Parameters" to "Passing Variables".

Issue #4850, opened 2024-01-03
testkube-api results/v1/config not found


fvogl:
For testkube-api > 1.16.16 we get this error on the dashboard

```
tests:1 Access to fetch at 'https://testkube-api.ourdomain.com/results/v1/config' from origin 'https://testkube-dashboard.ourdomain.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
fetchUtils.ts:72 GET https://testkube-api.ourdomain.com/results/v1/config net::ERR_FAILED 404 (Not Found)
```

Boxes show just a loading animation

![image](https://github.com/kubeshop/testkube/assets/26383910/cc6e8f57-e369-497f-8211-b32256ce7118)

Note: I don't use ingress from the chart, but have my route for openshift created manually. Was working <= 1.16.16

vsukhin:
@rangoo94 @ypoplavs anything wrong with ingress spec?

rangoo94:
* I don't think our API server is returning `Access-Control-*` headers, it is added on the ingress level
* @fvogl noted that it's some custom ingress, so unrelated to our ingress configuration too
* The UI was always using `fetch` with default `mode: cors, credentials: same-origin`, there were no changes regarding that

@vsukhin, considering that, I don't think it's anything related to Testkube changes.

@fvogl, maybe you had some manual ingress changes in the OpenShift previously, that were not preserved after the upgrade?

rangoo94:
@fvogl, I've got a suspicion that I know what is wrong - is the `testkube-api.ourdomain.com/results/v1/config` the one that should be? I would rather assume that `/results/v1` will be below the same domain as the Dashboard (what would solve CORS problem):

* `testkube.ourdomain.com` pointing to Testkube Dashboard
* `testkube.ourdomain.com/results/v1` - pointing to Testkube API Server

fvogl:
@rangoo94 , the thing is that testkube was working perfectly fine until I've updated it to 1.16.20. When I change the testkube-api image back to 1.16.16 it works fine again. If I change it to anything > 1.16.16 I get what I reported.

rangoo94:
The only changes I see between `1.16.16` and `1.16.17` ([**here**](https://github.com/kubeshop/testkube/compare/v1.16.16...v1.16.17)) are:
* changes to internal pipeline
* changes to the CLI
[...]

PR #4852, opened 2024-01-03
fix: recover /results/v1 alias

rangoo94:
## Pull request description 

Recover `/results/v1` alias deleted in https://github.com/kubeshop/testkube/commit/b80ca88af6f75fbc4464e26d62aceef6ac28a2b5 and https://github.com/kubeshop/testkube/commit/fc71643c52f2f3f3a24707b028adadea4a037010

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Fixes

- kubeshop/testkube#4850

### DIFF
--- a/internal/app/api/v1/server.go
+++ b/internal/app/api/v1/server.go
@@ -405,6 +405,10 @@ func (s *TestkubeAPI) InitRoutes() {
 	repositories := root.Group("/repositories")
 	repositories.Post("/", s.ValidateRepositoryHandler())
 
+	// mount everything on results
+	// TODO it should be named /api/ + dashboard refactor
+	s.Mux.Mount("/results", s.Mux)
+
 	// mount dashboard on /ui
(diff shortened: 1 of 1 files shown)


vsukhin (approved):

rangoo94:
It's not that simple `s.Mux.Mount` requires a standalone application that is not mounted anywhere.

Instead, maybe `HTTPServer.Routes` could need to be modified to struct that will implement `Router`, passing to multiple groups. It will have downsides anyway to:
* Complex implementation (as it should work recurrently this way)
* If that will mount some sub-application in any place of tree, it won't work (and lead to strange error that may be hard to debug)

@vsukhin, any other idea on how to solve it? or maybe the drawbacks from the solution above are acceptable?

rangoo94:
Closing, as I have no idea how to cover that.

PR #4921, opened 2024-01-22
fix: added cooldown time to not stop events too early

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/logs/events.go
+++ b/pkg/logs/events.go
@@ -150,7 +150,12 @@ func (ls *LogsService) handleStart(ctx context.Context) func(msg *nats.Msg) {
 func (ls *LogsService) handleStop(ctx context.Context) func(msg *nats.Msg) {
 	return func(msg *nats.Msg) {
 		ls.log.Debugw("got stop event")
-		time.Sleep(StopWaitTime)
+
+		t := time.NewTicker(StopWaitTime)
+		select {
+		case <-t.C:
+		case <-ctx.Done():
+		}
 
 		event := events.Trigger{}
 		err := json.Unmarshal(msg.Data, &event)
--- a/pkg/logs/events.go
+++ b/pkg/logs/events.go
@@ -25,6 +25,8 @@ const (
 
 	StartQueue = "logsstart"
 	StopQueue  = "logsstop"
+
+	StopWaitTime = 60 * time.Second // when stop event is faster than first message arrived
 )
 
 type Consumer struct {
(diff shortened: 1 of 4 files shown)


pkg/logs/events.go
```diff
@@ -147,6 +149,8 @@ func (ls *LogsService) handleStart(ctx context.Context) func(msg *nats.Msg) {
 // handleStop will handle stop event and stop logs consumers, also clean consumers state
 func (ls *LogsService) handleStop(ctx context.Context) func(msg *nats.Msg) {
 	return func(msg *nats.Msg) {
+		ls.log.Debugw("got stop event")
+		time.Sleep(StopWaitTime)
```
vsukhin, line 153: shouldn't it be ticker with context.done check?

vsukhin: looks a bit hacky

pkg/scheduler/test_scheduler.go
```diff
@@ -840,18 +841,20 @@ func (s *Scheduler) triggerLogsStartEvent(ctx context.Context, id string) error
 
 func (s *Scheduler) triggerLogsStopEvent(ctx context.Context, id string) error {
 	if s.featureFlags.LogsV2 {
-		r, err := s.logsStream.Stop(ctx, id)
-		if err != nil {
-			s.logger.Errorw("can't send stop event for logs", "id", id, "error", err)
-			return err
-		}
+		go func() {
```
vsukhin, line 844: sounds like the whole method is async, you can simply call go s.triggerLogsStopEvent(...)

vsukhin (approved):
couple of notes

pkg/logs/events.go
exu: Yep the issue is that it's all async - and this event can be handled before first message even arrive - so I need to deal somehow with it. 

Log is in consumer which I must remove from memory when not needed anymore. 
So there are events:
- START - triggers messages handling (need to know ID of execution to start consumer) 
- STOP (cleaning)

exu: passed ctx handling

pkg/scheduler/test_scheduler.go
exu: No it's sync - I want to know here that it was handled synchronously (to have better logging in case of issues)

exu: So we'll know here that event was handled by the logs-server end

pkg/logs/events.go
vsukhin: reminds me tcp package flow, they can also come async in a mixed form and must be reordered. Ideally, would be great to check, that log server received all messages in the stream with order number less that the stop message number, this way it will be safe to stop it. but I don't know if log server tracks it

exu: nope - but in general good idea - but will be hard to accomplish as assuming logs can be send from different places (e.g. job sidecar, or even test scheuduler in case of some init errors, so there will be no single counter where we can count lines :/)

PR #4942, opened 2024-01-25
feat: add minio adapter for logs

nicufk:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/logs/main.go
+++ b/cmd/logs/main.go
@@ -55,7 +55,14 @@ func main() {
 		WithGrpcAddress(cfg.GrpcAddress)
 
 	// TODO - add adapters here
-	svc.AddAdapter(adapter.NewDummyAdapter())
+	minioAdapter, err := adapter.NewMinioAdapter(cfg.MinioEndpoint, cfg.MinioAccessKey, cfg.MinioSecretKey, cfg.MinioRegion, cfg.MinioToken, cfg.MinioBucket)
+	if err != nil {
+		log.Errorw("error creating minio adapter", "error", err)
+		svc.AddAdapter(adapter.NewDummyAdapter())
+	} else {
+		log.Infof("minio adapter created")
+		svc.AddAdapter(minioAdapter)
+	}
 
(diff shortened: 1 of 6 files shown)


cmd/logs/main.go
```diff
@@ -55,7 +55,14 @@ func main() {
 		WithGrpcAddress(cfg.GrpcAddress)
 
 	// TODO - add adapters here
-	svc.AddAdapter(adapter.NewDummyAdapter())
+	minioAdapter, err := adapter.NewMinioAdapter(cfg.MinioEndpoint, cfg.MinioAccessKey, cfg.MinioSecretKey, cfg.MinioRegion, cfg.MinioToken, cfg.MinioBucket)
+	if err != nil {
+		log.Errorw("error creating minio adapter", "error", err)
+		svc.AddAdapter(adapter.NewDummyAdapter())
```
vsukhin, line 61: do we need DummyAdapter in prod?

vsukhin: sounds like, log service will not store logs without minio?

exu (approved):

cmd/logs/main.go
exu: Yeah I would rename it to LogAdapter or sth and attach it onlly when debug mode is set

vsukhin (approved):
a couple of questions

cmd/logs/main.go
exu: Good point - it should only try to set minio when it's not in agent mode

exu: I wold put whole logic into if when in OSS mode
additionally add additional dummyadapter when DEBUG is set 
and add cloud adapter when Agent key is set

nicufk: got it, I will make the changes then

PR #4943, opened 2024-01-25
fix: use pointer everywhere when passing log chunk

exu:
## Pull request description 

* one way to pass and create logs (was messy pointer and non-pointer) 
* passed logs stream as parameter to executors and api

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/executor/client/job.go
+++ b/pkg/executor/client/job.go
@@ -44,6 +44,8 @@ import (
 	"github.com/kubeshop/testkube/pkg/executor/env"
 	"github.com/kubeshop/testkube/pkg/executor/output"
 	"github.com/kubeshop/testkube/pkg/log"
+	logsclient "github.com/kubeshop/testkube/pkg/logs/client"
+	"github.com/kubeshop/testkube/pkg/logs/events"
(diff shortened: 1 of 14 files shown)


pkg/executor/client/job.go
```diff
@@ -225,6 +230,9 @@ func (c *JobExecutor) Execute(ctx context.Context, execution *testkube.Execution
 	if err != nil {
 		return result.Err(err), err
 	}
+
+	c.logsStream.Push(ctx, execution.Id, events.NewLog("created kubernetes job"))
```
vsukhin, line 234: shoud it be for v2 only?

exu: 👍🏻

vsukhin (commented):
sounds like feature check is missing

pkg/executor/client/job.go
exu: yeah need to pass feature flags too :/ good point!

exu:
> sounds like feature check is missing

addded handling of feature flags.

vsukhin (approved):

PR #4994, opened 2024-02-12
fix: update cloud env vars to pro

vLia:
## Pull request description 

Update the name of the environment variables to \*\_PRO\_\* from \*\_CLOUD\_\* 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


internal/app/api/v1/handlers.go
```diff
@@ -57,17 +51,22 @@ func (s *TestkubeAPI) AuthHandler() fiber.Handler {
 // InfoHandler is a handler to get info
 func (s *TestkubeAPI) InfoHandler() fiber.Handler {
 	apiContext := contextOSS
-	if os.Getenv(cloudApiKeyEnvName) != "" {
+	if s.proContext != nil && s.proContext.APIKey != "" {
```
vsukhin, line 54: shouldn't be contextCloud also renamed below?

vLia: this PR is only about the environment variables, the context is still called `cloud`. We should probably create a follow-up PR to rename the contexts as well

vsukhin (approved):

internal/app/api/v1/handlers.go
vsukhin: ok

PR #5005, opened 2024-02-14
feat: improve the Docker image inspection to cache the results in Memory and/or ConfigMap

rangoo94:
## Pull request description 

Make the Docker image inspector faster, by caching the results in memory and optionally persisting in the ConfigMap.

* By default it's only caching it in the memory
* The cache in ConfigMap is rolling the old results when the cache size is too big (as there is 1MB limit)

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


rangoo94:
The PR looks pretty big, but actually ~700 lines are tests and auto-generated mocks 👍

vsukhin (approved):

vsukhin:
good job! nice improvement

PR #5015, opened 2024-02-15
fix: source for running scripts

### LINKED ISSUE #5004: prerun.sh is not run using source command
**Describe the bug**
Unable to export environment variables in a manner that allows Playwright to access them.

**To Reproduce**
create .sh file with:

```bash
export SOME_ENV_VARIABLE=some_value
npx playwright test
```

In Playwright code read somewhere SOME_ENV_VARIABLE using process.env.SOME_ENV_VARIABLE

Create Testkube test that will use native Playwright executor. Add your .sh file to run in Pre-Run phase.
What will happen is test run in Pre-Run phase will have access to SOME_ENV_VARIABLE but test run using executor-binary will not have access to that variable.

**Expected behavior**
SOME_ENV_VARIABLE should be available in both cases. export SOME_ENV_VARIABLE should add it to global context

**Version / Cluster**
- Which testkube version? 
API Versionv1.16.26
Dashboard Versionv1.16.2
Helm Chart Versionv1.16.34

- What Kubernetes cluster? (e.g. GKE, EKS, Openshift etc, local KinD, local Minikube)
Kubernetes cluster OnPrem
Client Version: v1.28.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.20.2
 
**Additional context**
To resolve that problem, entrypoint.sh, prerun.sh, postrun.sh should all be run using the source command. We tried that and it worked.
We aim to utilize the native Playwright executor and prefer not to develop our own, especially just for incorporating environment variables.

vsukhin:
## Pull request description 

Fix #5004 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


nicufk (approved):

contrib/executor/init/pkg/runner/runner.go
```diff
@@ -87,18 +87,18 @@ func (r *InitRunner) Run(ctx context.Context, execution testkube.Execution) (res
 		postRunScript := shebang
 
 		if execution.PreRunScript != "" {
-			entrypoint += strconv.Quote(filepath.Join(r.Params.DataDir, preRunScriptName)) + "\n"
+			entrypoint += "source " + strconv.Quote(filepath.Join(r.Params.DataDir, preRunScriptName)) + "\n"
```
rangoo94, line 90: `. file.sh` is slightly more portable across shells than `source file.sh`, but nowadays `source` should be in most places anyway

pkg/executor/containerexecutor/tmpl.go
```diff
@@ -210,7 +210,7 @@ func NewJobOptions(log *zap.SugaredLogger, templatesClient templatesv1.Interface
 	execution testkube.Execution, options client.ExecuteOptions, natsUri string, debug bool) (*JobOptions, error) {
 	jobOptions := NewJobOptionsFromExecutionOptions(options)
 	if execution.PreRunScript != "" || execution.PostRunScript != "" {
-		jobOptions.Command = []string{filepath.Join(executor.VolumeDir, EntrypointScriptName)}
+		jobOptions.Command = []string{"source", filepath.Join(executor.VolumeDir, EntrypointScriptName)}
```
rangoo94, line 213: Isn't `Command` directly a `command` in the container? If so, we shouldn't have `source` here, as it's not executable - it's built-in function in shell:
```
$ type source
source is a shell builtin
```

contrib/executor/init/pkg/runner/runner.go
vsukhin: I was planning to test it today, might be we need a flag to use source

pkg/executor/containerexecutor/tmpl.go
vsukhin: Yes, I was goiny to check

contrib/executor/init/pkg/runner/runner.go
vsukhin: Will add a flag for flexibility

rangoo94: `.` is rather enough - in shells where `source` is implemented, rather always it's alias for `.`

PR #5021, opened 2024-02-19
feat: [TKC-1207] TCL license checker

vLia:
## Pull request description 

Added license checker code with reference to licenses

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/tcl/README.md
```diff
@@ -0,0 +1,7 @@
+# Testkube Operator - TCL Package
+
+This folder contains special code with the Testkube Community license.
+
+## License
+
+The code in this folder is licensed under the Testkube Community License. Please see the [LICENSE](../../licenses/TCL.txt) file for more information.
```
olensmar, line 7: we should add a link to the FAQ here is as well - but it could be in a separate PR if this needs to be merged..

vLia: Added a link to the FAQ as well given how these two PRs will likely be merged around the same time.

olensmar (approved):
lgtm!

pkg/tcl/checktcl/subscription.go
```diff
@@ -0,0 +1,114 @@
...
+	cloudconfig "github.com/kubeshop/testkube/pkg/cloud/data/config"
+	"github.com/kubeshop/testkube/pkg/cloud/data/executor"
+)
+
+type SubscriptionChecker struct {
+	proContext config.ProContext
+	orgPlan    *OrganizationPlan
+}
+
+// NewCLISubscriptionChecker creates a new subscription checker using a user token instead of the agent token
```
vsukhin, line 31: is this for running from testkube cli, because it might not be connected via http?

vsukhin: is it requesting cloud/enterprise api or agent api?

rangoo94: Is it actually needed on the CLI level? Shouldn't it be rather indicated by API/Agent response, ending up with `402 Payment Required` or `403 Forbidden`?

rangoo94: > yes, if we need it, then we should have agent api method, because it can be used also in Dashboard

In the Dashboard we already have an organization details, incl. plan (and features based on that), so the only problem is direct API access

vLia: Should I remove it for now, and we can add it if/when we need it? For now, the only use case of the CLI connecting directly to the cloud API server is on agent setup.

vLia: I removed this function. Let's keep Pro API communication in the agent unless there is a very good reason not to.

vsukhin (approved):

rangoo94 (approved):

PR #5026, opened 2024-02-19
feat: enhance init command telemetry

haneabogdan:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/cloud/init.go
+++ b/cmd/kubectl-testkube/commands/cloud/init.go
@@ -2,6 +2,7 @@ package cloud
 
 import (
 	"fmt"
+	
 	"github.com/spf13/cobra"
 
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
--- a/cmd/kubectl-testkube/commands/cloud/init.go
+++ b/cmd/kubectl-testkube/commands/cloud/init.go
@@ -1,13 +1,13 @@
 package cloud
 
 import (
+	"fmt"
 	"github.com/spf13/cobra"
 
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/config"
 	"github.com/kubeshop/testkube/pkg/telemetry"
 	"github.com/kubeshop/testkube/pkg/ui"
-	"fmt"
 )
 
 func NewInitCmd() *cobra.Command {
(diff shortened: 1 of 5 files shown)


cmd/kubectl-testkube/commands/cloud/init.go
```diff
@@ -43,23 +45,27 @@ func NewInitCmd() *cobra.Command {
 				ui.NL()
 
 				currentContext, err := common.GetCurrentKubernetesContext()
-				sendErrTelemetry(cmd, cfg, "k8s_context")
-				ui.ExitOnError("getting current context", err)
+				if err != nil {
+					sendErrTelemetry(cmd, cfg, "k8s_context", err)
```
vsukhin, line 49: will we send telemetry only in case of error?

cmd/kubectl-testkube/commands/cloud/init.go
```diff
@@ -43,23 +45,27 @@ func NewInitCmd() *cobra.Command {
...
 					return
 				}
 			}
 
 			spinner := ui.NewSpinner("Installing Testkube")
 			err = common.HelmUpgradeOrInstallTestkubeCloud(options, cfg, false)
-			sendErrTelemetry(cmd, cfg, "helm_install")
-			ui.ExitOnError("Installing Testkube", err)
+			if err != nil {
+				sendErrTelemetry(cmd, cfg, "helm_install", err)
```
vsukhin, line 66: same question

cmd/kubectl-testkube/commands/pro/init.go
```diff
@@ -38,23 +40,28 @@ func NewInitCmd() *cobra.Command {
 				ui.NL()
 
 				currentContext, err := common.GetCurrentKubernetesContext()
-				sendErrTelemetry(cmd, cfg, "k8s_context")
-				ui.ExitOnError("getting current context", err)
+
+				if err != nil {
+					sendErrTelemetry(cmd, cfg, "k8s_context", err)
+					ui.ExitOnError("getting current context", err)
```
vsukhin, line 46: same question

vsukhin (commented):
one question

haneabogdan:
we're currently sending error telemetry, regardless if an error actually occurred. Hence, we have atm some metrics which indicate that 100% of the users who try to install an agent, encounter an error. Many of these are false positives.
Hence, I've changed the logic, to only `sendErrorTelemetry` events, if an error actually occurred,

vsukhin:
but, in this case,  we will not have telemetry for successful commands, is it okay?

haneabogdan:
that's true, but a different topic. we don't have any success telemetry even now. will sync with the team to see if we decide to add any success telemetry for this command

PR #5057, opened 2024-02-23
feat(TKC-1465): add expressions language for TestWorkflows

rangoo94:
## Pull request description 

Adds basic expressions language. It will allow us to do expressions in the `TestWorkflow` definitions, like:

```yaml
env:
- name: "SOME_NAME"
  value: "some{{env.value}}here"
- name: "COUNT"
  value: "{{ env.count || 1 }}"
retry:
  count: "{{ config.retries > 1 ? config.retries : 1 }}"
shell: |
  {{ config.npm ? "npm install" : "yarn" }}
  npm run {{ shellquote(config.script) }}
```

## Features

* Executing math and conditional operations
* Calling built-in and custom functions
* Partial resolution (simplify the expression, even when it's not possible to compute all of it yet)
* Safe conversion between `Expression` and `Template` string
  * Expression: `A + B + "abc"` == Template: `{{ A }}{{ B }}abc`
  * Expression: `A * B` == Template: `{{ A * B }}`
  * Expression: `123 + 456` == Template: `{{ 123 + 456 }}` (but is immediately simplified to template: `579`)
  * Expression: `string(123) + string(456)` == Template: `{{ 123 }}{{ 456 }}` (but is immediately simplified to template: `123456`)
* The language is superset over JSON - therefore inlined values are JSON based (so `"abc\ndef"` or `{ "xyz": "abc" }` is a valid expression too)
* Allow literal `\n` character (in JSON it has to be represented by `"\n"` but it's easier for us to allow `\n` i.e. in YAML/JSON definitions)
* Common built-in functions
  * Casting to `int(any)`, `string(any)`, `float(any)`, `bool(any)`
  * Building `list(any...)`s
  * List operations: `split(string[, string = ","])`, or `join(list[, string = ","])`
  * Reading from `json(string)` and marshalling `tojson(any)`
  * Reading from `yaml(string)` and marshalling `toyaml(any)`
  * Sanitizing `shellquote(string...)`s, and `trim(string)`ming spaces

## Why not existing language

There are languages like [**cel-go**](https://github.com/google/cel-go) or [**tengo**](https://github.com/d5/tengo), or even sandboxed [**LUA**](https://github.com/Shopify/go-lua) or [**JS**](https://github.com/dop251/goja). Of course, they have advantage of external maintenance, but they are more complex than we need (inconvenient, but not a blocker) and **don't support partial resolution (blocker)**.

The partial resolution is needed, because part of the expression may need to be resolved before the TestWorkflow execution, but another part during the execution.

As an example, imagine a case:

```yaml
steps:
  - condition: "passed && config.enabled"
    shell: "npm test"
```

Now, while the `config.enabled` is available before the execution, the `passed` is a dynamic value that's not known until the step is executed.

With partial resolution, we will i.e. receive `condition: "passed"` when the `config.enabled=true` for the step, so the step itself doesn't need to know anything more than the actual dynamic context (`passed`).

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test **(80+% of coverage, 1200+ LoC for mocks and unit tests)**

## Fixes

- https://linear.app/kubeshop/issue/TKC-1465/testworkflows-expressions-engine


pkg/tcl/expressionstcl/call.go
```diff
@@ -0,0 +1,129 @@
...
+	for i := range args {
+		if args[i] == nil {
+			args[i] = None
+		}
+	}
+	return &call{name: name, args: args}
+}
+
+func (s *call) WillBeString() bool {
+	return s.name == "string"
```
vsukhin, line 32: should it be const?

rangoo94: Thanks, good point.

This is kind-of small workaround, to simplify the generated schema for known stdlib function; after you've mentioned it, I noticed that there are more functions that will be always string, like `join`, `tojson` or `toyaml`.

Which option do you think will be better:
* add `var knownStringFunctions = []string{"string", "tojson"} // etc` on top of `stdlib.go` and use it
* adjust `stdFunctions` to be map[string]struct{willBeString: bool, handler: func(...StaticValue) (Expression, error)}
?

vsukhin (approved):

pkg/tcl/expressionstcl/call.go
vsukhin: second one looks more consistent to me.

rangoo94: Pushed the change 👍

Added:
```go
type StdFunction struct {
	ReturnType string
	Handler    func(...StaticValue) (Expression, error)
}
```

And used it in `stdFunctions`

rangoo94: To make it more flexible, I have also replaced `StringAwareExpression.WillBeString() bool` with `Expression.Type() Type`, so every expression may by itself inform about the result type. This way in the future we could apply better reflection - both for validation and error messages, as well as predictions and optimizations.

PR #5090, opened 2024-02-29
feat(TKC-1580): prepare Init Process for TestWorkflow containers

rangoo94:
## Pull request description 

Prepare Init Process that will wrap all TestWorkflows operations in the containers. Additionally, it includes some fixes and improvements to the Expressions Engine (force simplifying unknown structs, better boolean prediction).

It supports:
* Skipping container execution
* Passing status to parents/children
* Sharing state across different containers
* Aborting steps
* Negative expectation
* Timeouts (nested already)
* Retry (single step only)
* Computing variables locally, i.e. reading files (`file("/data/some/path")`), reading shared data

In the future, we need to support:
* Nested retries (sharing iterations is already there)
* Obfuscating sensitive data in the output
* Saving the data from the output clauses in the shared state

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cmd/tcl/testworkflow-init/data/state.go
+++ b/cmd/tcl/testworkflow-init/data/state.go
@@ -112,10 +112,10 @@ func recomputeStatuses() {
 	State.GetStep(Step.Ref).SetStatus(status)
 
 	// Update expected failure statuses
-	for _, r := range Config.Resulting {
+	Iterate(Config.Resulting, func(r Rule) bool {
 		v, err := RefStatusExpression(r.Expr)
 		if err != nil {
-			continue
+			return false
 		}
 		vv, _ := v.Static().BoolValue()
 		if !vv {
@@ -127,7 +127,8 @@ func recomputeStatuses() {
 				}
 			}
 		}
-	}
+		return true
+	})
 }
 
 func persistStatus(filePath string) {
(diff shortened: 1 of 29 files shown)


cmd/tcl/testworkflow-init/main.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			fmt.Printf(`resolving command: %s: %s\n`, shellquote.Join(original...), err.Error())
+			os.Exit(155)
+		}
+	}
+
+	// Fail when there is nothing to run
+	if len(args) == 0 {
+		fmt.Println("missing command to run")
+		os.Exit(189)
```
vsukhin, line 172: I would use const

cmd/tcl/testworkflow-init/data/state.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	str, _ := v.Static().StringValue()
+	return str
+}
+
+func readState(filePath string) {
+	b, err := os.ReadFile(filePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			panic(err.Error())
```
vsukhin, line 84: do we need panic everywhere?

vsukhin (approved):

cmd/tcl/testworkflow-init/data/state.go
rangoo94: Basically, these are clearly not User issues, but likely actual implementation bugs if happens, nothing that User may deal with

rangoo94: Added `Failf` helper, and used in in places where error code is expected; `panic` is left in:

* places which could be used by other programs (i.e. emit)
* places where something unusual happened, likely due implementation bug (i.e. state not saved), not user error

PR #5096, opened 2024-03-03
feat(TKC-1581): add mechanism to build Kubernetes resources for the TestWorkflow

rangoo94:
## Pull request description 

* Add mechanism to build (and deploy) Kubernetes resources for the TestWorkflows
* Add API endpoint for running the TestWorkflows
  * Almost dummy - it doesn't return any data yet, and it's not persisting data
* Add basic processors for some operations
  * For the moment, only operations that do not require an additional toolkit: nested steps, content files, and running the containers
* Understands groups, conditions, optional steps, and all other syntaxes

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Fixes

- https://linear.app/kubeshop/issue/TKC-1581/testworkflows-mechanism-to-build-kubernetes-resources


exu (approved):

pkg/tcl/testworkflowstcl/testworkflowprocessor/constants.go
```diff
@@ -0,0 +1,54 @@
...
+	"path/filepath"
+
+	corev1 "k8s.io/api/core/v1"
+
+	testworkflowsv1 "github.com/kubeshop/testkube-operator/api/testworkflows/v1"
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
+)
+
+const (
+	defaultImage        = "busybox:1.36.1"
```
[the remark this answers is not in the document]
rangoo94: Default is already configurable on the TestWorkflow level, as you may set top-level `container: { image: "another-default-image" }`. We could add some environment variable on API Server level to default the default though, if you want 🙂 I'm not sure if it's worth too, considering that we will likely have some `global` TestWorkflowTemplate, that will be automatically added to all of the TestWorkflows, where it could be configured (with top-level `container` settings, as above).

When it comes to the other properties, like internal or data directory, I'm considering that we could add in the future option to pass another property to the TestWorkflow, like `internal: { dataDirectory: {path string, volumeName string} }` or something like that

pkg/tcl/testworkflowstcl/testworkflowprocessor/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/kubeshop/testkube/internal/common"
+	"github.com/kubeshop/testkube/pkg/imageinspector"
+)
+
+func AnnotateControlledBy(obj metav1.Object, testWorkflowId string) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels["testworkflowid"] = testWorkflowId
```
vsukhin, line 28: I would make this a const

pkg/tcl/apitcl/v1/testworkflows.go
```diff
@@ -236,6 +241,83 @@ func (s *apiTCL) PreviewTestWorkflowHandler() fiber.Handler {
...
+			_, err = s.Clientset.CoreV1().Secrets(s.Namespace).Create(context.Background(), &item, metav1.CreateOptions{})
+			if err != nil {
+				return s.BadRequest(c, errPrefix, "creating secret", err)
+			}
+		}
+		for _, item := range bundle.ConfigMaps {
+			_, err = s.Clientset.CoreV1().ConfigMaps(s.Namespace).Create(context.Background(), &item, metav1.CreateOptions{})
+			if err != nil {
+				return s.BadRequest(c, errPrefix, "creating configmap", err)
+			}
```
vsukhin, line 308: do we clear them after job finish?

rangoo94: We will, but not at this point. The [**TKC-1651**](https://linear.app/kubeshop/issue/TKC-1651/testworkflows-prepare-controller-for-testworkflows-in-kubernetes) ticket is for building the Controller, which will be handling the Job/Pod/Containers events, logs, and the cleanup 👍

pkg/tcl/testworkflowstcl/testworkflowprocessor/initprocess.go
```diff
@@ -0,0 +1,207 @@
...
+	return p
+}
+
+func (p *initProcess) Command() []string {
+	args := p.params
+
+	// TODO: Support nested retries
+	policy, ok := p.retry[p.ref]
+	if ok {
+		args = append(args, "--retryCount", strconv.Itoa(int(policy.Count)), "--retryUntil", expressionstcl.Escape(policy.Until))
```
vsukhin, line 62: a list of constants shared with init image would help to make naming consistent between them

pkg/tcl/testworkflowstcl/testworkflowprocessor/operations.go
```diff
@@ -0,0 +1,139 @@
...
+	if step.Delay == "" {
+		return nil, nil
+	}
+	t, err := time.ParseDuration(step.Delay)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("invalid duration: %s", step.Delay))
+	}
+	resources := map[corev1.ResourceName]intstr.IntOrString{
+		corev1.ResourceCPU:    {Type: intstr.String, StrVal: "50m"},
+		corev1.ResourceMemory: {Type: intstr.String, StrVal: "4Mi"},
```
vsukhin, line 32: shouldn't it be configurable params?

pkg/tcl/testworkflowstcl/testworkflowprocessor/initprocess.go
rangoo94: `--retryCount` and `--retryUntil` are hackish for this moment, as they will be replaced in the final implementation. At the moment we support retry only on single-level, while the retry should be supported in multi-step structure as well:
```yaml
steps:
- name: Run something
  retry:
    count: 5
  steps:
  - name: Run test
    retry:
      count: 5 # totaling in max of 25 retries
  - name: Run another test
    retry:
      count: 5 # totaling in max of 25 retries minus the top retries used already by the previous step
```

We could considering adding some constants though, which would be imported by the `InitProcess` 👍

pkg/tcl/testworkflowstcl/testworkflowprocessor/operations.go
rangoo94: I don't think so, but maybe I'm wrong. It's for a very specific case of delay, that it's better to limit to the lowest possible resources.

pkg/tcl/testworkflowstcl/testworkflowprocessor/processor.go
```diff
@@ -0,0 +1,286 @@
...
+		Steps: append(workflow.Spec.Setup, append(workflow.Spec.Steps, workflow.Spec.After...)...),
+	}
+	root, err := p.process(layer, layer.ContainerDefaults(), rootStep, "")
+	if err != nil {
+		return nil, errors.Wrap(err, "processing error")
+	}
+
+	s := root.Signature()
+	v, _ := json.Marshal(s)
+	fmt.Println(string(v))
```
vsukhin, line 120: is this debugging?

rangoo94: Thanks, it is! Deleted 👍

vsukhin (approved):

pkg/tcl/testworkflowstcl/testworkflowprocessor/utils.go
rangoo94: Changed to constant 👍

pkg/tcl/testworkflowstcl/testworkflowprocessor/initprocess.go
rangoo94: Added constants and reused them (with `TODO` note for retries to delete it) 👍

pkg/tcl/testworkflowstcl/testworkflowprocessor/constants.go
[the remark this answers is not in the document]
vsukhin: ok, let's see, if we have any questions about it

pkg/tcl/apitcl/v1/testworkflows.go
vsukhin: great

Issue #5048, opened 2024-02-22
Container executor - post-run script isn't run after command (test) fails


tkonieczny:
**Describe the bug**
Post-run script isn't run after command (test) fails in case of container executor. 
Example:
https://demo.testkube.dev/tests/expected-fail-container-pre-post-run-script - pre-run script is executed, test fails, post-run script isn't executed at all

Everything is fine for prebuilt executors (https://demo.testkube.dev/tests/expected-fail-pre-post-run-script/executions/65d70339d23cbe05dcebef98)

**Expected behavior**
Post-run script should be executed even if command failed (and the whole test should be marked as failed then)

vsukhin:
released

PR #5097, opened 2024-03-04
fix: [TKC-1611] execute post run script for container executors

povilasv:
## Pull request description 

Change the entrypoint generation to get rid of `set -e` and then store the exit code of the command and still run the post run script.

Fix  https://github.com/kubeshop/testkube/issues/5048

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/contrib/executor/init/pkg/runner/runner.go
+++ b/contrib/executor/init/pkg/runner/runner.go
@@ -81,27 +81,38 @@ func (r *InitRunner) Run(ctx context.Context, execution testkube.Execution) (res
 		}
 
 		shebang := "#!" + shell + "\nset -e\n"
-		entrypoint := shebang
+		// No set -e so that we can run the post-run script even if the command fails
+		entrypoint := "#!" + shell + "\n"
 		command := shebang
 		preRunScript := shebang
 		postRunScript := shebang
 
 		if execution.PreRunScript != "" {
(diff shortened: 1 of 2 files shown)


contrib/executor/init/pkg/runner/runner.go
```diff
@@ -81,7 +81,8 @@ func (r *InitRunner) Run(ctx context.Context, execution testkube.Execution) (res
 		}
 
 		shebang := "#!" + shell + "\nset -e\n"
-		entrypoint := shebang
+		// No set -e so that we can run the post-run script even if the command fails
+		entrypoint := "#!" + shell + "\n"
```
rangoo94, line 85: Failure of pre-run script should rather also lead to failure, without running the test

contrib/executor/init/pkg/runner/runner.go
```diff
@@ -102,6 +104,10 @@ func (r *InitRunner) Run(ctx context.Context, execution testkube.Execution) (res
 			postRunScript += execution.PostRunScript
 		}
 
+		if len(execution.Command) != 0 {
+			entrypoint += "exit $command_exit_code\n"
+		}
```
rangoo94, line 109: That means that when there is no command, it will always succeed

vsukhin (approved):

PR #5136, opened 2024-03-11
fix: run TestWorkflows artifacts step as root, to read FS without problems

rangoo94:

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/pkg/tcl/testworkflowstcl/testworkflowprocessor/container.go
+++ b/pkg/tcl/testworkflowstcl/testworkflowprocessor/container.go
@@ -73,6 +73,7 @@ type ContainerMutations[T any] interface {
 	ApplyCR(cr *testworkflowsv1.ContainerConfig) T
 	ApplyImageData(image *imageinspector.Info) error
 	EnableToolkit(ref string) T
+	RunAsRoot() T
 }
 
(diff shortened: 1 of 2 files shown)


pkg/tcl/testworkflowstcl/testworkflowprocessor/container.go
```diff
@@ -421,6 +422,13 @@ func (c *container) EnableToolkit(ref string) Container {
 	})
 }
 
+func (c *container) RunAsRoot() Container {
```
vsukhin, line 425: this might produce questions for OpenShift and other rootless confogurations

rangoo94: Sure thing, but we already have this problem - current artifacts scraper also works as a root:
* [**Dockerfile**](https://github.com/kubeshop/testkube/blob/6daf1ab9cada7969a2f3cfe667d8706244089d97/contrib/executor/scraper/build/agent/Dockerfile) is using default Alpine user == root
* [**Template**](https://github.com/kubeshop/helm-charts/blob/develop/charts/testkube-api/job-scraper-template.yml.tmpl) doesn't have any `securityContext` overrides, so it keeps this way

The problem is, that you need to have permissions to read the artifacts, so you either need to be root user, or you need to use same user for all the steps, or you need to apply proper permissions in each step

PR #5176, opened 2024-03-13
feat: workflow tests - negative cases

tkonieczny:
Negative cases (expected failures):
- test failure, `condition: always` for artifacts (no artifacts)
- test failure, `condition: always` for artifacts (with artifacts)
- resource limits (pod is terminated because of that)
- incorrect workingDir
- incorrect repo path

### DIFF
--- a/test/cypress/executor-tests/crd-workflow/smoke.yaml
+++ b/test/cypress/executor-tests/crd-workflow/smoke.yaml
@@ -241,3 +241,155 @@ spec:
       artifacts:
         paths:
         - '**/*'
+---
+apiVersion: testworkflows.testkube.io/v1
+kind: TestWorkflow
+metadata:
+  name: expected-fail-cypress-incorrect-repo-path
+  labels:
+    core-tests: expected-fail
+spec:
+  content:
+    git:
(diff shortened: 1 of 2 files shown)


exu (approved):

test/cypress/executor-tests/crd-workflow/smoke.yaml
```diff
@@ -241,3 +241,155 @@ spec:
...
+      - '{"screenshotsFolder":"/data/artifacts/screenshots","videosFolder":"/data/artifacts/videos"}'
+      env:
+      - name: CYPRESS_CUSTOM_ENV
+        value: CYPRESS_CUSTOM_ENV_value
+    steps:
+    - name: Saving artifacts
+      workingDir: /data/artifacts
+      artifacts:
+        paths:
+        - '**/*'
```
rangoo94, line 281: Could be as well:
```suggestion
    artifacts:
      workingDir: /data/artifacts
      paths:
      - '**/*'
```

Unless it shouldn't be run on `run` failure

rangoo94 (approved):
One comment, but similarly to all the other examples

PR #5277, opened 2024-04-04
fix: workflow telemetry pointer

vLia:
## Pull request description 

Catch nil pointers and refactor the workflow metrics code

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/tcl/apitcl/v1/testworkflowmetrics.go
+++ b/pkg/tcl/apitcl/v1/testworkflowmetrics.go
@@ -24,55 +24,19 @@ func (s *apiTCL) sendCreateWorkflowTelemetry(ctx context.Context, workflow *test
 		return
 	}
 
-	clusterID, err := s.configMap.GetUniqueClusterId(ctx)
-	if err != nil {
-		log.DefaultLogger.Debugf("getting cluster id error", "error", err)
-	}
-
-	host, err := os.Hostname()
-	if err != nil {
-		log.DefaultLogger.Debugf("getting hostname error", "hostname", host, "error", err)
-	}
-
-	var dataSource string
-	isKubeshopGitURI := false
-	if workflow.Spec.Content != nil {
-		if len(workflow.Spec.Content.Files) != 0 {
-			dataSource = "files"
-		} else if workflow.Spec.Content.Git != nil {
-			dataSource = "git"
-			if strings.Contains(workflow.Spec.Content.Git.Uri, "kubeshop") {
-				isKubeshopGitURI = true
-			}
-		}
-	}
-
-	hasArtifacts := false
-	for _, step := range workflow.Spec.Steps {
-		if step.Artifacts != nil {
-			hasArtifacts = true
-			break
-		}
-	}
-
-	image := ""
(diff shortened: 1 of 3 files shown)


pkg/tcl/apitcl/v1/testworkflowmetrics.go
```diff
@@ -166,61 +98,206 @@ func (s *apiTCL) sendRunWorkflowTelemetry(ctx context.Context, workflow *testwor
...
+	for _, step := range step.Steps {
+		if isKubeshopGitURI(step.Content) {
+			return true
+		}
+		if hasStepKubeshopGitURI(step) {
+			return true
+		}
+	}
+	return false
+}
```
rangoo94, line 266: Maybe worth to unify a walker, like:

```go
func hasStepLike(steps []testworkflowsv1.Step, fn (step testworkflowsv1.Step) bool) bool {
  for _, step := range steps {
    if fn(step) || hasStepLike(step.Setup, fn) || hasStepLike(step.Steps, fn) {
      return true
    }
}
```
and similar for template steps

pkg/tcl/apitcl/v1/testworkflowmetrics.go
```diff
@@ -166,61 +98,206 @@ func (s *apiTCL) sendRunWorkflowTelemetry(ctx context.Context, workflow *testwor
...
+	switch {
+	case content == nil:
+		return false
+	case content.Git == nil:
+		return false
+	case strings.Contains(content.Git.Uri, "kubeshop"):
+		return true
+	default:
+		return false
+	}
```
rangoo94, line 279: Maybe just:
```suggestion
	return content != nil && content.Git != nil && strings.Contains(content.Git.Uri, "kubeshop")
```

pkg/tcl/apitcl/v1/testworkflowmetrics.go
```diff
@@ -166,61 +98,206 @@ func (s *apiTCL) sendRunWorkflowTelemetry(ctx context.Context, workflow *testwor
...
+
+// getDataSource returns the data source of the content
+func getDataSource(content *testworkflowsv1.Content) string {
+	var dataSource string
+	if content != nil {
+		if len(content.Files) != 0 {
+			dataSource = "files"
+		} else if content.Git != nil {
+			dataSource = "git"
+		}
```
rangoo94, line 290: Actually both can be used at the same time. Probably worth to prioritize Git then

rangoo94 (approved):
Added some improvement requests, but anyway it's an improvement so LGTM 👍

rangoo94 (approved):
Committed some updates. I will merge it, so we will unblock `develop`, but we should in future revisit that:
* `TestWorkflowImage` - it's checking only default container, so ignores templates and inner steps
* `TestWorkflowSteps` - doesn't count nested steps, additionally some steps are automatically split (i.e. artifacts + shell + run == 3 actual steps)

PR #5312, opened 2024-04-18
feat: fetch pro context from cloud-api

WitoDelnat:
## Pull request description 

This PR fetches details for the ProContext from the cloud-api as an alternative to specifying it within environment variables. This is a stepping stone to deploy the control-plane and agent in parallel without dependencies between each other.

## Testing

Screenshot 2024-04-19 at 12 19 14


dejanzele:
Should we re-use the `Call` endpoint (https://github.com/kubeshop/testkube/blob/develop/proto/service.proto#L15) and add a new command or add a new endpoint?

cmd/api-server/main.go
```diff
@@ -831,3 +821,36 @@ func newGRPCTransportCredentials(cfg *config.Config) (credentials.TransportCrede
...
+		EnvID:                            cfg.TestkubeProEnvID,
+		OrgID:                            cfg.TestkubeProOrgID,
+		Migrate:                          cfg.TestkubeProMigrate,
+		ConnectionTimeout:                cfg.TestkubeProConnectionTimeout,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
+	getProContext, err := grpcClient.GetProContext(ctx, &emptypb.Empty{})
+	if err != nil {
```
dejanzele, line 843: maybe log with a `Warn` in case of an error

dejanzele (approved):

Issue #5279, opened 2024-04-05
Testkube-api downloads logs artifacts under incorrect S3 bucket path


KrPetrovDev:
**Describe the bug**
When I create and execute a simple curl-test from the testkube-dashboard UI, I cannot download the logs from the execution despite that I see that they are stored correctly in an S3 bucket in AWS.
Also this log artifact which has the following name `660fb2ff8358704f8a3f1d07` If I change It to be inside a directory named like the artifact which be resulting in the following path `660fb2ff8358704f8a3f1d07/660fb2ff8358704f8a3f1d07` testkube-api is able to install It correctly.

So the issue is that the way testkube-api has to upload the artifacts to S3 is different from the one when downloading them.

And as additional data I can see from testkube-api the following log line when trying to download the artifact:
`{"level":"info","ts":1712306203.6748464,"caller":"minio/minio.go:159","msg":"listing files","bucket":"s3-bucket","bucketFolder":"660fb2ff8358704f8a3f1d07"}`

**To Reproduce**
1. Create kubernetes cluster
2. Install testkube versions that I mentioned below
3. Make testkube-api and testkube-dashboard reachable from outside of the cluster
4. create the following yaml test to be executed from the testkube-dashboard:

> apiVersion: tests.testkube.io/v3
> kind: Test
> metadata:
>   name: curl-test
>   namespace: app-project
>   labels:
>     executor: curl-executor
>     test-type: curl-test
> spec:
>   type: curl/test
>   content:
>     type: string
>     data: "{\n \"command\": [\n  \"curl\",\n\"https://dummy.restapiexample.com/api/v1/employee/1\",\n  \"-H\",\n  \"Accept: application/json\"\n ],\n \"expected_status\": \"200\"\n}"
> status:
>   latestExecution:
>     id: 660fb2ff8358704f8a3f1d07
>     number: 1
>     startTime: 2024-04-05T08:14:55Z
>     endTime: 2024-04-05T08:14:58Z
>     status: passed

**Expected behavior**
I expected from testkube-api to know where is actually located the artifact and download It from there, which is this path `660fb2ff8358704f8a3f1d07` instead of this path `660fb2ff8358704f8a3f1d07/660fb2ff8358704f8a3f1d07`

**Version / Cluster**
- Which testkube version?
- testkube helm chart version is 1.17.9
- testkube-dashboard helm chart version is 1.16.7 
- testkube-api pod using the following docker image tag: 1.17.7
- What Kubernetes cluster?
[...]

PR #5327, opened 2024-04-23
fix: use log artifact path w/o prefix in minio

sngchlko:
## Pull request description 
When listing artifacts in MinIO, objects with the execution ID as a prefix are searched. For example, when the prefix is `660fb2ff8358704f8a3f1d07`, the list of objects is as follows: `660fb2ff8358704f8a3f1d07/artifact` and `660fb2ff8358704f8a3f1d07`.
When downloading artifacts, `downloadFile()` is assumed that all files are in the specific directory.
This assumption is not correct because the log artifact name itself is also execution ID.

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes
-

## Changes
-

## Fixes
- https://github.com/kubeshop/testkube/issues/5279

### DIFF
--- a/pkg/storage/minio/minio.go
+++ b/pkg/storage/minio/minio.go
@@ -259,7 +259,7 @@ func (c *Client) downloadFile(ctx context.Context, bucket, bucketFolder, file st
 		return nil, ErrArtifactsNotFound
 	}
 
-	if bucketFolder != "" {
+	if bucketFolder != "" && bucketFolder != file {
 		file = strings.Trim(bucketFolder, "/") + "/" + file
 	}
 


pkg/storage/minio/minio.go
```diff
@@ -259,7 +259,7 @@ func (c *Client) downloadFile(ctx context.Context, bucket, bucketFolder, file st
 		return nil, ErrArtifactsNotFound
 	}
 
-	if bucketFolder != "" {
+	if bucketFolder != "" && bucketFolder != file {
```
vsukhin, line 262: In general, nothing is wrong usinf the same name for bucket and file, the problem is appeared when storing logs and artifacts together

sngchlko: Thank you for the good comment. As you said, it seems better to separate the log and artifact buckets. Then even if the user sets up to use only one bucket, it seems like testkube need to create a bucket for logs somewhere. Did I understand correctly?

vsukhin: yes

sngchlko: Good. I will close this PR. Could I contribute this feature if it is not hard for newbie. Could you tell me your opinion about the contribution?

PR #5328, opened 2024-04-23
feat: add support for searching and sending junit reports to cloud

dejanzele:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/api/v1/testkube/model_test_workflow_execution.go
```diff
@@ -28,7 +28,9 @@ type TestWorkflowExecution struct {
 	Signature []TestWorkflowSignature `json:"signature,omitempty"`
 	Result    *TestWorkflowResult     `json:"result,omitempty"`
 	// additional information from the steps, like referenced executed tests or artifacts
-	Output           []TestWorkflowOutput `json:"output,omitempty"`
+	Output []TestWorkflowOutput `json:"output,omitempty"`
+	// generated reports from the steps, like junit
+	Report           []TestWorkflowReport `json:"report,omitempty"`
```
rangoo94, line 33: `reports`?

dejanzele: Yep, should be plural. I am still not sure could we have other reports besides junit, or have this generic structure for report, where you have a `kind` field which can denote the report type.

cmd/tcl/testworkflow-toolkit/commands/artifacts.go
```diff
@@ -94,18 +97,24 @@ func NewArtifactsCmd() *cobra.Command {
...
 			}
 
+			var handlerOpts []artifacts.HandlerOpts
 			// Archive
 			if env.CloudEnabled() {
+				ctx, cancel := context.WithTimeout(cmd.Context(), 30*time.Second)
+				defer cancel()
+				client := env.Cloud(ctx)
+				defer client.Close()
+				handlerOpts = append(handlerOpts, artifacts.WithPostProcessor(artifacts.NewJUnitPostProcessor(filesystem.NewOSFileSystem(), client)))
```
rangoo94, line 107: Shouldn't this be optional? Otherwise, when the Agent is updated, but the Cloud API is not, the step will break (as postprocessor will fail)

cmd/tcl/testworkflow-toolkit/commands/artifacts.go
```diff
@@ -94,18 +97,24 @@ func NewArtifactsCmd() *cobra.Command {
...
 				if compress != "" {
 					processor = artifacts.NewTarCachedProcessor(compress, compressCachePath)
```
rangoo94, line 109: I think that it should not be enabled when compression is enabled, as you won't be able to read the file

rangoo94 (approved):

cmd/tcl/testworkflow-toolkit/commands/artifacts.go
dejanzele: that makes sense

cmd/tcl/testworkflow-toolkit/commands/artifacts.go
dejanzele: we don't have an actual pipeline like `handler: processor -> uploader -> post-processor`, it is more 2 "parallel" flows like:
```
handler:
  - processor -> uploader
  - post-processor
```
meaning post-processing will be run on the same files which get traversed by the walker, and not on the output of the processor.

That way the junit post-processor can work with any processor.

Now when I think of it, I don't know is post-processor the best name, at least it is current best :D

dejanzele: I don't know will we do something else which would require processing on the same files. To be fair, it technically is a pre-processor, as it processes the input files, post-processor would logically process the files outputted by processor.

I like Analyzer tho, we can always rename this interface, it's impact is localized

cmd/tcl/testworkflow-toolkit/artifacts/junit_post_processor.go
```diff
@@ -0,0 +1,117 @@
...
+		return errors.Wrapf(err, "failed to send JUnit report %s", stat.Name())
+	}
+	return nil
+}
+
+// sendJUnitReport sends the JUnit report to the Agent gRPC API.
+func (p *JUnitPostProcessor) sendJUnitReport(path string, report []byte) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	_, err := p.client.Execute(ctx, testworkflow.CmdTestWorkflowExecutionAddReport, &testworkflow.ExecutionsAddReportRequest{
```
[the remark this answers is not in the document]
dejanzele: I thought about it initially, I wanted it to be event-driven and do processing in Cloud, but from what I saw, I think it would require more work and would have more moving pieces, example:
- are the files already uploaded?
- are the files unpacked?
- if no to any of those, we need to persist the logic in db and retry until complete

with current implementation is cheaper to iterate through XML files and check them is any of them a JUnit report, it would be costlier to do the same logic on remote S3, unless we search by regex or something similar, which has edge cases on its own.

Ultimately, I'd like stuff like this to be event-driven and in Cloud, but it would require more work than current. I'd start with this, and gradually move to more event-driven stuff, as also it would unblock graphql work.

PR #5333, opened 2024-04-23
feat: refactor test workflow execute

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/tcl/commontcl/common.go
rangoo94, line 1: Most of these should be rather in `testworkflowtcl`, as they are not common

vsukhin: ok, will move

PR #5340, opened 2024-04-25
feat(testworkflows): support more advanced Pod configuration

rangoo94:
## Pull request description 

* Support more advanced Pod configuration for TestWorkflows
* https://github.com/kubeshop/testkube-operator/pull/247

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


vsukhin (approved):
looks good, but for any changes or new fields we will need to update the code

rangoo94:
> looks good, but for any changes or new fields we will need to update the code

We can think of alternatives to lower the amount of the code:
* For OpenAPI - we can use just `map[string]interface{}` instead
   * We will have less strict OpenAPI, and the conversion back to CRD may be hard
* For merging - we can make it more generic with the reflection
   * But then we will have less precise merging

Overall, adding new properties still won't be automatic 😕

PR #5378, opened 2024-05-07
fix: issue with image pull secrets for workflows

dejanzele:
## Pull request description 

fix issue with image pull secrets for workflows

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/skopeo/client.go
+++ b/pkg/skopeo/client.go
@@ -86,14 +86,16 @@ func (c *client) Inspect(registry, image string) (*DockerImage, error) {
 	}
 
 	if len(c.dockerAuthConfigs) != 0 {
+		// TODO: Is it a good idea to randomly select a secret?
 		i := rand.Intn(len(c.dockerAuthConfigs))
 		args = append(args, "--creds", c.dockerAuthConfigs[i].Username+":"+c.dockerAuthConfigs[i].Password)
 	}
 
-	config := "docker://" + image
-	if registry != "" {
-		config = registry + "/" + image
+	// If registry is provided via config and the image does not start with the registry, prepend it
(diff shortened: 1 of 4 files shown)


rangoo94 (approved):

pkg/skopeo/client.go
```diff
@@ -86,14 +86,12 @@ func (c *client) Inspect(registry, image string) (*DockerImage, error) {
 	}
 
 	if len(c.dockerAuthConfigs) != 0 {
+		// TODO: Is it a good idea to randomly select a secret?
 		i := rand.Intn(len(c.dockerAuthConfigs))
 		args = append(args, "--creds", c.dockerAuthConfigs[i].Username+":"+c.dockerAuthConfigs[i].Password)
 	}
 
 	config := "docker://" + image
-	if registry != "" {
```
vsukhin, line 94: where is registry used, if it's removed there?

dejanzele: now nowhere

vsukhin: guess, it's wrong, users asked us about usage of private registries, if it's provided

dejanzele: I removed the append because first, we never set the `registry` field so it was always `""`, and the registry part is already contained in the `image` param, so it just doubled and made it incorrect

vsukhin: you mean, we don't set in our clustres? but our users use registry for long time. 
if we have it already in the image name, we can leave it empty or check, if it's already embedded and skip adding

vsukhin (approved):

PR #5401, opened 2024-05-08
feat: istio compatibility

emil2k:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


contrib/executor/init/cmd/agent/main.go
```diff
@@ -6,17 +6,20 @@ import (
 
 	"github.com/pkg/errors"
 
-	"github.com/kubeshop/testkube/contrib/executor/init/pkg/runner"
+	initrunner "github.com/kubeshop/testkube/contrib/executor/init/pkg/runner"
```
emil2k, line 9: This should not be in the init command.

internal/config/config.go
```diff
@@ -104,6 +104,18 @@ type Config struct {
 	TestkubeExecutionNamespaces                 string        `envconfig:"TESTKUBE_EXECUTION_NAMESPACES" default:""`
 	TestkubeDefaultStorageClassName             string        `envconfig:"TESTKUBE_DEFAULT_STORAGE_CLASS_NAME" default:""`
 	GlobalWorkflowTemplateName                  string        `envconfig:"TESTKUBE_GLOBAL_WORKFLOW_TEMPLATE_NAME" default:""`
+	// IstioProxyWait triggers logic in the test runner jobs to wait for Istio'
+	// Envoy proxy to become ready.
+	// This should only be necessary with Istio versions before 1.7, after
+	// which one should use either the global (1.7) or pod-level (1.8)
+	// holdApplicationUntilProxyStarts setting.
```
emil2k, line 111: Change these docs to include that some people might prefer not to use this as it changes the default container for the `kubectl` commands.

emil2k:
We are no longer pursuing this.

PR #5450, opened 2024-05-17
feat: scraper sidecar option

### LINKED ISSUE #4937: Avoid PVC creation and upload artifacts from the job runner
**Is your feature request related to a problem? Please describe.**
I'm launching hundreds of tests in my clusters and each tests require a PVC creation.
Intense and continuous PVC creation/deletion can be very painful for a storage system (e.g. network storage solution).
Moreover, if you are in cloud (e.g. AWS) you have to pay for this temporary storage.
Why you cannot just upload the artifact by the job runner itself into Minio/S3?
What are the benefits of having a separate scraper?

**Describe the solution you'd like**
I would like to avoid the creation of a PVC and be able to upload the artifacts directly from the job runner.
So I want the scraper feature (e.g. upload artifact to Minio/s3) will be handled by the job runner itself.

In the end, I would like to know if in your roadmap you are going to provide a solution that doesn't use the PVC or if can you suggest a way to avoid the creation.

Thanks
Bye

vsukhin:
## Pull request description 

Fix #4937 

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


contrib/executor/scraper/pkg/runner/runner.go
```diff
@@ -88,3 +123,33 @@ func (r *ScraperRunner) Run(ctx context.Context, execution testkube.Execution) (
 func (r *ScraperRunner) GetType() runner.Type {
 	return runner.TypeFin
 }
+
+// isContainerTerminated checks if pod container is terminated through kubernetes API
+func isContainerTerminated(clientset kubernetes.Interface, podName, containerName, namespace string) wait.ConditionWithContextFunc {
```
nicufk, line 128: Can be moved to k8sclient

nicufk (approved):
LGTM

contrib/executor/scraper/pkg/runner/runner.go
vsukhin: good idea

PR #5472, opened 2024-05-21
feat: [TKC-1948] disable webhooks

vLia:
## Pull request description 

This PR adds the option to disable webhooks and makes sure that no notification is sent out on a disabled webhook.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/event/kind/webhook/listener.go
+++ b/pkg/event/kind/webhook/listener.go
@@ -22,7 +22,7 @@ import (
 var _ common.Listener = (*WebhookListener)(nil)
 
 func NewWebhookListener(name, uri, selector string, events []testkube.EventType,
-	payloadObjectField, payloadTemplate string, headers map[string]string) *WebhookListener {
+	payloadObjectField, payloadTemplate string, headers map[string]string, disabled bool) *WebhookListener {
 	return &WebhookListener{
 		name:               name,
 		Uri:                uri,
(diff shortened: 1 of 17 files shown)


pkg/event/kind/webhook/listener.go
```diff
@@ -83,7 +86,16 @@ func (l *WebhookListener) Headers() map[string]string {
...
 }
 
+func (l *WebhookListener) Disabled() bool {
+	return l.disabled
+}
+
 func (l *WebhookListener) Notify(event testkube.Event) (result testkube.EventResult) {
+	if l.disabled {
+		l.Log.With(event.Log()...).Debug("webhook listener is disabled")
+		return testkube.NewFailedEventResult(event.Id, errors.New("webhook listener is disabled"))
```
vsukhin, line 96: do we need to return an error?

vsukhin (approved):
one question

pkg/event/kind/webhook/listener.go
vLia: The notification failed, even if it was expected to fail, that is why I chose to return a failed event result

vsukhin: I would not report an error, because we expect, it would not work

jfermi (approved):

pkg/event/kind/webhook/listener.go
vLia: Makes sense. I updated it.

PR #5478, opened 2024-05-22
feat: [TKC-1840] test workflow execution cr

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


api/v1/testkube.yaml
```diff
@@ -7488,6 +7489,9 @@ components:
           description: custom execution name
         config:
           $ref: "#/components/schemas/TestWorkflowConfigValue"
+        testWorkflowExecutionName:
```
rangoo94, line 7492: not important: maybe simpler to use just `executionName`?

docs/docs/articles/test-executions.md
```diff
@@ -1,10 +1,10 @@
-# Test and Test Suite Execution CRDs
+# Test, Test Suite and Test Workflow Execution CRDs
```
rangoo94, line 1: Worth to migrate it to [**testkube-docs**](https://github.com/kubeshop/testkube-docs)

pkg/tcl/testworkflowstcl/testworkflowexecutor/executor.go
```diff
@@ -148,6 +151,22 @@ func (e *executor) Recover(ctx context.Context) {
 	}
 }
 
+func (e *executor) updateStatus(execution *testkube.TestWorkflowExecution) {
+	if execution.TestWorkflowExecutionName != "" {
```
rangoo94, line 155: Maybe better to check only once if the execution exists in Kubernetes? It may be called unnecessarily too often

api/v1/testkube.yaml
rangoo94: or maybe something that is more related with CR, considering that it's not changing the execution name?

pkg/tcl/testworkflowstcl/testworkflowexecutor/executor.go
vsukhin: good point

docs/docs/articles/test-executions.md
vsukhin: will copy

api/v1/testkube.yaml
vsukhin: yes, the idea was to have it similar to TestWorkflowExecution.Name

PR #5571, opened 2024-06-12
fix(testworkflows): support +00:00 timezone in CLI TestWorkflows logs

rangoo94:
## Pull request description 

* Similar to FE, support `+00:00` in Log Output

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cmd/kubectl-testkube/commands/testworkflows/run.go
+++ b/cmd/kubectl-testkube/commands/testworkflows/run.go
@@ -217,7 +217,13 @@ func watchTestWorkflowLogs(id string, signature []testkube.TestWorkflowSignature
 		// Strip timestamp + space for all new lines in the log
 		for len(l.Log) > 0 {
 			if isLineBeginning {
-				l.Log = l.Log[LogTimestampLength+1:]
+				if len(l.Log) >= 29 && l.Log[29] == '+' {
+					// Custom timezone (+00:00)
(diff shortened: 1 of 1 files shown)


cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -217,7 +217,13 @@ func watchTestWorkflowLogs(id string, signature []testkube.TestWorkflowSignature
 		// Strip timestamp + space for all new lines in the log
 		for len(l.Log) > 0 {
 			if isLineBeginning {
-				l.Log = l.Log[LogTimestampLength+1:]
+				if len(l.Log) >= 29 && l.Log[29] == '+' {
```
vsukhin, line 220: should be const

rangoo94: I'm not sure - probably it may be less clean with that. I would think of something like `LogTimeFormatDiscriminatorCharIndex = 29` and `LogTimeFormatDiscriminatorTimezone = '+'`, but while it's descriptive, it's quite obfuscating. Do you have some other proposal?

vsukhin: At least it will tell you what is 29 about. Or add the comment instead. It's a matter of taste, ofc

PR #5591, opened 2024-06-18
feat: tests - workflows updated after switching to official v1 templates

tkonieczny:


test/gradle/executor-smoke/crd-workflow/smoke.yaml
```diff
@@ -61,3 +61,32 @@ spec:
...
+      - contrib/executor/gradle/examples/hello-gradle
+  container:
+    resources:
+      requests:
+        cpu: 512m
+        memory: 512Mi
+    workingDir: /data/repo/contrib/executor/gradle/examples/hello-gradle
+    env:
+      - name: TESTKUBE_GRADLE
+        value: "true"
```
rangoo94, line 86: Guess that it's more common in our codebase to avoid indent for lists:
```suggestion
    - name: TESTKUBE_GRADLE
      value: "true"
```

rangoo94 (approved):

PR #5606, opened 2024-06-21
feat(TKC-2140): add basic secret mgmt / support auto-creating secrets for plain-text data

rangoo94:
### Secrets Management

* Add more granular permissions for listing secrets
   * Support listing Testkube-owned secrets only
   * For now they are only internally, based on the same Helm Charts / env variables as before
* Allow to define prefix for creating secrets
   * Thanks to that, we won't disclosure any external secrets by "already exists" error
* Expose more information in secrets list
   * Earlier: name and keys
   * Now: also owner (based on `ownerReferences`), labels, creation and update time, is it controlled by Testkube
* Add endpoints to modify/delete/create secrets owned by Testkube

### Auto-creation of secrets

* ~~When provided `?autoCreateSecrets=true` query parameter, creation and update of TestWorkflow/TestWorkflowTemplate will create the secrets instead of plain text Git credentials~~ It will create secrets for marked properties, see: [**new mechanism**](https://github.com/kubeshop/testkube/pull/5606#issuecomment-2188232496)
   * The secrets will be automatically created with such resource ownership = will be deleted after i.e. TestWorkflow will be deleted
   * It may be disabled same as now, via Helm chart `disableSecretCreation` value

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Fixes

- https://linear.app/kubeshop/issue/TKC-2140/%5Bworkflows%5D-secret-management


exu (approved):

rangoo94:
I had to change the implementation. Earlier, it was `autoCreateSecrets` parameter, that was creating secrets for every plain-text Git credential.

Now, you can select the values that should be external, using `EnvVarSource` with `name: "<computed>"` and `key: "the-secret-value"`. This way it's possible to:
* select a subset of credentials to create (i.e. only token, not username)
* create secrets for `content.files` (with `contentFrom`)
* create secrets for environment variables (with `valueFrom`)

While it requires now to specifically tell which credentials should be stored outside, it's much more flexible, what is required by the UI.

Example:
```yaml
spec:
  content:
    git:
      uri: https://github.com/kubeshop/testkube-dashboard
      username: plain-text-username
      tokenFrom:
        secretKeyRef:
          name: "<computed>"
          key: "very-secret-token"
```

PR #5609, opened 2024-06-25
fix: namespaced agent events with agent prefix

exu:
## Pull request description 

Control Plane and Agent are on same `events` prefix - some edge cases (like recent invalid NATS selector settings caused strange events ping pongs. to allow even use single NATS for On prem setup I've changed prefix to `agent` in agent 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/api/v1/testkube/model_event_extended.go
+++ b/pkg/api/v1/testkube/model_event_extended.go
@@ -8,8 +8,9 @@ import (
 )
 
 const (
-	TestStartSubject = "events.test.start"
-	TestStopSubject  = "events.test.stop"
+	Subscription     = "agent"
+	TestStartSubject = Subscription + ".test.start"
+	TestStopSubject  = Subscription + ".test.stop"
 )
 
 // check if Event implements model generic event type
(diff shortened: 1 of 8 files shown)


vsukhin (approved):

mgazza (commented):
A couple more suggestions beyond what Greptile flagged.

pkg/api/v1/testkube/model_event_extended.go
```diff
@@ -8,8 +8,9 @@ import (
 )
 
 const (
-	TestStartSubject = "events.test.start"
-	TestStopSubject  = "events.test.stop"
+	Subscription     = "agent"
+	TestStartSubject = Subscription + ".test.start"
```
mgazza, line 12: The `"agent"` prefix is now defined in three places: here, `pkg/logs/client/interface.go`, and `pkg/event/bus/nats.go` (`SubscriptionName`). If any of them drifts, topics silently stop matching.

Consider importing from a single source of truth. If circular imports are a concern, a small shared package (e.g. `pkg/event/constants`) would work.

pkg/api/v1/testkube/model_event_extended.go
```diff
@@ -262,14 +263,14 @@ func (e Event) Topic() string {
 	}
 
 	if e.Resource == nil {
-		return "events.all"
+		return Subscription + ".all"
 	}
 
 	if e.ResourceId == "" {
-		return "events." + string(*e.Resource)
+		return Subscription + "." + string(*e.Resource)
```
mgazza, line 270: Runtime string concatenation with `+` across multiple segments is harder to scan than `strings.Join`. The compile-time const expressions (like `Subscription + ".test.start"`) are fine, but these runtime ones would benefit:

```suggestion
	return strings.Join([]string{Subscription, string(*e.Resource), e.ResourceId}, ".")
```

courageousillumination:

PR #5614, opened 2024-06-27
fix: (TKC-2164) valid container executor exit code handling

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/executor/common.go
```diff
@@ -557,7 +557,7 @@ func GetPodErrorMessage(ctx context.Context, client kubernetes.Interface, pod *c
 
 	for _, containerStatus := range pod.Status.ContainerStatuses {
 		if containerStatus.State.Terminated != nil &&
-			(containerStatus.State.Terminated.ExitCode > 1 || containerStatus.State.Terminated.ExitCode < -1) &&
+			(containerStatus.State.Terminated.ExitCode >= 1 || containerStatus.State.Terminated.ExitCode < -1) &&
```
vsukhin, line 560: then also for init container above. but the idea was for 1 and -1 to show test error message from the log, because it's just failed test

exu: There is empty message in real time long case of `1`

vsukhin (approved):

pkg/executor/common.go
vsukhin: ok, it's fine for me, if @tkonieczny has no objections

vsukhin (approved):

pkg/executor/common.go
exu: ok added also in init container

PR #5631, opened 2024-07-03
fix: error handling when creating TW

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/testworkflows/create.go
+++ b/cmd/kubectl-testkube/commands/testworkflows/create.go
@@ -68,7 +68,9 @@ func NewCreateTestWorkflowCmd() *cobra.Command {
 			}
 
 			workflow, err := client.GetTestWorkflow(obj.Name)
-			ui.ExitOnError("getting test workflow "+obj.Name+" in namespace "+obj.Namespace, err)
+			if err != nil {
+				ui.Debug("getting test workflow: " + err.Error())
+			}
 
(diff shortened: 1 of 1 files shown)


vsukhin (approved):

cmd/kubectl-testkube/commands/testworkflows/create.go
```diff
@@ -67,7 +67,9 @@ func NewCreateTestWorkflowCmd() *cobra.Command {
 				}
 			}
 
-			workflow, _ := client.GetTestWorkflow(obj.Name)
+			workflow, err := client.GetTestWorkflow(obj.Name)
+			ui.ExitOnError("getting test workflow "+obj.Name+" in namespace "+obj.Namespace, err)
```
rangoo94, line 71: That should not be there, as will disallow creating TestWorkflow on 404

rangoo94 (changes requested):

rangoo94 (approved):
Not sure if debug message makes sense in case of lack of workflow, but it's not critical issue so approving 👍

exu:
> Not sure if debug message makes sense in case of lack of workflow during creation, but it's not critical issue so approving 👍


```
You'll get it only when passing --verbose flag
```

PR #5633, opened 2024-07-04
feat: running context model for test workflows

vsukhin:

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


api/v1/testkube.yaml
```diff
@@ -7874,6 +7875,9 @@ components:
           type: boolean
           description: whether webhooks on the executions of this test workflow are disabled
           default: false
+        runningContext:
```
rangoo94, line 7878: I would suggest designing a new replacement for the existing Running Context, as it's very limiting - having only type and context (= i.e. execution ID) string is limiting a lot, as we can't i.e. build the proper tree.

vsukhin: ok, do we have a list of required fields?

rangoo94: I'm not aware of any solid research conducted on this topic, so we rather don't have anything like that yet

vsukhin: hey @jmorante-ks need some imput about what is required to store besides what we do now

exu: I'm agree that it would be better do design this first from UX PoV. 

@fivenp @TheBrunoLopes any thoughts? 

Currently our "context" is more about OSS. I think for now we have only one customer requirements about this: 

We know at least what Adobe is missing here: 
* they want to know who run the test - what means that it'll be the user or some API token, but we have also events taken from triggers.

api/v1/testkube.yaml
```diff
@@ -7038,10 +7038,71 @@ components:
...
+    TestWorkflowRunningContext:
+      description: running context for test workflow execution
+      type: object
+      required:
+        - interface
+        - actor
+      properties:
+        interface:
+          $ref: "#/components/schemas/TestWorkflowRunningContextInterface"
+        actor:
```
rangoo94, line 7055: Actor should rather have detailed information about actor.

api/v1/testkube.yaml
```diff
@@ -7038,10 +7038,71 @@ components:
...
+          $ref: "#/components/schemas/TestWorkflowRunningContextActor"
+        caller:
```
rangoo94, line 7057: I don't understand the caller idea here

rangoo94 (commented):
Generally, I would suggest starting from defining the solution architecture, to define what we need. It doesn't look good, and it will be basically impossible to fix it in future, whatever we will do - as backwards compatibility will be pretty hard.

vsukhin:
ok, stopping here till final requirements from @jmorante-ks and @TheBrunoLopes

PR #5660, opened 2024-07-12
feat: TKC:2213: add support for defining image credentials cache TTL

dejanzele:
## Pull request description 

Testkube caches the image pull secrets indefinitely and this can cause issues if the image registry auth gets rotated.
This PR adds support for defining a TTL for the image pull secrets.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

- Issue where customers reported stale image pull credentials being used in Testkube while inspecting image via Skopeo

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1,0 +1,33 @@
+package cache
+
+import (
+	"context"
+	"math"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	ErrNotFound = errors.New("item not found")
+)
+
+type Cache[T any] interface {
+	// Get retrieves the cached value for the given key.
+	// If the key is not found or expired, the method should return ErrNotFound.
+	Get(ctx context.Context, key string) (T, error)
+	// Set stores the value in the cache with the given key.
+	// If ttl is 0, the item should not be cached and this method should return no error.
+	Set(ctx context.Context, key string, value T, ttl time.Duration) error
+}
+
+// IsCacheMiss returns true if the error is a cache miss error.
+// This is a helper function to determine so users don't have to compare errors manually.
+func IsCacheMiss(err error) bool {
+	return errors.Is(err, ErrNotFound)
+}
+
+// InfiniteTTL returns a time.Duration that represents an infinite TTL.
+func InfiniteTTL() time.Duration {
+	return math.MaxInt64
+}
--- a/pkg/cache/inmem.go
+++ b/pkg/cache/inmem.go
@@ -1,0 +1,71 @@
+package cache
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+type item[T any] struct {
+	value     T
+	expiresAt *time.Time
+}
+
+// timeGetter is a function that returns the current time.
+type timeGetter func() time.Time
+
+type InMemoryCache[T any] struct {
+	cache      sync.Map
+	timeGetter timeGetter
+}
+
+// NewInMemoryCache creates a new in-memory cache.
+// The underlying cache implementation uses a sync.Map so it is thread-safe.
+func NewInMemoryCache[T any]() *InMemoryCache[T] {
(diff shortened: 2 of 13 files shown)


pkg/cache/cache.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	ErrStaleCache = errors.New("item's ttl has expired")
```
rangoo94, line 11: Probably `ErrNotFound` is enough - from outside of cache, the item is then not found, as it's expired. `ErrStaleCache` is implementation-dependent - it would not happen if i.e. obsolete items would be cleared externally, like in worker, or by key-value store TTL.

pkg/cache/inmem.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+type item struct {
```
rangoo94, line 11: Could be generic, so the `value` + `Cache.Get` is typed too

rangoo94 (approved):
LGTM, possible minor improvements for the interface

pkg/cache/cache.go
dejanzele: My original idea was to be able to detect is item missing from cache or is it stale, and then you could do different actions, but to be fair, we don't have an use case for it, so it could be removed and to stick only with `ErrNotFound`

pkg/cache/inmem.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (c *InMemoryCache[T]) Set(ctx context.Context, key string, value T, ttl time.Duration) error {
+	if ttl < 0 {
+		return errors.New("ttl must be greater than 0")
+	}
+
+	i := &item[T]{
+		value: value,
+	}
+	if ttl > 0 {
```
rangoo94, line 59: Optionally, dunno: Should it be indefinite for `0` or rather ignored (as cached for 0 seconds)? Could be alternatively `0` for ignore and `< 0` for indefinite

pkg/imageinspector/secretfetcher.go
```diff
@@ -2,49 +2,83 @@ package imageinspector
...
+	return s
 }
 
 func (s *secretFetcher) Get(ctx context.Context, name string) (*corev1.Secret, error) {
-	// Get cached secret
-	s.mu.RLock()
-	if v, ok := s.cache[name]; ok {
-		s.mu.RUnlock()
-		return v, nil
+	if s.ttl > 0 {
```
rangoo94, line 43: Looks like a clash with TTL from cache - zero here won't cache

rangoo94 (approved):

pkg/cache/inmem.go
dejanzele:
```
// InfiniteTTL returns a time.Duration that represents an infinite TTL.
func InfiniteTTL() time.Duration {
	return math.MaxInt64
}
```

This sets the expiration date somewhere around year 2316

pkg/imageinspector/secretfetcher.go
dejanzele: I thought about adding a new ennvvar, but it seemed simpler to just use 0 for no cache, and you could set the cache duration to a really high duration for "infinite".

I don't have a strong opinion on this.

rangoo94: In this comment, I'm only showing that there is clash between `InMemoryCache` and `SecretFetcher` when `ttl: 0`:
* `SecretFetcher` - when `ttl: 0` it won't use cache (as `(0 > 0) == false`)
* `InMemoryCache` - when `ttl: 0` it will cache indefinitely

dejanzele: oh, thanks, good catch

PR #5720, opened 2024-08-02
fix: [TKC-2168] id sequence (#5681)

vsukhin:
* fix: change sequence for numbers

* fix: add line

* fix: add sequence repository

* fix: add sequence repository

* fix: add sequence for test suites

* fix: rename option method

* fix: add sequence for test workflows

* fix: skip crd deletion

* fixL use insert instead of update

* fix: integration test sync

* fix: intergrattion test

* fix: parallel integration test

* fix: use default value

* fix: typo nil check

* fix: change prefix

* fix: test suite prefix

* fix: new cloud methods

* fix: dep update

---------

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


exu (approved):

pkg/repository/result/mongo.go
```diff
@@ -695,9 +695,11 @@ func (r *MongoRepository) DeleteByTest(ctx context.Context, testName string) (er
 	if err != nil {
 		return
 	}
-	err = r.DeleteExecutionNumber(ctx, testName)
-	if err != nil {
-		return
+	if r.sequenceRepository != nil {
```
rangoo94, line 698: Shouldn't it panic when there is no sequenceRepository? Same for other places.

rangoo94, line 698: Shouldn't it panic when there is no sequenceRepository? Same for other places.

pkg/repository/sequence/mongo.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+type executionNumber struct {
+	Id            string        `bson:"_id"`
+	Number        int           `bson:"number"`
+	ExecutionType ExecutionType `bson:"executionType"`
+}
+
+// GetNextExecutionNumber gets next execution number by name and type
+func (r *MongoRepository) GetNextExecutionNumber(ctx context.Context, name string, executionType ExecutionType) (number int32, err error) {
```
rangoo94, line 63: This have a downside, that the Test / Test Workflow etc won't necessarily be created after `GetNextExecutionNumber`, so consecutive executions won't have consecutive sequence numbers.

From a technical perspective it's rather a better situation than before, but from the user perspective - the gaps between executions will lead to questions.

Rather creation + sequence increment should be atomic.

pkg/repository/testworkflow/mongo.go
```diff
@@ -360,12 +369,26 @@ func composeQueryAndOpts(filter Filter) (bson.M, *options.FindOptions) {
 
 // DeleteByTestWorkflow deletes execution results by workflow
 func (r *MongoRepository) DeleteByTestWorkflow(ctx context.Context, workflowName string) (err error) {
+	if r.sequenceRepository != nil {
```
rangoo94, line 372: Should this be part of TestWorkflow's repository? Same for other similar

pkg/repository/sequence/mongo.go
vsukhin: consecutive is a bit weird in parallel world. if we run 100 executions in parallel, the number doesn't mean this much. still we can  create a transaction on inserting executtion and getting the number to avoid gaps

vsukhin: transactions in mongo are supported only for  replicaset, we can't guarantee

PR #5756, opened 2024-08-13
fix: controlling issues

rangoo94:

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/testworkflows/testworkflowcontroller/controller.go
+++ b/pkg/testworkflows/testworkflowcontroller/controller.go
@@ -231,6 +231,7 @@ func (c *controller) WatchLightweight(parentCtx context.Context) <-chan Lightwei
 	prevNodeName := ""
 	prevPodIP := ""
 	prevStatus := testkube.QUEUED_TestWorkflowStatus
+	prevIsFinished := false
 	sig := stage.MapSignatureListToInternal(c.signature)
 	ch := make(chan LightweightNotification)
 	go func() {
@@ -245,6 +246,7 @@ func (c *controller) WatchLightweight(parentCtx context.Context) <-chan Lightwei
 			podIP, _ := c.PodIP(parentCtx)
 			current := prevCurrent
 			status := prevStatus
+			isFinished := prevIsFinished
 			if v.Value.Result != nil {
 				if v.Value.Result.Status != nil {
 					status = *v.Value.Result.Status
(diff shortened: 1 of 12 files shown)


pkg/testworkflows/testworkflowcontroller/controller.go
```diff
@@ -245,20 +246,25 @@ func (c *controller) WatchLightweight(parentCtx context.Context) <-chan Lightwei
...
 					status = testkube.QUEUED_TestWorkflowStatus
 				}
 				current = v.Value.Result.Current(sig)
+				isFinished = v.Value.Result.IsFinished()
 			}
 
-			if nodeName != prevNodeName || podIP != prevPodIP || prevStatus != status || prevCurrent != current {
+			// TODO: the final status should always have the finishedAt too,
+			//       there should be no need for checking isFinished diff
+			if nodeName != prevNodeName || isFinished != prevIsFinished || podIP != prevPodIP || prevStatus != status || prevCurrent != current {
```
vsukhin, line 262: such code will not be simple, to maintain (

pkg/testworkflows/testworkflowcontroller/watchinstrumentedpod.go
```diff
@@ -65,13 +66,27 @@ func WatchInstrumentedPod(parentCtx context.Context, clientSet kubernetes.Interf
...
-		podObj := <-pod.Peek(ctx)
+		var podObj *corev1.Pod
+		select {
+		// Obtain the Pod information for further execution
+		case p := <-pod.Peek(ctx):
+			podObj = p
+		// Handle when the execution have been finished, but there may be no Pod
+		case <-state.Finished(""):
+			select {
+			case <-time.After(1 * time.Second):
```
vsukhin, line 83: would be nice to use const

pkg/testworkflows/testworkflowcontroller/podstate.go
```diff
@@ -381,58 +411,97 @@ func (p *podState) ContainerResult(name string) (ContainerResult, error) {
...
+				// Try to firstly finish with the Pod information when it's possible
+				if IsJobDone(v.Value) && state.FinishedAt("").IsZero() && HadPodScheduled(v.Value) {
+					select {
+					case p, ok := <-podCh:
+						if p.Error != nil {
+							errorHandler(p.Error)
+						} else if ok {
+							state.RegisterPod(p.Value)
+						}
+					case <-time.After(2 * time.Second):
```
vsukhin, line 498: some const value

vsukhin (approved):

pkg/testworkflows/testworkflowcontroller/controller.go
rangoo94: I fully agree, although this code is rather not meant to be maintained. This PR, along actual bugfixes, is meant to "enable" the new orchestration with similar watching system.

The (mainly) `WatchInstrumentedPod` and `TestWorkflowResult` have so many edge cases handled, clock calibration, and auto-healing mechanisms implemented, that at this point it's probably better to just rewrite them based on the observations that we have for the last few months of running Test Workflows. After that, we will likely not need conditions like these at all.

PR #5771, opened 2024-08-19
feat: add test triggers client for external use

dejanzele:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/api/v1/testkube/model_test_workflow_extended.go
```diff
@@ -80,3 +80,19 @@ func (w *TestWorkflow) GetObjectRef() *ObjectRef {
 
 func (w *TestWorkflow) QuoteWorkflowTextFields() {
 }
+
+func (w TestWorkflow) GetName() string {
```
WitoDelnat, line 84: For RBAC we need Workflow, Webhook, Trigger and WorkflowTemplate. The latter seems to be missing from this PR. If we'd add it then we have to avoid making another release just for that.

WitoDelnat (approved):

PR #5775, opened 2024-08-21
fix: http/2 stream leaks

ed382:
## Pull request description 

When response bodies are not properly closed an HTTP/2 client can leak streams which can eventually lead to a failure case (i.e. `stream error: stream ID 1; INTERNAL_ERROR`) related HTTP/2 flow-control. Here is a more thorough description of the issue: https://github.com/googleapis/google-cloud-go/issues/7440#issuecomment-1491008639

This fixes cases where the response body for an HTTP request was not closed, and also refactors some of the code to make it easier to see when the stream has been closed. 

In a couple of cases, it relied on the `SendStream` method from the `fiber` package to close the stream, but I could not find anywhere in their docs or code base where it actually closed the stream.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/testworkflow-toolkit/transfer/server.go
+++ b/cmd/testworkflow-toolkit/transfer/server.go
@@ -121,6 +121,7 @@ func (t *server) handler() http.Handler {
 		}
 
 		err := common.UnpackTarball(dirPath, request.Body)
+		defer request.Body.Close()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


exu (approved):

vsukhin (approved):
lgtm. not sure, it will help in this case, but we can try

PR #5821, opened 2024-09-05
fix: sensitive data in events

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/event/kind/webhook/listener_test.go
+++ b/pkg/event/kind/webhook/listener_test.go
@@ -34,7 +34,7 @@ func TestWebhookListener_Notify(t *testing.T) {
 		svr := httptest.NewServer(testHandler)
 		defer svr.Close()
 
-		l := NewWebhookListener("l1", svr.URL, "", testEventTypes, "", "", nil, false, nil, nil, nil, v1.NewMetrics(), nil, map[string]string{})
+		l := NewWebhookListener("l1", svr.URL, "", testEventTypes, "", "", nil, false, nil, nil, nil, v1.NewMetrics(), nil, nil)
 
 		// when
 		r := l.Notify(testkube.Event{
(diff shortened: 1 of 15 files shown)


rangoo94 (commented):
* Emitter don't need `Envs` passed then to `NewEmitter`
* Will webhooks have access to `Envs` then?

vsukhin:
Might be, we need to leave onyl WEBHOOKS_ prefixed env

rangoo94:
> Might be, we need to leave onyl WEBHOOKS_ prefixed env

Will it work, if the User will have some secrets exposed in API Server and required for specific webhooks (like Slack's API token)? Or maybe we should just populate `Envs` in the place where webhook is actually sent, and avoid it there?

vsukhin:
That would be better

exu:
> * Emitter don't need `Envs` passed then to `NewEmitter`
> * Will webhooks have access to `Envs` then?

Do we want to expose all envs to some external webhook calls at all?

rangoo94:
> Do we want to expose all envs to some external webhook calls at all?

Probably better would be to have them whitelisted by some prefix (as @vsukhin mentioned), but at the moment Users are already using it with no patterns, so it would be breaking change to add prefix 😕

pkg/event/kind/webhook/listener_test.go
```diff
@@ -34,7 +34,7 @@ func TestWebhookListener_Notify(t *testing.T) {
 		svr := httptest.NewServer(testHandler)
 		defer svr.Close()
 
-		l := NewWebhookListener("l1", svr.URL, "", testEventTypes, "", "", nil, false, nil, nil, nil, v1.NewMetrics(), nil)
+		l := NewWebhookListener("l1", svr.URL, "", testEventTypes, "", "", nil, false, nil, nil, nil, v1.NewMetrics(), nil, map[string]string{})
```
rangoo94, line 37: `nil` would be fine too (here and everywhere below)

vsukhin (approved):

pkg/event/kind/webhook/listener_test.go
exu: done

exu:
> LGTM, but it would be great to test it 👍

Will check it on sandbox

exu:
image

webhooks are there on sandbox

PR #5844, opened 2024-09-11
feat: Multi agent

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/api-server/main.go
```diff
@@ -431,13 +438,24 @@ func main() {
...
+	proContext.ClusterId = clusterId
 
 	// Check Pro/Enterprise subscription
 	var subscriptionChecker checktcl.SubscriptionChecker
 	if mode == common.ModeAgent {
 		subscriptionChecker, err = checktcl.NewSubscriptionChecker(ctx, proContext, grpcClient, grpcConn)
 		exitOnError("Failed creating subscription checker", err)
+
+		// Load environment/org details based on token grpc call
+		environment, err := controlplanetcl.GetEnvironment(ctx, proContext, grpcClient, grpcConn)
```
exu, line 451: executor need to follow loading it from this one - not from inlined env variables

pkg/agent/agent.go
```diff
@@ -378,10 +391,31 @@ func (ag *Agent) runWorkers(ctx context.Context, numWorkers int) error {
...
+	}
+
+	ag.logger.Errorf("command not found", "availableHandlers", names, "command", c.Command)
+	return &cloud.ExecuteResponse{MessageId: c.MessageId, Status: 404, Body: []byte("command not found, you've passed unhandled GRPC request command, check agent / control plane versions for API consistency, available handlers: " + names)}
+}
+
 func (ag *Agent) executeCommand(ctx context.Context, cmd *cloud.ExecuteRequest) *cloud.ExecuteResponse {
 	switch {
-	case cmd.Url == healthcheckCommand:
+	case cmd.Url == HealthcheckCommand || cmd.Command == string(cloud.HealthcheckCommand):
```
exu, line 413: This is not needed anymore

pkg/agent/handlers/response.go
```diff
@@ -0,0 +1,47 @@
+package handlers
```
exu, line 1: would it be better to have some errors wrapped
e.g. 
```
errors.Wrap(err, errors.NotFound)

```

and next the mapper which will check what kinds of error this is and set valid response

pkg/testworkflows/testworkflowexecutor/executor.go
```diff
@@ -542,11 +532,55 @@ func (e *executor) Execute(ctx context.Context, workflow testworkflowsv1.TestWor
 		log.DefaultLogger.Errorw("failed to encode tags", "id", id, "error", err)
 	}
 
+	// Get (for centralized mode) TW execution or create it
+	if request.Id != "" {
```
exu, line 536: We need to de-concrete executions logic from here to split fully executor engine from data implementation

pkg/testworkflows/testworkflowexecutor/executor.go
```diff
@@ -558,34 +592,38 @@ func (e *executor) Execute(ctx context.Context, workflow testworkflowsv1.TestWor
...
+		Steps: stage.MapSignatureListToStepResults(bundle.Signature),
+	}
+
+	execution.Output = []testkube.TestWorkflowOutput{}
+	execution.Workflow = testworkflowmappers.MapKubeToAPI(initialWorkflow)
+	execution.ResolvedWorkflow = testworkflowmappers.MapKubeToAPI(resolvedWorkflow)
+	execution.Tags = tags
+
+	// Insert or save execution
+	if request.Id != "" {
```
exu, line 611: exctract it outside? 

executor could be in form
```
func(ExecutionRequest, Workflow) Result
```

pkg/agent/handlers/execute.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/kubeshop/testkube/pkg/agent"
+	"github.com/kubeshop/testkube/pkg/api/v1/testkube"
+	"github.com/kubeshop/testkube/pkg/cloud"
+	"github.com/kubeshop/testkube/pkg/testworkflows/testworkflowexecutor"
+	"github.com/kubeshop/testkube/pkg/utils/codec"
+)
+
+// TODO - valid error handling
+
+func NewExecuteTestWorkflowHandler(
```
rangoo94, line 19: This is for sure not needed:
* There is already option to execute Test Workflow via generic gRPC command we have (that calls API)
* For future Execution Worker, the command will look differently - as the Execution Worker needs to be stateless

rangoo94:
Considering that runner IDs are not dynamic (but needs to be pre-created), either:
* There should be no Runner ID thing, but rather separate Agent Keys for each of them
* The Runner ID should be completely informal

pkg/agent/handlers/execute.go
exu: I've planned here to fully decouple from API server - as it's additional unneded thing.

exu:
> Considering that in this PR the runner IDs are not dynamic (but needs to be pre-created), either:
> 
> * There should be no Runner ID thing, but rather separate Agent Keys for each of them
> * The Runner ID should be completely dynamic (and informal + not unique)
> 
> On the other hand, the **best** solution is to avoid runner IDs at all, and have runner tags instead (like K8S `nodeAffinity` -> Testkube `runnerAffinity`). It would work great:
> 
> * simple,
> * Kubernetes-like,
> * fully customizable,
> * scalable (as runners could be created on demand too)
> * `runnerAffinity` could be defined on any level (globally, per workflow, per execution)

But you need some kind of ID here (whatever we name it)  to e.g. schedule against it - I'm not sure if we really should follow Kubernetes naming here at all. These points are both valid for IDs and tags like in affinity. 


I agree about separate keys - if we want to decouple runners from environments - it'll be next thing to do for sure.

pkg/agent/handlers/execute.go
exu: These was the attempt to decouple it from the API server totally - can be refactored/reordered if we decide about API later

rangoo94:
> But you need some kind of ID here (whatever we name it) to e.g. schedule against it

* You only need to target a tag, not ID - if someone wants to have actually unique way to identify them, simply there may be some tag unique 🙂 
* IDs *are* required, but not in the way that User needs to pass them. They may only be useful to pull the already started execution information, or control it (pause, resume, abort), That ID should be provided on the Agent only though (and associated to i.e. execution)

pkg/agent/handlers/execute.go
rangoo94: > I've planned here to fully decouple from API server - as it's additional unneded thing. So it's quite needed to not spawn this APIServer

But when it will be decoupled from the API Server, the signature will also differ, so (A) this function is for decoupling, yet (B) it needs to be deleted after decoupling, as it will look differently 🙂

PR #5890, opened 2024-10-02
fix: clean completed migration job before reinitialize on-prem demo

exu:
## Pull request description 

The idea is to clean long TTL migration jobs when they exist and we're trying to recreate them. 

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/common/helper.go
+++ b/cmd/kubectl-testkube/commands/common/helper.go
@@ -162,6 +162,30 @@ func updateHelmRepo(helmPath string, dryRun bool, isOnPrem bool) *CLIError {
 	return nil
 }
 
+// It cleans existing migrations job with long TTL
+func CleanExistingCompletedMigrationJobs(namespace string) (cliErr *CLIError) {
+	kubectlPath, cliErr := lookupKubectlPath()
+	if cliErr != nil {
+		return cliErr
+	}
+
+	// Clean the job only when it's found and it's state is successful - ignore pending migrations.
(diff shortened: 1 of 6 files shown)


cmd/kubectl-testkube/commands/common/helper.go
```diff
@@ -162,6 +162,30 @@ func updateHelmRepo(helmPath string, dryRun bool, isOnPrem bool) *CLIError {
...
 
+// It cleans existing migrations job with long TTL
+func CleanExistingCompletedMigrationJobs(namespace string) (cliErr *CLIError) {
+	kubectlPath, cliErr := lookupKubectlPath()
+	if cliErr != nil {
+		return cliErr
+	}
+
+	// Clean the job only when it's found and it's state is successful - ignore pending migrations.
+	succeeded, _ := runKubectlCommand(kubectlPath, []string{"get", "job", "testkube-enterprise-api-migrations", "-n", namespace, "-o", "jsonpath={.status.succeeded}"})
```
vsukhin, line 173: is it okay to skip error or it will give us additional info?

vsukhin (approved):
one question

exu:
> one question

TBH I'm not sure here if I want to inform user about false positives here

cmd/kubectl-testkube/commands/common/helper.go
exu: I think I would need to filter errors related to state when there is no migration job at all 
Assuming happy path for now

PR #5957, opened 2024-10-22
feat: extract Control Plane compatible with OSS version

rangoo94:
## Pull request description 

* Extract Control Plane compatible with OSS version
   * The system will either connect to Cloud or local Control Plane - won't interact with database or Object Storage different way
   * **Exception:** Legacy Tests and Test Suites may interact directly with the Object Storage to store/read artifacts - to avoid breaking changes for not-compatible features (`ArtifactRequest.StorageBucket` and `ArtifactRequest.OmitFolderPerExecution`)
* Isolated a bit more deprecated resources
  * We will be able to make feature flag / environment variable to gracefully disable deprecated features

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


pkg/controlplane/server.go
```diff
@@ -0,0 +1,263 @@
...
+	return g.Wait()
+}
+
+// TODO: Consider deleting that
+func (s *Server) GetTestWorkflowNotificationsStream(srv cloud.TestKubeCloudAPI_GetTestWorkflowNotificationsStreamServer) error {
+	ctx, cancel := context.WithCancel(srv.Context())
+	defer cancel()
+	g, _ := errgroup.WithContext(ctx)
+
+	ticker := time.NewTicker(30 * time.Second)
```
vsukhin, line 108: looks like const

pkg/controlplane/server.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+
+// TODO: Consider deleting that
+func (s *Server) GetLogsStream(srv cloud.TestKubeCloudAPI_GetLogsStreamServer) error {
+	ctx, cancel := context.WithCancel(srv.Context())
+	defer cancel()
+	g, _ := errgroup.WithContext(ctx)
+
+	ticker := time.NewTicker(30 * time.Second)
```
vsukhin, line 164: looks like const

vsukhin (approved):

PR #5958, opened 2024-10-22
fix: handle container optimization for default images when pure shell script is run at last

rangoo94:
## Pull request description 

* Handle container optimization (merging containers) for default images properly

## Example

```yaml
spec:
  system:
    pureByDefault: true
  steps:
  - run:
      image: curlimages/curl:latest
  - shell: echo hello
```

Fails with `/.tktw-bin/sh` not found

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/pkg/testworkflows/testworkflowprocessor/action/group.go
+++ b/pkg/testworkflows/testworkflowprocessor/action/group.go
@@ -169,5 +169,15 @@ merging:
 		i++
 	}
 
+	// Fix the auto-merged internal images
+	for i := range groups {
+		image := groups[i].Image()
+		// Re-use /.tktw/bin/sh when the internal image has been merged into different container
+		if image != constants.DefaultToolkitImage && image != constants.DefaultInitImage {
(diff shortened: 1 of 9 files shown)


pkg/testworkflows/testworkflowprocessor/action/group.go
```diff
@@ -169,5 +169,15 @@ merging:
 		i++
 	}
 
+	// Fix the auto-merged internal images
+	for i := range groups {
+		image := groups[i].Image()
+		// Re-use /.tktw/bin/sh when the internal image has been merged into different container
+		if image != constants.DefaultToolkitImage && image != constants.DefaultInitImage {
+			groups[i] = groups[i].RewireCommandDirectory(constants.DefaultInitImage, "/.tktw-bin", constants.InternalBinPath)
```
vsukhin, line 177: should we use const for "/.tktw-bin" ?

vsukhin (approved):

PR #5969, opened 2024-10-24
fix: junit workflow suite fixed

tkonieczny:
Sandbox test after `recurse: true`, junit workflow suite fixed


tkonieczny:
All workflows are created correctly with only sandbox trigger (suite-small)
```
$ kubectl get testworkflows -nenterprise-sandbox-sandbox-test-fixes|grep cron
postman-workflow-smoke-cron-trigger-test              3m20s
tw-suite-small-sandbox-smoke-cron-trigger             3m19s
```

PR #5976, opened 2024-10-28
feat: improve agent startup time

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/gofiber/fiber/v2/middleware/cors"
-	"google.golang.org/grpc"
 
 	executorsclientv1 "github.com/kubeshop/testkube-operator/pkg/client/executors/v1"
 	testkubeclientset "github.com/kubeshop/testkube-operator/pkg/clientset/versioned"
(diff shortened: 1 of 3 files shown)


cmd/api-server/main.go
```diff
@@ -126,16 +124,32 @@ func main() {
 
 	inspector := commons.CreateImageInspector(cfg, configMapClient, secretClient)
 
+	// Connect to NATS
```
WitoDelnat, line 127: Moving NATS first staggers the gRPC connection a bit when we start agent and control plane together. gRPC backoff is quite steep, but I want to avoid tweaking it until absolutely needed as these defaults are there for a  reason.

cmd/api-server/main.go
```diff
@@ -126,16 +124,32 @@ func main() {
...
 		controlPlaneUrl = fmt.Sprintf("127.0.0.1:%d", cfg.GRPCServerPort)
 	}
+
+	log.DefaultLogger.Infow("Connecting to control plane", "server", controlPlaneUrl, "insecure", cfg.TestkubeProTLSInsecure, "skipVerify", cfg.TestkubeProSkipVerify, "certFile", cfg.TestkubeProCertFile, "keyFile", cfg.TestkubeProKeyFile, "caFile", cfg.TestkubeProCAFile)
+	grpcConn, err := agent.NewClient(controlPlaneUrl, cfg.TestkubeProTLSInsecure, cfg.TestkubeProSkipVerify, cfg.TestkubeProCertFile, cfg.TestkubeProKeyFile, cfg.TestkubeProCAFile)
+	grpcClient := cloud.NewTestKubeCloudAPIClient(grpcConn)
+	defer grpcConn.Close()
+
+	// First request will 'wait for ready' with a timeout of agent.InitialTimeout.
+	proContext, err := commons.ReadProContext(ctx, cfg, grpcClient)
```
WitoDelnat, line 150: Moved this up a bit, as it groups nicely with the grpcConnection part, this is where we actually wait for the connection to move from idle -> transient error -> ready.

pkg/agent/agent.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	timeout                = 10 * time.Second
+	InitialTimeout         = 30 * time.Second
```
WitoDelnat, line 34: With gRPC backoff this would amount to: 1 + 1.6 + 2.5 + 4 +6.5 + 10.5 -> CRASH after 30s, last attempt at 26.1s.

PR #6016, opened 2024-11-08
feat: [TKC-2765] run test workflows by selector

vsukhin:

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -45,12 +46,13 @@ func NewRunTestWorkflowCmd() *cobra.Command {
 		format                   string
 		masks                    []string
 		tags                     map[string]string
+		selectors                []string
```
rangoo94, line 49: Are these selectors or a single `AND` selector? In example, does `[]string{"a=b", "c=d"}` mean items that match both or either?

vsukhin: Both, it prepares standard k8s selector, that we use to select test workflows, user can submit either --label a=b,c=d or --label a=b --label c=d

For either user can submit --label "a in (b,c)"

rangoo94: By reading the comment - it means that it's `OR`, right? In example `--label="a in (b, c)" --label="a=d" --label="c=d"` will run workflows:
* have `a` equal to `b`, `c`, or `d`, or
* have `c` equal to `d`

cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -45,12 +46,13 @@ func NewRunTestWorkflowCmd() *cobra.Command {
...
+		concurrencyLevel         int
```
rangoo94, line 50: The `--concurrency` is not supported, so should be rather deleted

pkg/api/v1/client/interface.go
```diff
@@ -151,6 +151,7 @@ type TestWorkflowAPI interface {
 	UpdateTestWorkflow(workflow testkube.TestWorkflow) (testkube.TestWorkflow, error)
 	DeleteTestWorkflow(name string) error
 	ExecuteTestWorkflow(name string, request testkube.TestWorkflowExecutionRequest) (testkube.TestWorkflowExecution, error)
+	ExecuteTestWorkflows(sselector string, concurrencyLevel int, request testkube.TestWorkflowExecutionRequest) ([]testkube.TestWorkflowExecution, error)
```
rangoo94, line 154:
```suggestion
	ExecuteTestWorkflows(selector string, concurrencyLevel int, request testkube.TestWorkflowExecutionRequest) ([]testkube.TestWorkflowExecution, error)
```

api/v1/testkube.yaml
```diff
@@ -8990,6 +9041,9 @@ components:
               $ref: "#/components/schemas/TestWorkflowTarballRequest"
           config:
             $ref: "#/components/schemas/TestWorkflowConfigValue"
+          selector:
```
rangoo94, line 9044: Shouldn't we just use our `$ref: "#/components/schemas/LabelSelector"`?

internal/app/api/v1/testworkflows.go
```diff
@@ -371,34 +388,37 @@ func (s *TestkubeAPI) ExecuteTestWorkflowHandler() fiber.Handler {
...
-		go workerpoolService.SendRequests(requests)
-		go workerpoolService.Run(ctx)
-
-		for r := range workerpoolService.GetResponses() {
-			results = append(results, r.Result)
-			if r.Err != nil {
-				errs = append(errs, r.Err)
+		if len(testWorkflows) != 0 {
+			request.TestWorkflowExecutionName = strings.Clone(c.Query("testWorkflowExecutionName"))
+			concurrencyLevel, err := strconv.Atoi(c.Query("concurrency", strconv.Itoa(scheduler.DefaultConcurrencyLevel)))
```
rangoo94, line 393: Should we support concurrency there? It will be only for scheduling, not actually executing concurrently

pkg/api/v1/client/testworkflow.go
```diff
@@ -121,6 +122,24 @@ func (c TestWorkflowClient) ExecuteTestWorkflow(name string, request testkube.Te
...
+	uri := c.testWorkflowExecutionTransport.GetURI("/test-workflow-executions")
+
+	body, err := json.Marshal(request)
+	if err != nil {
+		return executions, err
+	}
+
+	params := map[string]string{
+		"selector":    selector,
+		"concurrency": strconv.Itoa(concurrencyLevel),
```
rangoo94, line 137: As above - I don't think that we should have concurrency at this level

cmd/kubectl-testkube/commands/testworkflows/run.go
vsukhin: ```--label="a=b" --label="c=d"``` means AND  -  a equall b and c equal d. it will  be stringified as selector ```a=b,c=d``` 
```--label="a in (b, c)"``` means OR - a equal b or a equal c
 ```--label="a in (b, c)" --label="a=b" --label="c=d"```  means (a equal b or a equal c) AND (a equal b) AND (c equal d)

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
```The API currently supports two types of selectors: equality-based and set-based. A label selector can be made of multiple requirements which are comma-separated. In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.```

For expresssions
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement

internal/app/api/v1/testworkflows.go
vsukhin: Ok, we can remove it

pkg/api/v1/client/testworkflow.go
vsukhin: ok

PR #6055, opened 2024-11-26
fix: enabled video recording and webkit browser

olensmar:
image


vsukhin (approved):

olensmar:
@tkonieczny are you ok with this? perhaps there was a reason it wasn't initially enabled?

olensmar:
closing - not doing this as it would require too many resources..

PR #6058, opened 2024-11-28
feat: detect namespace of enterprise installation

exu:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/common/helper.go
+++ b/cmd/kubectl-testkube/commands/common/helper.go
@@ -813,6 +813,58 @@ func KubectlDescribeIngresses(namespace string) error {
 	return process.ExecuteAndStreamOutput(kubectl, args...)
 }
 
+func KubectlGetNamespacesHavingSecrets(selector string) ([]string, error) {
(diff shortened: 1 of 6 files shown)


cmd/kubectl-testkube/commands/common/helper.go
```diff
@@ -813,6 +813,58 @@ func KubectlDescribeIngresses(namespace string) error {
 	return process.ExecuteAndStreamOutput(kubectl, args...)
 }
 
+func KubectlGetNamespacesHavingSecrets(selector string) ([]string, error) {
```
vsukhin, line 816: is sellector used?

vsukhin (approved):
one typo

cmd/kubectl-testkube/commands/common/helper.go
exu: good catch - fixed (should be used ;) )

PR #6070, opened 2024-12-05
feat: [TKC-2966] Store TestWorkflowExecutionRequest config in TestWorkflowExecution

povilasv:
## Pull request description 

Store request config inside TestWorkflowExecution

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/testworkflows/testworkflowexecutor/executor.go
+++ b/pkg/testworkflows/testworkflowexecutor/executor.go
@@ -32,6 +32,7 @@ import (
 	configRepo "github.com/kubeshop/testkube/pkg/repository/config"
 	"github.com/kubeshop/testkube/pkg/repository/testworkflow"
 	"github.com/kubeshop/testkube/pkg/secretmanager"
+	"github.com/kubeshop/testkube/pkg/testworkflows"
 	"github.com/kubeshop/testkube/pkg/testworkflows/executionworker/controller"
 	"github.com/kubeshop/testkube/pkg/testworkflows/executionworker/executionworkertypes"
 	"github.com/kubeshop/testkube/pkg/testworkflows/testworkflowconfig"
@@ -45,6 +46,8 @@ const (
 
 	SaveLogsRetryMaxAttempts = 10
 	SaveLogsRetryBaseDelay   = 300 * time.Millisecond
+
+	ConfigSizeLimit = 3 * 1024 * 1024
 )
 
 //go:generate mockgen -destination=./mock_executor.go -package=testworkflowexecutor "github.com/kubeshop/testkube/pkg/testworkflows/testworkflowexecutor" TestWorkflowExecutor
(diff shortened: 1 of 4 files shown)


pkg/testworkflows/testworkflowexecutor/executor.go
```diff
@@ -465,6 +465,7 @@ func (e *executor) initialize(ctx context.Context, workflow *testworkflowsv1.Tes
 		DisableWebhooks:           request.DisableWebhooks,
 		Tags:                      map[string]string{},
 		RunningContext:            request.RunningContext,
+		Config:                    request.Config,
```
rangoo94, line 468: * It may contain sensitive data
* The data in the configuration may be huge in size

pkg/testworkflows/testworkflowexecutor/executor.go
```diff
@@ -467,7 +468,23 @@ func (e *executor) initialize(ctx context.Context, workflow *testworkflowsv1.Tes
 		RunningContext:            request.RunningContext,
 	}
 
-	// Try to resolve tags initially
+	// Store the configuration if it is small and not sensitive
+	if request.Config != nil && testworkflows.CountMapBytes(request.Config) < 10*1024*1024 {
```
rangoo94, line 472:
```suggestion
	if testworkflows.CountMapBytes(request.Config) < 10*1024*1024 {
```

Nullish `request.Config` is still a config - it's an empty map that may still match.

pkg/testworkflows/testworkflowexecutor/executor.go
```diff
@@ -467,7 +468,23 @@ func (e *executor) initialize(ctx context.Context, workflow *testworkflowsv1.Tes
...
+		storeConfig := true
+		schema := workflow.Spec.Config
+
+		for k, _ := range request.Config {
+			if s, ok := schema[k]; ok && s.Sensitive {
+				storeConfig = false
+			}
+		}
```
rangoo94, line 480: I'm not 100% sure, but could be that iterating over `workflow.Spec.Config` could be more correct - as a sensitive property set to default value may also be kind of sensitive *(similar to uncertainty when working with PII rules - single information may not be sensitive, but in conjunction with others - it can become sensitive)*

But, I guess, we can ignore that for now 👍

pkg/testworkflows/utils.go
```diff
@@ -108,3 +108,12 @@ func GetClusterID(ctx context.Context, configMap configRepo.Repository) string {
 	}
 	return clusterID
 }
+
+// CountMapBytes returns the total bytes of the map
+func CountMapBytes(m map[string]string) int {
```
rangoo94, line 113: It's only approximate, as the data will be stored in MongoDB as a BSON structure. As an example:
* `"♥♥♥♥♥♥♥♥♥♥"`
   * `len(str)` -> 30
   * BSON size -> 115 bytes
* `"XXXXXXXXXX"`
   * `len(str)` -> 10
   * BSON size -> 95 bytes

Considering that, I think that the best would be to decrease the limit above to - some still reasonable, but lower - 1-3 MB, to ensure that we won't hit limits due to conversion.

pkg/testworkflows/testworkflowexecutor/executor.go
rangoo94: Also, worth to move the size as constant

pkg/testworkflows/testworkflowexecutor/executor.go
rangoo94: > I.e. if it's kind of sensitive in relation to others, you can still set it to sensitive, right?

Yes, but I'm more saying about the default value - i.e. if you specify:

```yaml
spec:
  config:
    apiUrl: { type: string, default: "my-value", sensitve: true }
```

And you request execution with config:
```json
{}
```

The code above will save the configuration properly, despite it actually is using sensitive `my-value` value.

As mentioned though, I think that we can ignore it for now :)

rangoo94 (approved):

PR #6074, opened 2024-12-11
feat: [TKC-2895] pvc config

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/mapper/testworkflows/openapi_kube.go
```diff
@@ -1485,3 +1489,14 @@ func MapTestWorkflowTagSchemaAPIToKube(v testkube.TestWorkflowTagSchema) testwor
 		Tags: v.Tags,
 	}
 }
+
+func MapPvcConfigAPIToKube(v testkube.TestWorkflowPvcConfig) testworkflowsv1.TestWorkflowPvcConfig {
+	return testworkflowsv1.TestWorkflowPvcConfig{
+		Shared:           v.Shared,
```
rangoo94, line 1495: Should there be `shared` option? I think that PVC is in general shared, as long as you know its name

vsukhin: yes, the question is that most of pvc storage classes support single pod access mode and they will not work with multiple pods. should we ignore it and leave it tothe user or should we create pvc for each pod, if it's  not `shared``?

rangoo94: There is only one execution Pod; by "for each pod" do you mean that such PVC would be shared with multiple executions?

pkg/testworkflows/testworkflowresolver/merge.go
```diff
@@ -414,3 +415,35 @@ func MergeTags(dst, src map[string]string) map[string]string {
...
 	return dst
 }
+
+func MergePvcConfig(dst, include *testworkflowsv1.TestWorkflowPvcConfig) *testworkflowsv1.TestWorkflowPvcConfig {
+	if dst == nil {
+		return include
+	} else if include == nil {
+		return dst
+	}
+	if include.Shared {
```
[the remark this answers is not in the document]
vsukhin: the idea was to support pvcs defined in workflow and workflow template. if they are defined in both places, then merge them together, should we ignore template pvcs, if they are defined in workflow?

pkg/mapper/testworkflows/openapi_kube.go
rangoo94: Unless you mean the parallel steps and services; in such case `pvcs` should be possible to create for each layer, and accessible all the way down:

```yaml
spec:
  pvcs:
    # created for the whole execution (accessed only by execution)
    oneAtATime:
      storageClassName: standard
    # created for the whole execution (accessed by execution and all services and parallel steps)
    allAtATime:
      storageClassName: standard-many
      accessModes:
      - ReadWriteMany
  pod:
    volumes:
    - name: one-at-a-time
      persistentVolumeClaim:
        claimName: "{{ pvcs.oneAtATime.name }}"
    - name: all-at-a-time
      persistentVolumeClaim:
        claimName: "{{ pvcs.allAtATime.name }}"
  container:
    volumeMounts:
    - name: one-at-a-time
      mountPath: /private
    - name: all-at-a-time
      mountPath: /public
  services:
    count: 3
    pvcs:
      # create for each service instance (accessed only by that instance)
      oneServiceAtATime:
        storageClassName: standard
    pod:
      volumes:
      - name: one-service-at-a-time
        persistentVolumeClaim:
          claimName: "{{ pvcs.oneServiceAtATime.name }}"
      - name: all-at-a-time
        persistentVolumeClaim:
          claimName: "{{ pvcs.allAtATime.name }}"
    container:
      volumeMounts:
      - name: one-service-at-a-time
        mountPath: /private
      - name: all-at-a-time
        mountPath: /public
  steps:
  - parallel:
      count: 3
      pvcs:
        # create for each parallel step (accessed only by that parallel step)
        oneStepAtATime:
          storageClassName: standard
      pod:
        volumes:
        - name: one-step-at-a-time
          persistentVolumeClaim:
            claimName: "{{ pvcs.oneStepAtATime.name }}"
        - name: all-at-a-time
          persistentVolumeClaim:
            claimName: "{{ pvcs.allAtATime.name }}"
      container:
        volumeMounts:
        - name: one-step-at-a-time
          mountPath: /private
        - name: all-at-a-time
          mountPath: /public
```

vsukhin: I mean the executtion pod and service pods

pkg/testworkflows/testworkflowresolver/merge.go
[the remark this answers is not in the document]
rangoo94: In my opinion, the PVC definition is atomic, so shouldn't be extended, but rather overridden (when workflow have the same PVC name), or optionally failed (but rather it's better not)

pkg/mapper/testworkflows/openapi_kube.go
vsukhin: yes, this makese sense! thank you for suggestion!

pkg/testworkflows/testworkflowprocessor/intermediate.go
```diff
@@ -97,6 +107,18 @@ func (s *intermediate) AppendPodConfig(cfg *testworkflowsv1.PodConfig) Intermedi
 	return s
 }
 
+func (s *intermediate) AppendPvcs(cfg map[string]corev1.PersistentVolumeClaimSpec) Intermediate {
+	for name, spec := range cfg {
+		s.Ps[name] = corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: s.NextRef(),
```
rangoo94, line 114: More likely:
```suggestion
				Name: fmt.Sprintf("{{resource.id}}-%s", ref.NextRef()),
```

As the `NextRef()` is unique in the context of the Execution, but not necessarily between multiple executions

rangoo94 (approved):
One issue with unique names, besides LGTM 👍

pkg/testworkflows/testworkflowprocessor/intermediate.go
vsukhin: thank you, good catch, fixed!

PR #6089, opened 2024-12-18
feat: move execution scheduling to Control Plane

rangoo94:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/api/v1/testkube/model_test_workflow_extended.go
```diff
@@ -115,3 +120,19 @@ func (w TestWorkflow) HasService(name string) bool {
...
 	return false
 }
+
+func (w *TestWorkflow) DeepCopy() *TestWorkflow {
+	if w == nil {
+		return nil
+	}
+	v, err := json.Marshal(w)
+	if err != nil {
+		panic(fmt.Sprintf("cannot deep copy testkube.TestWorkflow: %v", err))
```
vsukhin, line 130: in Clone method marshaling errors are ignored, but panic in this method

rangoo94: No problem, changed to ignore errors - this path would never be hit anyway.

cmd/tcl/testworkflow-toolkit/commands/execute.go
```diff
@@ -202,69 +187,87 @@ func buildWorkflowExecution(workflow testworkflowsv1.StepExecuteWorkflow, async
...
+		var wg sync.WaitGroup
+		wg.Add(len(execs))
+		for i := range execs {
+			go func(exec testkube.TestWorkflowExecution) {
+				defer wg.Done()
+				prevStatus := testkube.QUEUED_TestWorkflowStatus
+			loop:
+				for {
+					// TODO: Consider real-time Notifications without logs instead
+					time.Sleep(200 * time.Millisecond)
```
vsukhin, line 221: I would ue const

pkg/runner/executionsaver.go
```diff
@@ -0,0 +1,135 @@
...
+	var prev *testkube.TestWorkflowResult
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case _, ok := <-ch:
+			if !ok {
+				return
+			}
+			for i := 0; i < 10; i++ {
```
vsukhin, line 70: I would use const

pkg/runner/runner.go
```diff
@@ -0,0 +1,238 @@
...
+	for i := 0; i < 10; i++ {
+		notifications = r.worker.Notifications(ctx, execution.Id, executionworkertypes.NotificationsOptions{})
+		if notifications.Err() == nil {
+			break
+		}
+		if errors.Is(notifications.Err(), registry.ErrResourceNotFound) {
+			// TODO: should it mark as job was aborted then?
+			return registry.ErrResourceNotFound
+		}
+		time.Sleep(500 * time.Millisecond)
```
vsukhin, line 78: possible const

pkg/runner/runner.go
```diff
@@ -0,0 +1,238 @@
...
+				continue
+			}
+			saver.UpdateResult(*n.Result)
+		} else {
+			if n.Ref != currentRef && n.Ref != "" {
+				currentRef = n.Ref
+				err = logs.WriteStart(n.Ref)
+				if err != nil {
+					// FIXME: what to do then?
+					panic("logs write ref error")
```
vsukhin, line 116: should we panic there, might be logging errors?

pkg/runner/runner.go
```diff
@@ -0,0 +1,238 @@
...
+				}
+			}
+			_, err = logs.Write([]byte(n.Log))
+			if err != nil {
+				// FIXME: what to do then?
+				panic("logs write error")
```
vsukhin, line 122: should we panic there, might be logging errors?

pkg/runner/runner.go
```diff
@@ -0,0 +1,238 @@
...
+		if lastResult == nil {
+			lastResult = execution.Result
+		}
+		if !lastResult.IsFinished() {
+			log.DefaultLogger.Errorw("failed to recover TestWorkflow result, marking as fatal error...", "id", execution.Id)
+			lastResult.Fatal(err, true, time.Now())
+		}
+	}
+
+	for i := 0; i < 100; i++ {
```
vsukhin, line 158: possible const

pkg/testworkflows/testworkflowexecutor/prepared.go
```diff
@@ -0,0 +1,347 @@
...
+	e.dirty = true
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *IntermediateExecution) ApplyDynamicConfig(config map[string]string) error {
+	if e.cr == nil {
+		panic("workflow not set yet")
```
vsukhin, line 206: should we panic or return an error?

vsukhin: there and below

vsukhin (approved):
lgtm. some minor comments

pkg/runner/runner.go
rangoo94: It shouldn't rather ever happen, so I have added an error log + `continue` (to handle it gracefully)

pkg/runner/runner.go
rangoo94: It shouldn't rather ever happen, so I have added an error log + continue (to handle it gracefully)

pkg/testworkflows/testworkflowexecutor/prepared.go
rangoo94: I think that we should panic - if something will try to call these methods before setting the workflow, it means that the code that is using this model is broken.

On the other hand, handling it gracefully should work in most of the cases (as long as the code that is using it has error handling), so I'm not sure.

Do you have some preference? Either way is fine for me - both have advantages and disadvantages.

vsukhin: if we return error, I would not panic. still it's up to you to decide

rangoo94: I have just noticed, that we have it also in other places in this file, but where it cannot be handled gracefully - I think that I'll just leave it there for consistency

rangoo94:
Closing in favor of https://github.com/kubeshop/testkube/pull/6089

PR #6090, opened 2024-12-18
feat: [TKC-3016] add defaultConfigs for TestWorkflow executions

povilasv:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/repository/testworkflow/mongo.go
+++ b/pkg/repository/testworkflow/mongo.go
@@ -21,7 +21,10 @@ import (
 
 var _ Repository = (*MongoRepository)(nil)
 
-const CollectionName = "testworkflowresults"
+const (
+	CollectionName       = "testworkflowresults"
+	configParamSizeLimit = 100
+)
 
 func NewMongoRepository(db *mongo.Database, allowDiskUse bool, opts ...MongoRepositoryOpt) *MongoRepository {
 	r := &MongoRepository{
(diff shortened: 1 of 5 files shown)


pkg/repository/testworkflow/mongo.go
```diff
@@ -255,16 +260,36 @@ func (r *MongoRepository) GetExecutionsSummary(ctx context.Context, filter Filte
...
+		executions[i].UnscapeDots()
 
-	for i := range result {
-		result[i].UnscapeDots()
+		if executions[i].ResolvedWorkflow != nil && executions[i].ResolvedWorkflow.Spec != nil {
+			for k, v := range executions[i].ResolvedWorkflow.Spec.Config {
+				if v.Default_ != nil {
+					if executions[i].ConfigParams == nil {
+						executions[i].ConfigParams = make(map[string]testkube.TestWorkflowExecutionConfigValue)
+					}
```
rangoo94, line 278: It should ignore then, as in such case it means that the passed configuration was ignored (i.e. sensitive)

pkg/repository/testworkflow/mongo.go
```diff
@@ -60,6 +60,34 @@ type MongoRepositoryOpt func(*MongoRepository)
...
+				}
+				if _, ok := result.ConfigParams[k]; !ok {
+					result.ConfigParams[k] = testkube.TestWorkflowExecutionConfigValue{
+						DefaultValue: v.Default_.Value,
+					}
+				} else {
+					value := result.ConfigParams[k].Value
+					truncated := false
+					if len(value) > 100 {
+						value = value[:100]
```
vsukhin, line 78: define const

pkg/repository/testworkflow/mongo.go
```diff
@@ -255,16 +288,43 @@ func (r *MongoRepository) GetExecutionsSummary(ctx context.Context, filter Filte
...
+						executions[i].ConfigParams = make(map[string]testkube.TestWorkflowExecutionConfigValue)
+					}
+					if _, ok := executions[i].ConfigParams[k]; !ok {
+						executions[i].ConfigParams[k] = testkube.TestWorkflowExecutionConfigValue{
+							DefaultValue: v.Default_.Value,
+						}
+					} else {
+						value := executions[i].ConfigParams[k].Value
+						truncated := false
+						if len(value) > 100 {
```
vsukhin, line 314: define const

pkg/repository/testworkflow/mongo.go
rangoo94: No ConfigParams should not be `nil`, but rather `map[string]string{}`.

vsukhin (approved):

pkg/repository/testworkflow/mongo.go
```diff
@@ -60,9 +63,39 @@ type MongoRepositoryOpt func(*MongoRepository)
...
 	err = r.Coll.FindOne(ctx, bson.M{"$or": bson.A{bson.M{"id": id}, bson.M{"name": id}}}).Decode(&result)
+
+	if result.ResolvedWorkflow != nil && result.ResolvedWorkflow.Spec != nil && result.ConfigParams != nil {
+		r.populateConfigParams(result.ResolvedWorkflow, result.ConfigParams)
+	}
+
 	return *result.UnscapeDots(), err
 }
 
+func (r *MongoRepository) populateConfigParams(resolvedWorkflow *testkube.TestWorkflow, configParams map[string]testkube.TestWorkflowExecutionConfigValue) {
```
rangoo94, line 74: Not important: I don't think it should be part of this struct, as it's completely independent

pkg/repository/testworkflow/mongo.go
```diff
@@ -241,8 +274,13 @@ func (r *MongoRepository) GetExecutions(ctx context.Context, filter Filter) (res
...
 }
 
+type TestWorkflowExecutionSummaryWithResolvedWorkflow struct {
+	testkube.TestWorkflowExecutionSummary `json:",inline" bson:",inline"`
+	ResolvedWorkflow                      *testkube.TestWorkflow `json:"resolvedWorkflow,omitempty"`
+}
+
 func (r *MongoRepository) GetExecutionsSummary(ctx context.Context, filter Filter) (result []testkube.TestWorkflowExecutionSummary, err error) {
-	result = make([]testkube.TestWorkflowExecutionSummary, 0)
+	executions := make([]TestWorkflowExecutionSummaryWithResolvedWorkflow, 0)
```
rangoo94, line 283: Better to avoid fetching unnecessary data there. You can either - change the query to fetch only important part, or even better - adjust the aggregate to provide these data.

rangoo94: Also, we have different pipelines for Enterprise, so it's very important to test it through.

rangoo94 (approved):

PR #6093, opened 2024-12-19
feat: [TKC-3016] fix defaultConfigs for TestWorkflow executions

povilasv:
## Pull request description 

When request.Config is empty we don't provide defaultValues.

This fixes that, but also keeps so when sensitive is provided no values are returned

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


exu (approved):

rangoo94 (commented):
Why we should populate the default values in that case? It will make it impossible to determine between [lack of config] vs [deleted config (sensitive value)] vs [empty string].

This means, that both FE and Re-Run won't know if the configParams is empty because defaults have been used, or is empty because sensitive (or too big) data are used.

PR #6153, opened 2025-02-05
fix: fix crash due to nil pointer

WitoDelnat:
Fixes a crash due to a nil pointer.


pkg/runner/agent.go
```diff
@@ -87,7 +88,7 @@ func (a *agentLoop) _saveEmptyLogs(ctx context.Context, environmentId string, ex
 	if execution.Workflow != nil {
 		workflowName = execution.Workflow.Name
 	}
-	return a.client.SaveExecutionLogs(ctx, environmentId, execution.Id, workflowName, nil)
+	return a.client.SaveExecutionLogs(ctx, environmentId, execution.Id, workflowName, strings.NewReader(""))
```
WitoDelnat, line 91: This would go all the way to pkg/bufferedstream/bufferedstream.go:33, which then leads to a panic in stdlib /opt/homebrew/opt/go/libexec/src/io/io.go:429: `nr, er := src.Read(buf)`.

rangoo94: This rather should be specifically `nil` - otherwise the behavior may be ambiguous. The problem is Golang's `http` implementation - if the `nil` or empty `bytes.Buffer` is passed - it's correctly assigning `Content-Length`, otherwise it may not assign `Content-Length` leading to errors on proxy/S3.

rangoo94: Nevermind, it's passed through the `BufferedStream` anyway 👍

rangoo94 (approved):

build/_local/agent-server.Dockerfile
```diff
@@ -36,10 +37,9 @@ ENTRYPOINT ["/go/bin/dlv", "exec", "--headless", "--continue", "--accept-multicl
 ###################################
 ## Distribution
 ###################################
-FROM scratch AS dist
+FROM gcr.io/distroless/static AS dist
```
rangoo94, line 40: Is that expected?

WitoDelnat: From the skaffold docs
> our application should be built with the -gcflags='all=-N -l' flag to disable optimizations and inlining. If you do not add this flag, debugging can becoming confusing due to seemingly-random execution jumps that happen due to statement reordering and inlining. Skaffold configures Docker builds with a SKAFFOLD_GO_GCFLAGS build argument that you should use in your Dockerfile:
>
> **Note that the Alpine Linux-based golang container images do not include a C compiler which is required to build with -gcflags='all=-N -l'.**

When going from alpine to default the apk package manager needed to be changes. We only use this to get ca-certificates which [distroless/base](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md) does for us. It just felt the easier path. Since this is only for local development it's a good way to test whether this base image works well before starting to use it elsewhere.

rangoo94: Makes sense, thank you!

PR #6204, opened 2025-02-27
feat: [TKC-3299] move cronjobs

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/cronjob/executor.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/kubeshop/testkube/pkg/testworkflows/testworkflowexecutor"
+	"github.com/kubeshop/testkube/pkg/workerpool"
+)
+
+const (
+	concurrencyLevel = 10
+)
+
+func (s *Scheduler) executeTestWorkflow(ctx context.Context, testWorkflowName string, cron *testkube.TestWorkflowCronJobConfig) {
+	request := &cloud.ScheduleRequest{
```
rangoo94, line 23: Not sure about edge case - shouldn't it check the last execution? Otherwise each service re-initialization will i.e. reconsider again i.e. `* * * * *` cronjob

vsukhin: i can try to reproduce, tif it happens

PR #6258, opened 2025-03-19
chore: fix build args

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -1,15 +1,16 @@
-variable "GOCACHE" {
-  default = "/go/pkg"
-}
-variable "GOMODCACHE" {
-  default = "/root/.cache/go-build"
-}
-variable "BUSYBOX_IMAGE" {
-  default = "busybox:1.37.0-musl"
-}
-variable "ALPINE_IMAGE" {
-  default = "alpine:3.21.3"
-}
+variable "GOCACHE"       { default = "/go/pkg" }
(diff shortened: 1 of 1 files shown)


vsukhin (approved):

ypoplavs (approved):

docker-bake.hcl
```diff
@@ -22,6 +22,14 @@ target "api" {
   dockerfile = "build/new/api-server.Dockerfile"
   platforms = ["linux/arm64", "linux/amd64"]
   args = {
+    VERSION = "${VERSION}"
```
WitoDelnat, line 25: I might actually need to define all these variables as I did for the ones above at line 4-12

WitoDelnat: Can confirm, you must define all these variables before you can inject them through env variables

```
$ docker buildx bake --set api.platform=linux/arm64 --set api.args.GOCACHE="/Users/wito/Library/Caches/go-build" --set api.args.GOMODCACHE="/Users/wito/go/pkg/mod" api

ERROR: docker-bake.hcl:26,18-25: Unknown variable; There is no variable named "GIT_SHA"., and 7 other diagnostic(s)
```

PR #6283, opened 2025-03-25
feat: add targets to triggers

WitoDelnat:
This PR contributes to https://linear.app/kubeshop/issue/TKC-3444/add-targeting-to-trigger-execute-and-cron


api/v1/testkube.yaml
```diff
@@ -8716,6 +8719,8 @@ components:
             - false
         tags:
           $ref: "#/components/schemas/TestWorkflowTagValue"
+        target:
+          $ref: "#/components/schemas/ExecutionTarget"
```
rangoo94, line 8723: There is already `runnerTarget` and `runnerOriginalTarget`.

WitoDelnat: wdym? As for the component ref? These use the same references:

```
        runnerTarget:
          $ref: "#/components/schemas/ExecutionTarget"
        runnerOriginalTarget:
          $ref: "#/components/schemas/ExecutionTarget"
```

internal/app/api/v1/testtriggers.go
```diff
@@ -116,10 +116,9 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 		testTrigger.Spec = crdTestTrigger.Spec
 		testTrigger.Labels = request.Labels
 		testTrigger.Annotations = request.Annotations
-		testTrigger, err = s.TestTriggersClient.Update(testTrigger)
 
 		s.Metrics.IncUpdateTestTrigger(err)
-
+		testTrigger, err = s.TestTriggersClient.Update(testTrigger)
```
rangoo94, line 121: Should be back the other way around. Otherwise `err` won't be passed to the metric

pkg/api/v1/testkube/model_test_workflow_execution.go
```diff
@@ -49,6 +49,7 @@ type TestWorkflowExecution struct {
 	// whether webhooks on the execution of this test workflow are disabled
 	DisableWebhooks bool                                        `json:"disableWebhooks,omitempty"`
 	Tags            map[string]string                           `json:"tags,omitempty"`
+	Target          *ExecutionTarget                            `json:"target,omitempty"`
```
rangoo94, line 52: As above, it is not needed. There is already `RunnerTarget` and `RunnerOriginalTarget`

pkg/crd/templates/testtrigger.tmpl
```diff
@@ -103,6 +103,39 @@ spec:
...
+      {{- range $key, $value := .ActionParameters.Tags }}
+      {{ $key }}: {{ $value }}
+      {{- end }}
+    {{- end }}
+    {{- if .ActionParameters.Target }}
+    target:
+      {{- if .ActionParameters.Target.Match }}
+      match:
+        {{- range $key, $value := .ActionParameters.Target.Match }}
+        {{ $key }}: {{ $value }}
```
rangoo94, line 125: Does it work? `$value` here is a `[]string`

pkg/crd/templates/testtrigger.tmpl
```diff
@@ -103,6 +103,39 @@ spec:
...
+        {{- end }}
+      {{- end }}
+      {{- if .ActionParameters.Target.Not }}
+      not:
+        {{- range $key, $value := .ActionParameters.Target.Not }}
+        {{ $key }}: {{ $value }}
+        {{- end }}
+      {{- end }}
+      {{- if .ActionParameters.Target.Replicate }}
+      replicate: {{ .ActionParameters.Target.Replicate }}
```
rangoo94, line 135: Does it work? `$value` here is a `[]string`

pkg/crd/templates/testtrigger.tmpl
```diff
@@ -103,6 +103,48 @@ spec:
   {{- if .Action }}
   action: {{ .Action }}
   {{- end }}
+  {{- if .ActionParameters }}
+  actionParameters:
+    {{- if ne (len .ActionParameters.Config) 0 }}
+    config:
+      {{- range $key, $value := .ActionParametersConfig }}
+      {{ $key }}: {{ $value }}
```
rangoo94, line 111: Will it work, given that there may be any data deep down in the `$value`?

WitoDelnat: This is just `Config map[string]string `json:"config,omitempty"`` right?

rangoo94 (approved):

PR #6288, opened 2025-03-26
feat: allow transferring env variables from Agent to Execution pod / transfer gRPC ALPN config

rangoo94:
## Pull request description 

* allow transferring env variables from Agent to Execution pod
* transfer gRPC ALPN configuration environment variable

## Checklist (choose what happened)

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -188,30 +188,31 @@ type Config struct {
 	ImageInspectorConfig
 	GitOpsSyncConfig
 	CronJobConfig
-	DisableDefaultAgent             bool   `envconfig:"DISABLE_DEFAULT_AGENT" default:"false"`
-	TestkubeConfigDir               string `envconfig:"TESTKUBE_CONFIG_DIR" default:"config"`
-	TestkubeAnalyticsEnabled        bool   `envconfig:"TESTKUBE_ANALYTICS_ENABLED" default:"false"`
-	TestkubeNamespace               string `envconfig:"TESTKUBE_NAMESPACE" default:"testkube"`
-	TestkubeProWorkerCount          int    `envconfig:"TESTKUBE_PRO_WORKER_COUNT" default:"50"`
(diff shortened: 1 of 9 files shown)


internal/config/config.go
```diff
@@ -188,30 +188,31 @@ type Config struct {
...
+	TestkubeHelmchartVersion        string   `envconfig:"TESTKUBE_HELMCHART_VERSION" default:""`
+	DebugListenAddr                 string   `envconfig:"DEBUG_LISTEN_ADDR" default:"0.0.0.0:1337"`
+	EnableDebugServer               bool     `envconfig:"ENABLE_DEBUG_SERVER" default:"false"`
+	Debug                           bool     `envconfig:"DEBUG" default:"false"`
+	Trace                           bool     `envconfig:"TRACE" default:"false"`
+	DisableSecretCreation           bool     `envconfig:"DISABLE_SECRET_CREATION" default:"false"`
+	TestkubeExecutionNamespaces     string   `envconfig:"TESTKUBE_EXECUTION_NAMESPACES" default:""`
+	GlobalWorkflowTemplateName      string   `envconfig:"TESTKUBE_GLOBAL_WORKFLOW_TEMPLATE_NAME" default:""`
+	GlobalWorkflowTemplateInline    string   `envconfig:"TESTKUBE_GLOBAL_WORKFLOW_TEMPLATE_INLINE" default:""`
+	TransferEnvVariables            []string `envconfig:"TRANSFER_ENV_VARS" default:"GRPC_ENFORCE_ALPN_ENABLED"`
```
dejanzele, line 211: Should we keep it as `default` here, as users will need to remember to add it, in case they define `TRANSFER_ENV_VARS ` for some other variable?

dejanzele (approved):

internal/config/config.go
rangoo94: It depends what we want to achieve. I believe it will be used only for workarounds. In any case (1) we can enforce it on Helm Chart, (2) this env variable will be barely needed, (3) it's deprecated and soon deleted.

PR #6304, opened 2025-03-31
fix: pass execution copy

vsukhin:
## Pull request description 

Becuase Wokflow spec is passed using pointers in execution, we need to clone them before insert/update

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/services/controlplane.go
+++ b/cmd/api-server/services/controlplane.go
@@ -293,10 +293,10 @@ func CreateControlPlane(ctx context.Context, cfg *config.Config, features featur
 			return
 		}),
 		cloudtestworkflow.CmdTestWorkflowExecutionInsert: controlplane.Handler(func(ctx context.Context, data cloudtestworkflow.ExecutionInsertRequest) (r cloudtestworkflow.ExecutionInsertResponse, err error) {
-			return r, testWorkflowResultsRepository.Insert(ctx, data.WorkflowExecution)
+			return r, testWorkflowResultsRepository.Insert(ctx, *data.WorkflowExecution.Clone())
 		}),
 		cloudtestworkflow.CmdTestWorkflowExecutionUpdate: controlplane.Handler(func(ctx context.Context, data cloudtestworkflow.ExecutionUpdateRequest) (r cloudtestworkflow.ExecutionUpdateResponse, err error) {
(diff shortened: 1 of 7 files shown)


cmd/api-server/services/controlplane.go
```diff
@@ -293,10 +293,10 @@ func CreateControlPlane(ctx context.Context, cfg *config.Config, features featur
 			return
 		}),
 		cloudtestworkflow.CmdTestWorkflowExecutionInsert: controlplane.Handler(func(ctx context.Context, data cloudtestworkflow.ExecutionInsertRequest) (r cloudtestworkflow.ExecutionInsertResponse, err error) {
-			return r, testWorkflowResultsRepository.Insert(ctx, data.WorkflowExecution)
+			return r, testWorkflowResultsRepository.Insert(ctx, *data.WorkflowExecution.Clone())
```
rangoo94, line 296: Is there a reason to clone it there? It's serialized immediately.

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -117,7 +117,7 @@ func NewScheduler(
 
 func (s *scheduler) insert(ctx context.Context, execution *testkube.TestWorkflowExecution) error {
 	err := retry(SaveResultRetryMaxAttempts, SaveResultRetryBaseDelay, func() error {
-		err := s.resultsRepository.Insert(ctx, *execution)
+		err := s.resultsRepository.Insert(ctx, *execution.Clone())
```
rangoo94, line 120: Is there a reason to clone it? It's not used during save.

rangoo94, line 120: This should rather be cloned inside of repository. Otherwise the `Insert()`/`Update()` methods in the repository still have side effect.

vsukhin: agree - better!

cmd/api-server/services/controlplane.go
vsukhin: will change

rangoo94 (approved):

PR #6328, opened 2025-04-08
feat: [TKC-3371] add env service flag

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/tcl/testworkflow-toolkit/commands/services.go
```diff
@@ -159,6 +160,7 @@ func NewServicesCmd() *cobra.Command {
 						Pvcs: svcSpec.Pvcs,
 					}
 					spec.Steps[0].Run.ContainerConfig = testworkflowsv1.ContainerConfig{}
+					spec.Container.Env = testworkflowresolver.DedupeEnvVars(append(spec.Container.Env, config.Config().Execution.GlobalEnv...))
```
rangoo94, line 163: Shouldn't it preserve order? In the current way, the global environments will have precedence over overrides.

vsukhin: agree, it's better

PR #6334, opened 2025-04-10
fix: remove file closing

vsukhin:

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


rangoo94 (commented):
Won't this leave open handler?

vsukhin:
Actually, this is a reverse change, becuase it doens't work now. Sounds like it's closed

// SetBodyStream sets response body stream and, optionally body size.
//
// If bodySize is >= 0, then the bodyStream must provide exactly bodySize bytes
// before returning io.EOF.
//
// If bodySize < 0, then bodyStream is read until io.EOF.
//
**// bodyStream.Close() is called after finishing reading all body data
// if it implements io.Closer.**
//
// See also SetBodyStreamWriter.

rangoo94:
BodyStream is an internal struct in there, which after `Close()` is not closing the underlying stream (in this case `file`).

Issue #6194, opened 2025-02-24
Addition of the hostPID property to TestWorkflowSpec


FraNemo:
**Is your feature request related to a problem? Please describe.**

I'm trying to perform a security analysis with kube-bench on my minikube cluster (k8s 1.28) through a TestWorkflow. As requested from the tool documentation (under "Running in a Kubernetes cluster" https://aquasecurity.github.io/kube-bench/v0.6.15/running/), when kube-bench is run inside a pod it requires the access to the host's PID namespace in order to completely execute the scan. This is allowed by the property "hostPID" set to "true" in the Job manifest. This field is not available in a TestWorkflow right now. 

**Describe the solution you'd like**

Addition of the "hostPID" field to TestWorkflowSpec in order to allow this type of security test.

vsukhin:
thank you @FraNemo cc @olensmar @jmorante-ks

vsukhin:
hey, @FraNemo sorry for delay, it's possible to add, but can lead to some potentail security issues.
pinging @olensmar @jmorante-ks for prioritisation

FraNemo:
No problem @vsukhin, i'm not in a hurry regarding this feature request. If you consider it too risky to be added i'm ok with that. Thank you!

vsukhin:
Probably. we will add one more helm flag, like allowLowSecurity

FraNemo:
Thank you!

PR #6354, opened 2025-04-21
fix: hostpid field

vsukhin:
## Pull request description 

#6194 support hostpid field

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/testworkflows/testworkflowprocessor/processor.go
```diff
@@ -384,6 +386,12 @@ func (p *processor) Bundle(ctx context.Context, workflow *testworkflowsv1.TestWo
 	if podConfig.SecurityContext.FSGroup == nil {
 		podConfig.SecurityContext.FSGroup = common.Ptr(constants.DefaultFsGroup)
 	}
+	hostPID := false
+	if options.AllowLowSecurityFields {
+		if podConfig.HostPID != nil {
+			hostPID = *podConfig.HostPID
```
rangoo94, line 392: Shouldn't it throw error instead if not allowed?

vsukhin: agree, should be better

PR #6386, opened 2025-05-02
feat: Add Execution CRD control to Agent

admacleod:
## Pull request description 

These changes add Kubernetes controllers for the `TestWorkflowExecution` CRD as well as the `TestSuiteExecution` and `TestExecution` legacy CRDs.
The controllers mimic the effect of the `testkube-operator` on the CRDs, however as they are integrated into the agent they are able to call the execution functions directly rather than having to go via the API.

Initialisation of the controllers is behind a new `ENABLE_K8S_CONTROLLERS` environment variable as it would not be desirable to have these controllers running at the same time as the `testkube-operator`.

Additionally the Agent will now also attempt to clean up legacy `CronJob` resources when using the newer cronjob scheduler. This is a duplication of the existing logic in the operator, however the two should not conflict and enables fully operator-less deployment.

The TestKube CLI `devbox` command has also been updated to create the correct permissions for the `Role` as well as to include a new `--enable-k8s-controllers` flag.

- [x] tested locally
- [ ] tested on cluster
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Changes

- Added Execution CRD controllers to Agent behind `ENABLE_K8S_CONTROLLERS` environment variable.

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -11,10 +11,13 @@ import (
 	"github.com/gofiber/fiber/v2/middleware/cors"
 	"google.golang.org/grpc"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	k8sctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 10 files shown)


cmd/api-server/main.go
```diff
@@ -603,6 +662,30 @@ func main() {
 		&proContext,
 	)
 	if scheduler != nil {
+		// Remove any remaining legacy cronjobs.
```
vsukhin, line 665: Stylish (up to you) I would move it to separate func

admacleod: I couldn't remember the conversation we had last week, did we agree that this code would be temporary (just until the next major release) or that it would have to stick around for a long time? If it is longer I will move this, otherwise it can stay as an obvious "to be removed" section.

vsukhin: ok, add the comment then - that it's temporal code

PR #6388, opened 2025-05-05
feat: [TKC-3569] dry-run for spec check against crd schema

vsukhin:
## Pull request description 

- add dry-run flag to cli for test workflow and test workflow template creation/modification
- use operator schemas for above resources and add validation method against the schema
- add server  validation methods for these resources
- add client validation methods for these resources

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


rangoo94:
1. Should we copy schema to another place? We have it already in `testkube-operator` and `helm-charts`, maybe it's possible to reference the `testkube-operator` one?
2. Validation, if not `--dry-run`, should take place in the Control Plane (or, in case of OSS, the corresponding API endpoint). The CLI may have a different version than the Agent / Control Plane, what could lead to faulty results

vsukhin:
1. Wll try, if i could
2. Yes, good point. Actually, cli is more like poc. Will move to api methods

PR #6402, opened 2025-05-13
fix: [TKC-3639] mask secret volume data

vsukhin:
## Pull request description 

Mask secret volume data

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/testworkflow-init/orchestration/setup.go
+++ b/cmd/testworkflow-init/orchestration/setup.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	scopedRegex              = regexp.MustCompile(`^_(00|01|02|03|04|\d|[1-9]\d*)(C)?(S?)_`)
+	scopedRegex              = regexp.MustCompile(`^_(00|01|02|03|04|05|\d|[1-9]\d*)(C)?(S?)_`)
 	Setup                    = newSetup()
 	defaultWorkingDir        = getWorkingDir()
 	commonSensitiveVariables = []string{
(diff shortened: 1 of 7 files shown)


cmd/testworkflow-init/orchestration/setup.go
```diff
@@ -307,3 +308,37 @@ func (c *setup) SetConfig(config lite.LiteContainerConfig) {
...
+				return nil
+			}
+
+			if !info.Mode().IsRegular() {
+				return nil
+			}
+
+			data, err := os.ReadFile(path)
+			if err != nil {
+				return err
```
rangoo94, line 326: Shouldn't it continue?

pkg/testworkflows/testworkflowprocessor/processor.go
```diff
@@ -356,6 +366,9 @@ func (p *processor) Bundle(ctx context.Context, workflow *testworkflowsv1.TestWo
 			if !filepath.IsAbs(containers[i].VolumeMounts[j].MountPath) {
 				containers[i].VolumeMounts[j].MountPath = filepath.Clean(filepath.Join(workingDir, containers[i].VolumeMounts[j].MountPath))
 			}
+			if _, ok := volumeNameMap[containers[i].VolumeMounts[j].Name]; ok {
```
rangoo94, line 369: The mounted volumes will differ per container, so there should be container or step granularity. Otherwise, you may use directory that is not mounted and treat it as secret volume:

```
steps:
  pod:
    volumes:
    - name: secret-vol
      secret: { secretName: "some-secret" }
- container:
     volumeMounts:
     - mountPath: /usr/bin
        name: "secret-vol"
  shell: cat /usr/bin/something
- shell: "echo this step will take looooong as image's /usr/bin will be treat as secret"
```

vsukhin: looks like an edge case, but agree, need container granularity

PR #6407, opened 2025-05-20
chore: standardise protobuf generation

admacleod:
## Pull request description 

Switching to a containerised protobuf generation ensures that we're always using the correct versions of tools and developers do not need to have those tools installed locally to generate code.


admacleod:
Question for @kubeshop/testkube-backend: do we want to take this opportunity to standardise the tool versions for protobuf generation? Currently the `logs` package is using a newer set of tools than the `cloud` package. I wasn't sure if it was appropriate to do in this PR but I can add a new commit doing so. Also relocating the `logs.proto` definition to the top-level `protos` directory to keep our proto definitions in the same location.

Makefile
```diff
@@ -263,20 +250,3 @@ port-forward-api:
 
 run-proxy:
 	go run cmd/proxy/main.go --namespace $(NAMESPACE)
-
```
vsukhin, line 266: it's fine to use one more opttion like Dockerfile for protoc, but I don't think we need to remove existing ones

admacleod: I don't think this change is worthwhile if we're just going to add to the ever growing list of ways to achieve the same result here. We already have `protobuf-generate` **_and_** `generate-protobuf` as recipes in this `Makefile`, which seems kind of ridiculous in the first place (especially as you cannot tell from the recipe name what either actually does).

This change is about attempting to standardise the build tooling and versioning that we are using. I would like to take this change further, move `/pkg/logs/pb/logs.proto` up to `/proto` and combine both recipes into a single one that builds all protos in the `/proto` directory. I chose not to right now because for some reason that proto was built using a different tool set version than the `service.proto` was, which further highlights the issue we currently have with installing the tooling on people's machines without any standardisation. There just currently is not a clear single way to produce generated code from proto definitions in this project.

PR #6442, opened 2025-06-03
feat:[TKC-3759]  sql db scheme and queries

vsukhin:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/database/migrations/20250603170830_init.sql
```diff
@@ -0,0 +1,62 @@
...
+    scheduled_at TIMESTAMP WITH TIME ZONE,
+    assigned_at TIMESTAMP WITH TIME ZONE,
+    status_at TIMESTAMP WITH TIME ZONE,
+    signature JSONB,
+    result JSONB,
+    output JSONB,
+    reports JSONB,
+    resource_aggregations JSONB,
+    workflow JSONB,
+    resolved_workflow JSONB,
```
WitoDelnat: I know this is all still early draft. But putting some thoughts down while looking at this PR. 

Workflow and resolved_workflow are **HUGE** json documents. It could be beneficial to experiment with extracting those to a separate table.

I also guess we want to re-use all of this code in the control plane like we did in the past? In this case, it might be worth to also introduce some mechanism to avoid needing table per org. For instance, a `tenant_id` which for open source is always the same constant and the org_id for control plane. We'd still need a second "abstract" column for environment; I guess `sub_tenant_id` does not really work 😋 Maybe the tenant_id would simply be org_id and env_id concatenated? Anyway I'm definitely running a bit ahead of this project here.

vsukhin: Sure, it's just a first pass. I will keep it mind on next iterations!

vsukhin: And thank you for earlier review!

vsukhin: ask AI to split to multiple tables

vsukhin: For control plane we will use a bit different db scheme

pkg/controlplane/server.go
```diff
@@ -691,3 +695,7 @@ func (s *Server) SaveExecutionArtifactPresigned(ctx context.Context, req *cloud.
 	}
 	return &cloud.SaveExecutionArtifactPresignedResponse{Url: url}, nil
 }
+
+func (s *Server) GetRepositoryMaanger() repository.DatabaseRepository {
```
WitoDelnat: `GetRepositoryMaanger` tiny typo here

pkg/database/migrations/20250603170830_init.sql
olensmar: will that make it hard to migrate data from standalone to control-plane?

vsukhin: no, will just have 2 additional fields ennvironment_id and organization_id, they will be filled with current env and ord id if we need to migrate from OSS to Enterprise. Actually, we might want to use materialied views in control plave with env and org id filter, so even queries might be the same just to work with views instead of tables

pkg/controlplane/server.go
vsukhin: fixed

PR #6445, opened 2025-06-05
feat: [TKC-3737] workflow tests - XUnit example with JUnit report

tkonieczny:


devcatalin (approved):

olensmar (approved):
nice - could we add this to the docs also?

PR #6453, opened 2025-06-11
feat: [TKC-3656] espresso+android emulator - example workflow

tkonieczny:
Example espresso app + tests + workflow (K8s+KVM).
Gradlew binary removed on purpose (despite https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper recommendations) - it's not intended to be run locally - need to handled then.


test/espresso/example-project/gradle/wrapper/gradle-wrapper.jar
WitoDelnat, line 1: Do we want to commit the jar 🤔 ?

This appears to be something from gradle though and not the Android app itself? Still a bit weird that we're committing binaries into git.

tkonieczny: I was surprised it's not included in the gitignore by default, so I checked and based on Gradle docs:

> To make the Wrapper files available to other developers and execution environments, you need to check them into version control. Wrapper files, including the JAR file, are small. Adding the JAR file to version control is expected.(...)

(https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper)

I think in our case (running things in container using specific gradle image) it won't be needed and it should be fine to remove any gradle files other than `gradle.properties`, but I need to check. Assuming it's not intended to be ever run "locally"

tkonieczny: Removed gradlew binary, workflow works fine - it's not intended to be run locally anyway.

WitoDelnat (approved):

PR #6464, opened 2025-06-20
fix: [TKC-3819] waiting for multiple running workflows

vsukhin:
## Pull request description 

Use waiting flag when run test workflows using label selector

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/testworkflows/run.go
+++ b/cmd/kubectl-testkube/commands/testworkflows/run.go
@@ -191,16 +191,10 @@ func NewRunTestWorkflowCmd() *cobra.Command {
 							}
 
 							if len(args) > 0 {
-								ec := uiWatch(execution, pServiceName, serviceIndex, pParallelStepName, parallelStepIndex, client)
+								exitCode = uiWatch(execution, pServiceName, serviceIndex, pParallelStepName, parallelStepIndex, client)
 								ui.NL()
 								if downloadArtifactsEnabled {
 									tests.DownloadTestWorkflowArtifacts(execution.Id, downloadDir, format, masks, client, outputPretty)
-								}
-
-								if ec != 0 {
-									mu.Lock()
-									exitCode = ec
-									mu.Unlock()
 								}
 							} else {
 								wg.Add(1)
@@ -225,7 +219,6 @@ func NewRunTestWorkflowCmd() *cobra.Command {
 										mu.Unlock()
 									}
 								}(&execution)
-
 							}
 						} else {
 							uiShellWatchExecution(execution.Id)


cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -187,10 +190,42 @@ func NewRunTestWorkflowCmd() *cobra.Command {
...
+									options := []Options{{Prefix: prefix}}
+									ec := uiWatch(*exec, pServiceName, serviceIndex, pParallelStepName, parallelStepIndex, client, options...)
+									ui.NL()
+									if downloadArtifactsEnabled {
+										tests.DownloadTestWorkflowArtifacts(exec.Id, downloadDir, format, masks, client, outputPretty)
+									}
+
+									if ec != 0 {
+										mu.Lock()
+										exitCode = ec
```
ed382:
```suggestion
										exitCode |= ec
```

There might be a race condition here between the various go routines and determining the exit code, maybe you just want to coalesce the exit code from all of them. Since it seems it only returns 0 or 1, you can do something like this.

cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -187,10 +190,42 @@ func NewRunTestWorkflowCmd() *cobra.Command {
 								pParallelStepName = &parallelStepName
 							}
 
-							exitCode = uiWatch(execution, pServiceName, serviceIndex, pParallelStepName, parallelStepIndex, client)
-							ui.NL()
-							if downloadArtifactsEnabled {
-								tests.DownloadTestWorkflowArtifacts(execution.Id, downloadDir, format, masks, client, outputPretty)
+							if len(args) > 0 {
```
ed382, line 193: This is a bit hard to read, the predicate `len(args) > 0` seems to indicate that there should be only one execution. Maybe this block can be removed entirely as the main difference with the block below seems to be the `prefix` configuration which maybe can be blank based on whether there is only one execution.

Yes, the goroutine would not be required there in case there is one execution, but maybe it would simplify the code.

cmd/kubectl-testkube/commands/testworkflows/run.go
vsukhin: good idea

cmd/kubectl-testkube/commands/testworkflows/run.go
vsukhin: yes, I will change it

PR #6482, opened 2025-07-04
fix: assignment lock logic

ed382:
## Pull request description 

The locking via assignment logic was not quite tight.

The runnerid can be a non-nil blank string, because when there is an error sending start request to the runner the orchestrator immediately tries to unassign the execution by setting it to blank string. If this happens, then it seems the execution will then be unassignable through the current logic. The `GetUnassigned` logic takes this case into account returning the values with blank strings as well, but it does not seem `Assign` is able to handle such a precondition.

Another possible problem is with reassignment, before reassigning an execution to a different runner, the lock should enforce an assignment timeout this should help avoid multiple orchestrator instances from reassigning an execution in short order. Particularly, this could be happening in the recovery loop after an execution has been assigned for one minute or longer. The one minute timeout in these conditions comes from the recovery loop period and other preexisting logic that seems to attempt to do the same thing without accounting for concurrency of orchestrator instances.

The immediate unassigning on start request failure will need to be removed also because of the above, and in this cases it should just fall back to the one minute recovery loop for the sake of simplicity.

Other guard rails put in place to make sure the assigned at time monotonicly increases.

- [ ] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/repository/testworkflow/mongo.go
+++ b/pkg/repository/testworkflow/mongo.go
@@ -780,11 +780,35 @@ func (r *MongoRepository) Init(ctx context.Context, id string, data InitData) er
 }
 
 func (r *MongoRepository) Assign(ctx context.Context, id string, prevRunnerId string, newRunnerId string, assignedAt *time.Time) (bool, error) {
+	oneMinuteAgo := time.Now().Add(-1 * time.Minute)
 	res, err := r.Coll.UpdateOne(ctx, bson.M{
 		"$and": []bson.M{
 			{"id": id},
 			{"result.status": testkube.QUEUED_TestWorkflowStatus},
-			{"$or": []bson.M{{"runnerid": prevRunnerId}, {"runnerid": newRunnerId}, {"runnerid": nil}}},
+			{"$or": []bson.M{
+				// New assignment - workflow has no runner assigned
+				{
+					"$or": []bson.M{
+						{"runnerid": nil},
+						{"runnerid": ""},
+					},
+				},
+				// Extension of existing assignment - extension to assignment timeout
+				{
+					"$and": []bson.M{
+						{"runnerid": newRunnerId},
+						{"assignedat": bson.M{"$lt": assignedAt}},
+					},
+				},
+				// Reassignment to new runner - must wait one minute between assignments
+				{
+					"$and": []bson.M{
+						{"runnerid": prevRunnerId},
+						{"assignedat": bson.M{"$lt": oneMinuteAgo}},
+						{"assignedat": bson.M{"$lt": assignedAt}},
+					},
+				},
+			}},
 		},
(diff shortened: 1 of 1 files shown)


pkg/repository/testworkflow/mongo.go
```diff
@@ -780,11 +780,35 @@ func (r *MongoRepository) Init(ctx context.Context, id string, data InitData) er
 }
 
 func (r *MongoRepository) Assign(ctx context.Context, id string, prevRunnerId string, newRunnerId string, assignedAt *time.Time) (bool, error) {
+	oneMinuteAgo := time.Now().Add(-1 * time.Minute)
```
WitoDelnat, line 783: > Particularly, this could be happening in the recovery loop after an execution has been assigned for one minute or longer. 

I'm double checking to see if I understand this sentence correctly. There is a recovery loop somewhere which - after a minute - will try to reassign? If so, I wonder if we should perhaps take 55 seconds here then so that we do not have race conditions in this timing?

pkg/repository/testworkflow/mongo.go
```diff
@@ -780,11 +780,35 @@ func (r *MongoRepository) Init(ctx context.Context, id string, data InitData) er
...
 		"$and": []bson.M{
 			{"id": id},
 			{"result.status": testkube.QUEUED_TestWorkflowStatus},
-			{"$or": []bson.M{{"runnerid": prevRunnerId}, {"runnerid": newRunnerId}, {"runnerid": nil}}},
+			{"$or": []bson.M{
+				// New assignment - workflow has no runner assigned
+				{
+					"$or": []bson.M{
+						{"runnerid": nil},
+						{"runnerid": ""},
```
WitoDelnat, line 793: Great catch! This could possibly be the cause of workflows stuck in queue indefinitely!

pkg/repository/testworkflow/mongo.go
ed382: The loop is in the control plane instances of the orchestrator, so all these loops will be out of sync, but this enforces that one has to wait at least one minute before reassigning a specific execution across _all_ instances of the orchestrator.

This is the loop which runs every minute:
https://github.com/kubeshop/testkube-cloud-api/blob/main/internal/agentserver/agent_testworkflows.go#L985-L998

Specifically, it triggers logic like this which could concurrently (across replicas of the orchestrator) try to reassign the executions to various runners and because of the loose logic (specifically previous runner id _or_ next runner id matching current runner id) around the assignment lock multiple executions could acquire the lock and fire off start requests to runners in a short window:
https://github.com/kubeshop/testkube-cloud-api/blob/main/internal/agentserver/agent_testworkflows.go#L949-L952

PR #6496, opened 2025-07-10
feat(TKC-3869): add health field mapping for TestWorkflow CRD

devcatalin:

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Changes

- added mapping for workflow health


dejanzele (approved):

vsukhin (approved):

dejanzele (approved):

api/v1/testkube.yaml
```diff
@@ -8630,8 +8630,6 @@ components:
           description: if test workflow is offline and cannot be executed
         status:
           $ref: "#/components/schemas/TestWorkflowStatusSummary"
-        health:
```
vsukhin, line 8633: in general we change api, if someone uses it, then it will not work

devcatalin: this one was not populated or used before this PR, it was just introduced before this one a couple weeks ago in the schema only

vsukhin: then it's okay to remove

PR #6497, opened 2025-07-11
refactor: change default to aborted from passed

ed382:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/testworkflows/executionworker/controller/notifier.go
+++ b/pkg/testworkflows/executionworker/controller/notifier.go
@@ -277,7 +277,7 @@ func (n *notifier) fillGaps(force bool) {
 	}
 
 	// Mark the initialization step as running
-	if n.state.PodCreated() && n.result.Initialization.NotStarted() {
+	if n.state.PodCreated() && n.result.Initialization.Status.NotStarted() {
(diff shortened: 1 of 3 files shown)


pkg/testworkflows/executionworker/controller/notifier.go
```diff
@@ -277,7 +277,7 @@ func (n *notifier) fillGaps(force bool) {
 	}
 
 	// Mark the initialization step as running
-	if n.state.PodCreated() && n.result.Initialization.NotStarted() {
+	if n.state.PodCreated() && n.result.Initialization.Status.NotStarted() {
```
vsukhin, line 280: Is it worth to check Status for nil?

ed382: It seems to be checked in those methods already:

```
func (s *TestWorkflowStepStatus) NotStarted() bool {
	return s != nil && (*s == "" || *s == QUEUED_TestWorkflowStepStatus)
}
```

vsukhin: ok

vsukhin (approved):
looks ok

PR #6522, opened 2025-07-22
chore: update golangci-lint version to v2; disable tests for contrib/ folder; redo test and lint ci

dejanzele:
## Pull request description 

Tidy Makefile and CI.
* Run integration tests only for tests ending in `_Integration`
* Remove installing legacty testing tools in integration tests CI
* Do not run tests for `contrib/` folder
* Optimize cache in CI
* Fix a security issue with CI_BOT_TOKEN

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vsukhin (approved):

.github/workflows/test.yaml
```diff
@@ -3,11 +3,16 @@ name: Code build and checks
...
   push:
     branches: [main]
-    paths-ignore: "docs/**"
+    paths-ignore: ["docs/**"]
   pull_request:
-    paths-ignore: "docs/**"
+    paths-ignore: ["docs/**"]
     branches: [main]
 
+permissions:
```
WitoDelnat, line 11: This runs on `ubuntu-latest`. Should we run it on Depot? Caching Go modules and build caches should speed up drastically if we do. This will likely increase our cache usage though and we are already over our plan's included usage so this will increase cost (which I'd argue is fine, but maybe sth to check).

vsukhin (approved):

PR #6523, opened 2025-07-22
feat: add targeting to workflow templates

WitoDelnat:
This step adds comments to better understand what the almighty Schedule does.


pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -278,6 +291,26 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
 	}
 
 	// Flatten target replicas
+	// ScheduleExecution has a `targets` parameter which determines __where__ to execute (i.e. which runner).
+	// - Each target has a selector `match` and anti-selector `not`.
+	//   - e.g. `target.match.name: ["a", "b"]` can by default be interpreted as executing on a runner with name "a" OR "b".
+	// - Each target has a `replicate` flag which changes behaviour from "OR" to "AND":
+	//   - e.g. `target.match.name: ["a", "b"]` with `replica: ["name"]` can by default be interpreted as executing on runners with name "a" AND "b".
+	//   - `replicate: ["name", "service"]` will do TO_BE_DETERMINED?
```
WitoDelnat: Does anyone know what will happen if you specify multiple `replicate`?

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -252,6 +252,19 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+	//
+	// Input:
+	// - List of originally scheduled executions.
+	//
+	// Output:
+	// - `intermediateSelectors`: List of scheduled executions, fanned-out across selector.labels.
+	//
+	// This block does the following:
+	// - It transforms _one_ scheduled execution with `labels` selector into _many_ scheduled execution with `name` selectors.
+	// - Unrelated, it also takes Custom Resources targets and puts it as a default if no targets are found. This probably should be moved elsewhere.
```
WitoDelnat: For the issue where we want to target through WorkflowTemplates:

It's only after this block that we know which test workflows to target. Within the next block we will start using said targets. That means that we should best extract this little unrelated part to happen in between; and then expand it's logic to also ApplyTemplates and see if we should use a target from there.

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -278,6 +291,26 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
 	}
 
 	// Flatten target replicas
+	// ScheduleExecution has a `targets` parameter which determines __where__ to execute (i.e. which runner).
+	// - Each target has a selector `match` and anti-selector `not`.
+	//   - e.g. `target.match.name: ["a", "b"]` can by default be interpreted as executing on a runner with name "a" OR "b".
+	// - Each target has a `replicate` flag which changes behaviour from "OR" to "AND":
+	//   - e.g. `target.match.name: ["a", "b"]` with `replica: ["name"]` can by default be interpreted as executing on runners with name "a" AND "b".
```
admacleod: I'm not sure this is quite worded correctly. `replica` causes a fan out based on the indicated key. So I think the logic is reversed? Without `replica` set then it will only match on runners having all of `name=a && name=b`, when `replica` is set to `name` then it will match on any runner having `name=a || name=b`.

WitoDelnat: You are right. It was not worded right. I updated the comments

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -552,6 +656,15 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
 }
 
+func getTemplateNames(w *testkube.TestWorkflow) []string {
+	if w.Spec.Use == nil {
+		return []string{}
+	}
+	return lo.Map(w.Spec.Use, func(t testkube.TestWorkflowTemplateRef, _ int) string {
+		return t.Name
+	})
+}
```
admacleod: I really dislike libraries like `lo` its just JavaScript-style reaching for an external library that hides the implementation, which here is fairly trivial:
```suggestion
func getTemplateNames(w *testkube.TestWorkflow) []string {
	var ret []string
	if w.Spec.Use == nil {
		return ret
	}
	for _, tmpl := range w.Spec.Use {
		ret = append(ret, tmpl.Name)
	}
	return ret
}
```
I find in Go it is better to write out simple implementations like this in standard Go rather than hide it behind an external function like `Map` as it aids readability for a reader who is familiar with Go.

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +273,104 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+		if err != nil {
+			close(ch)
+			return ch, err
+		}
+
+		w, ok := lo.First(workflows)
+		if !ok {
+			close(ch)
+			return ch, err
+		}
```
admacleod: I think using `lo` here has caused an issue here in that it is both challenging to follow the logic flow, and you do not have an appropriate `err` to emit here (should be `nil` if all other error checks are working as expected).
```suggestion
		if len(workflows) == 0 {
			close(ch)
			return ch, errors.New("no workflows for target") // Maybe this should just be a straight `continue`?
		}
		// Should be safe to access because of the length check above.
		w := workflows[0]
```

pkg/testworkflows/testworkflowexecutor/scheduler.go
WitoDelnat: I do not fully agree, but willing to removed it for now 👍 I've removed Lo in my latest commit

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +272,107 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+		// Check the workflow for targets
+		workflows, err := testWorkflows.Get(exec.Selector)
+		if err != nil {
+			close(ch)
+			return ch, err
+		}
+
+		if len(workflows) == 0 {
+			close(ch)
+			return ch, errors.New("workflow not found")
```
admacleod, line 309: I think we should log and continue here, we've successfully selected no workflows to be targeted so its not really an error that we can't handle, instead we can just ignore and move on.

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +272,107 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+			target := commonmapper.MapTargetApiToGrpc(w.Spec.Execution.Target)
+			exec.Targets = []*cloud.ExecutionTarget{target}
+			continue
+		}
+
+		// Check the templates for targets
+		var target *testkube.ExecutionTarget
+		var targetTemplateName string
+		templateNames := getTemplateNames(w)
+		for _, templateName := range templateNames {
```
admacleod, line 322: Looking at your actual usage here I don't even think there is much value in the separate function, especially as it is valid to `range` over a `nil` slice (produces no iterations). Instead you can just simply:
```suggestion
		if w.Spec == nil {
			continue
		}
		for _, template := range w.Spec.Use {
			templateName := template.Name
```

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +272,107 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+			t, err := testWorkflowTemplates.Get(templateName)
+			if err != nil {
+				close(ch)
+				return ch, err
```
admacleod, line 326: I think this is fine but it always gives me pause when theres an early return from a nested loop, could we just log and `continue` here instead?

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +272,107 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+			var templateTarget *testkube.ExecutionTarget
+			if t.Spec != nil && t.Spec.Execution != nil {
+				templateTarget = t.Spec.Execution.Target
+			}
+
+			if templateTarget == nil {
+				continue
+			}
+
+			if target != nil {
```
admacleod, line 338: Am I correct in reading that the idea is that this `for` will only ever execute at most once? I know that is briefly mentioned at the top of the block but I think it would be good to have a follow-up comment here clarifying that is the intended behaviour and why. As I was reading this section I was wondering if this was a mistake or intentional so its worth making it super clear to future readers.

pkg/testworkflows/testworkflowexecutor/scheduler.go
```diff
@@ -260,24 +272,107 @@ func (s *scheduler) Schedule(ctx context.Context, sensitiveDataHandler Sensitive
...
+				close(ch)
+				return ch, DuplicateTargetError{
+					Template1: templateName,
+					Template2: targetTemplateName,
+				}
```
admacleod, line 343: Just a suggestion here but what could be nice for debugging or helping end users might be to continue to loop but create some `returnErr` outside of the loop and then here:
```
returnErr = errors.Join(returnErr, DuplicateTargetError{/* values */})
```
then at the end of the loop do an:
```
if returnErr != nil
```
and close the channel and return from there. Yes potentially we waste some cpu cycles but it allows for _all_ failing cases to be caught in one run and communicated, rather than having to fix one, then failing again, then fix another, etc.

admacleod (approved):
Just a few comments but none of them blocking, they are more thoughts for possible improvements/simplifications but I'm happy to merge this as it currently is.

PR #6529, opened 2025-07-23
feat: add new workflow statuses

admacleod:
Add new workflow statuses to represent interstitial states of an Execution.


api/v1/testkube.yaml
```diff
@@ -9211,10 +9211,15 @@ components:
       type: string
       enum:
         - queued
+        - assigned
```
WitoDelnat, line 9214: I'm wondering about the ramifications of this within clients. Those are a lot of extra states to consider. While we should have detailed states, I wonder which ones can be grouped together (perhaps initially simply through utility functions). Also, do we store timestamps for all of these state transitions?

PR #6558, opened 2025-08-04
feat(TKC-4033): Testkube MCP server and new CLI command to serve it 

devcatalin:
This PR adds MCP integration under the new `pkg/mcp` package.

It introduces a new `testkube mcp serve` command in the CLI.
The server currently supports workflows, executions, and artifact operations.

The MCP server uses a tools package that defines handlers following a consistent structure, for example:
```go
func FetchExecutionLogs(client ExecutionLogger) (tool mcp.Tool, handler server.ToolHandlerFunc)
```

~~Each function receives a `getClient` function (used to retrieve the Testkube client for backend data access) and returns an `mcp.Tool` and a `ToolHandlerFunc` (from the [mcp-go library](https://github.com/mark3labs/mcp-go)).~~

Each function receives a small, focused interface (e.g., `ExecutionLogger`, `ArtifactLister`) and returns an `mcp.Tool` and a `ToolHandlerFunc` (from the [mcp-go library](https://github.com/mark3labs/mcp-go)).

This design is inspired by [GitHub's MCP implementation](https://github.com/github/github-mcp-server), which also uses the mcp-go library.

OAuth is strictly enforced. The MCP server only starts if the user is properly authenticated.

~~The default implemented TestkubeClient for the MCP uses HTTP to call API endpoints, but this approach will allow us to provide a custom implementation of the client in the control plane, to use direct calls to internal repositories.~~
The default implemented APIClient for the MCP uses HTTP to call API endpoints, but the interface-based design allows us to provide custom implementations in the control plane using direct calls to internal repositories.
It will allow us to expose a dedicated endpoint in the control plane for the MCP.


pkg/mcp/server.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	// Dashboard tools
+	urlTool, urlHandler := tools.BuildDashboardUrl(cfg.DashboardUrl, cfg.OrgId, cfg.EnvId)
+	mcpServer.AddTool(urlTool, urlHandler)
+
+	// Workflow tools
+	listWorkflowsTool, listWorkflowsHandler := tools.ListWorkflows(getClient)
+	mcpServer.AddTool(listWorkflowsTool, listWorkflowsHandler)
```
admacleod: You can simplify these calls if you aren't intending to reuse these variables anywhere:
```suggestion
	mcpServer.AddTool(tools.ListWorkflows(getClient))
```
A working example https://go.dev/play/p/Xjh6Jd9Qchb

pkg/mcp/tools/artifacts.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+func ListArtifacts(getClient GetClientFn) (tool mcp.Tool, handler server.ToolHandlerFunc) {
```
admacleod: Why take a dynamic client factory function here rather than an `interface` like:
```
type ArtifactLister interface {
    ListArtifacts(ctx context.Context, workflowName, executionId string) (string, error)
}
```
We spoke briefly about large vs small interfaces, but I think here in this implementation you want to even go beyond that. Each of these tools requires different specific functions from a client, that means each should be defining their own `interface` to receive a client that implements the required functionality.

What you could then do in `pkg/mcp` is rather than define the `customGetClient tools.GetClientFn` instead create a composed interface like:
```
type client interface {
    tools.ArtifactLister
    tools.ArtifactReader
    // all the other tool interfaces for the tools that are being added to the server
}
```

devcatalin: I was thinking maybe we need to build the tool using the context but didn't think it could even be passed to the actual methods 🤔 I'll try some making some changes to see how it would work

devcatalin: I refactored the code based on your suggestions, I like how it turned out, let me know what you think!

admacleod: Yeah this looks great! The composability of interfaces is something I often forget about but I think this was a great use of them.

pkg/mcp/README.md
```diff
@@ -0,0 +1,62 @@
+# Testkube MCP (Model Context Protocol) Integration
+
+This package provides MCP integration for Testkube, enabling AI assistants to interact with Testkube workflows, executions, and artifacts through the Model Context Protocol.
```
olensmar: add link to MCP spec?!

pkg/mcp/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Overview
+
+The MCP server is exposed via the `testkube mcp serve` CLI command, which leverages the existing OAuth authentication flow, for security purposes.
```
olensmar: perhaps in addition to mentioning OAuth (which not all users know what it is) also  add a note on how the mcp server requires you to log in so it can use the correct TK account / permissions!?

pkg/mcp/README.md
```diff
@@ -0,0 +1,62 @@
...
+### Client Abstraction
+
+The package uses an interface-based client design that supports multiple implementations:
+
+- **HTTP Client** (default): Makes REST API calls to Testkube endpoints
+- **Direct Repository Access** (future): For control plane integration with direct database access
+
+This flexibility allows the same MCP tools to work in different deployment scenarios without code changes.
+
+## Usage
```
olensmar, line 42: suggest to move the Usage section before the Archicture - and make it an H2 instead of H3 !?

pkg/mcp/README.md
```diff
@@ -0,0 +1,62 @@
...
+# Start the MCP server
+testkube mcp serve
+
+# Start with debug output
+testkube mcp serve --debug
+
+# Use --verbose if you need to check what context is used, but this will log things to stdout
+testkube mcp serve --verbose
+```
+
```
olensmar, line 61: can we add some sample prompts?

PR #6559, opened 2025-08-05
fix: atomic updates

WitoDelnat:
This PR contributes to https://linear.app/kubeshop/issue/TKC-4023/scheduler-updateresult-is-called-often-and-unsafe

### DIFF
--- a/pkg/repository/config/mongo_test.go
+++ b/pkg/repository/config/mongo_test.go
@@ -7,16 +7,11 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 
-	"github.com/kubeshop/testkube/internal/config"
 	"github.com/kubeshop/testkube/pkg/utils/test"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/kubeshop/testkube/pkg/api/v1/testkube"
-)
-
-var (
-	cfg, _ = config.Get()
 )
 
 func getRepository(dbName string) (*MongoRepository, error) {


pkg/repository/testworkflow/mongo/mongo.go
```diff
@@ -28,6 +28,10 @@ const (
 	configParamSizeLimit = 100
 )
 
+var (
+	ErrUnmodified = errors.New("record was unmodified")
```
admacleod, line 32: Is this really an error state? I would usually go for a different pattern to indicate that nothing was modified (if such a state is not really an error), either using a `updated bool` or `modified int` type of return value.

WitoDelnat: Fair enough. So interesting is that MongoDB makes even further differentiation with`matchedCount` and `modifiedCount` which surprisingly can differ if the modification would lead to a no-op (e.g. it was already in this status).

Do we want to be able to differentiate between this in clients? For example, if nothing got updated I see it as success but perhaps side-effects should not happen (e.g. no additional event logged). If nothing got matched its a not found error (should likely throw forbidden error).

admacleod: Good point, I think here we want perhaps a `bool` for "something was updated" and an `error` if not found. Then callers can handle:
```
updated, err := UpdateResultStrict()
switch {
case errors.Is(err, mongo.ErrNoDocuments):
    // Nothing was found
case err != nil:
    // Some other error
case !updated:
    // Nothing changed
}
// Successfully changed
```
(or invert the logic on that last check as appropriate).

admacleod (approved):

PR #6566, opened 2025-08-06
feat: improve debugging of MCP server

devcatalin:
Added flexible debug mode support to Testkube MCP server. 

This implementation introduces a `DebugInfo` type alias (map[string]any) that allows any client type (HTTP, file, database, mock, gRPC, etc.) to provide their own debug information structure. 

The core pattern uses a `DebugInfoProvider` interface with a `GetLastDebugInfo()` method, and `CreateToolResultWithDebug` helper function that automatically includes debug information as additional MCP tool content when available. 

For HTTP clients, we've implemented a `DebugAPIClient` wrapper with `DebugRoundTripper` that captures request/response details including URL, method, headers (with sensitive data redacted), timing, and response previews.

### DIFF
--- a/cmd/kubectl-testkube/commands/mcp.go
+++ b/cmd/kubectl-testkube/commands/mcp.go
@@ -28,6 +28,7 @@ to determine the organization and environment to connect to.`,
 
 func NewMcpServeCmd() *cobra.Command {
 	var mcpBaseURL string
+	var debugMode bool
 
 	cmd := &cobra.Command{
 		Use:   "serve",
@@ -105,7 +106,7 @@ over stdio. Use --verbose to see detailed output during startup.`,
 			}
 
 			// Start the MCP server
-			if err := startMCPServer(accessToken, cfg.CloudContext.OrganizationId, cfg.CloudContext.EnvironmentId, cfg.CloudContext.ApiUri, cfg.CloudContext.UiUri); err != nil {
+			if err := startMCPServer(accessToken, cfg.CloudContext.OrganizationId, cfg.CloudContext.EnvironmentId, cfg.CloudContext.ApiUri, cfg.CloudContext.UiUri, debugMode); err != nil {
 				if ui.IsVerbose() {
 					ui.Failf("Failed to start MCP server: %v", err)
 				}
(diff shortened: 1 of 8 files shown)


admacleod (commented):
Looking at the docs for the `mcp` package I believe there is a cleaner way to achieve this without updating all of your tool implementations by using a middleware on the server:
```

	// If no client is provided, use the default API client
	if client == nil {
		httpClient := &http.Client{}
		if cfg.Debug {
			client = NewDebugAPIClient(&cfg, httpClient)
		} else {
			client = NewAPIClient(&cfg, httpClient)
		}
	}

	mcpServer := server.NewMCPServer(
		"testkube-mcp",
		cfg.Version,
		server.WithToolCapabilities(true),
		server.WithToolHandlerMiddleware(func(next server.ToolHandlerFunc) server.ToolHandlerFunc {
			return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
				result, err := next(ctx, request)
				if debugClient, ok := client.(tools.DebugInfoProvider); ok {
					if debugInfo := debugClient.GetLastDebugInfo(); debugInfo != nil {
						debugJSON, marshalErr := json.MarshalIndent(debugInfo, "", "  ")
						if marshalErr == nil {
							debugContent := mcp.TextContent{
								Type: "text",
								Text: fmt.Sprintf("\n--- DEBUG INFO ---\n%s\n--- END DEBUG ---", string(debugJSON)),
							}
							result.Content = append(result.Content, debugContent)
						}
					}
				}
				return result, err
			}
		}),
	)
```

But I must warn that this implementation does not look like it is concurrency safe, it seems very likely that on a busy server you would end up with mismatched debug information, however if we are only looking at having this run locally, and even then rarely with debug mode enabled, then it may not ever come up.

admacleod:
Having now read through the docs for `mcp` I think you should also change the signature of your tool functions to return an `mcp.ServerTool` type, that would allow you to further tidy up adding tools to the server with `server.AddTools()` like this:
```
	mcpServer.AddTools(
		// Dashboard tools
		tools.BuildDashboardUrl(cfg.DashboardUrl, cfg.OrgId, cfg.EnvId),
		// Workflow tools
		tools.ListWorkflows(client),
		tools.GetWorkflow(client),
		tools.GetWorkflowDefinition(client),
		tools.CreateWorkflow(client),
		tools.RunWorkflow(client),
		// Execution tools
		tools.FetchExecutionLogs(client),
		tools.ListExecutions(client),
		tools.LookupExecutionId(client),
		tools.GetExecutionInfo(client),
		// Artifact tools
		tools.ListArtifacts(client),
		tools.ReadArtifact(client),
	)
```

devcatalin:
Thanks @admacleod for the feedback!
I tried a different approach for this and created a new PR with it as I started from scratch.
Closing this in favor of: https://github.com/kubeshop/testkube/pull/6569

PR #6569, opened 2025-08-06
feat: MCP server debug middleware

devcatalin:
This implementation introduces a context-based debug middleware that automatically collects and includes debug information in MCP tool responses when debug mode is enabled.

### DIFF
--- a/pkg/mcp/api.go
+++ b/pkg/mcp/api.go
@@ -49,8 +49,18 @@ func (c *APIClient) buildURL(path string, pathParams map[string]string) string {
 }
 
 func (c *APIClient) makeRequest(ctx context.Context, apiReq APIRequest) (string, error) {
+	// Check if debug collection is active
+	debugInfo := GetDebugInfo(ctx)
+	if debugInfo != nil {
+		debugInfo.Source = "http"
+		debugInfo.Data["method"] = apiReq.Method
+	}
+
 	// Build the URL
 	fullURL := c.buildURL(apiReq.Path, apiReq.PathParams)
+	if debugInfo != nil {
+		debugInfo.Data["url"] = fullURL
+	}
 
 	// Add query parameters if present
 	if len(apiReq.QueryParams) > 0 {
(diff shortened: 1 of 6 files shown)


pkg/mcp/api.go
```diff
@@ -131,9 +146,42 @@ func (c *APIClient) makeRequest(ctx context.Context, apiReq APIRequest) (string,
...
+		lowerKey := strings.ToLower(key)
+		if sensitiveHeaders[lowerKey] {
+			redacted[key] = []string{"[REDACTED]"}
+		} else {
+			redacted[key] = values
+		}
+	}
+
+	return redacted
+}
```
admacleod: You can do this in a slightly more efficient way by using an empty `struct` rather than a `bool` for a lookup `map` and moving the reference to a `var` outside the function, its a minor tweak and somewhat unnecessary but its a useful pattern to know:
```suggestion
var sensitiveHeaders = map[string]struct{}{
	"authorization": {},
	"cookie":        {},
	"x-api-key":     {},
	"x-auth-token":  {},
}

func redactSensitiveHeaders(headers http.Header) map[string][]string {
	if headers == nil {
		return nil
	}

	redacted := make(map[string][]string)

	for key, values := range headers {
		lowerKey := strings.ToLower(key)
		if _, found := sensitiveHeaders[lowerKey]; found {
			redacted[key] = []string{"[REDACTED]"}
		} else {
			redacted[key] = values
		}
	}

	return redacted
}
```

Another potential option that is less CPU efficient (which doesn't really matter but important to understand the tradeoffs) but much more readable with the new `slices` package is to use a string slice for the headers:
```suggestion
var sensitiveHeaders = []string{
	"authorization",
	"cookie",
	"x-api-key",
	"x-auth-token",
}

func redactSensitiveHeaders(headers http.Header) map[string][]string {
	if headers == nil {
		return nil
	}

	redacted := make(map[string][]string)

	for key, values := range headers {
		lowerKey := strings.ToLower(key)
		if slices.Contains(sensitiveHeaders, lowerKey) {
			redacted[key] = []string{"[REDACTED]"}
		} else {
			redacted[key] = values
		}
	}

	return redacted
}
```

admacleod (approved):
@devcatalin this is great! A much simpler implementation which avoids touching implementations. Only one suggestion but its more a style/different available options rather than required change.

@dejanzele the pattern that Catalin is using here is pretty much the one we were discussing the other day about storing log information in a context rather than an entire logger.

pkg/mcp/api.go
devcatalin: Thanks 🙏 I went with the second approach as it reads nicely

PR #6578, opened 2025-08-08
feat: add listener-agent and gitops mode

povilasv:
## Pull request description 

Adds new toggles to disable features such as apiServer, eventListener etc

Makes mode configure testkube in certain way.

Let me know what you think about this approach

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -93,6 +93,12 @@ func main() {
 
 	log.DefaultLogger.Infow("configuration loaded",
 		"mode", func() string {
+			if cfg.ListenerAgent {
+				return "listener-agent"
+			}
+			if cfg.GitOpsAgent {
+				return "gitops-agent"
+			}
 			if cfg.TestkubeProAPIKey != "" || cfg.TestkubeProAgentRegToken != "" {
 				return "agent"
 			}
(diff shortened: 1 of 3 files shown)


internal/common/mode.go
```diff
@@ -3,8 +3,10 @@ package common
 import "errors"
 
 const (
-	ModeStandalone = "standalone"
-	ModeAgent      = "agent"
+	ModeStandalone    = "standalone"
+	ModeAgent         = "agent"
+	ModeListenerAgent = "listener-agent"
+	ModeGitOpsAgent   = "gitops-agent"
```
WitoDelnat, line 9: I would intuitively also expect a `runner-agent` here.

internal/common/mode.go
WitoDelnat, line 1: More general thoughts around the mode approach. AFAIK, a deployment can only have one mode so we are pretty much cementing here that it runs either as A or as B, but never both. I know Dawid was a big advocate for this topology, but I still see a benefit in having a single server with simply multiple modules that can be enabled.

The reason is that I'm quite confident that this will make deployment and operations work easier for the majority of customers who have simple setups but want all functionality (e.g. run tests in one or two clusters).

A second benefit is that we can add some implicit coupling and labelling on it. A customer recently asked to make a deep coupling between a listener and runner, and we have to resort to tricks to map them together through externally provided metadata.

A third benefit is seemless transition for the superagent. It never has to disappear. It's simply an agent with all 3 modules enabled and can keep existing as it is today.

What are your thoughts around this?

cmd/api-server/main.go
```diff
@@ -105,7 +111,31 @@ func main() {
...
+		cfg.DisableDefaultAgent = true
+		cfg.DisableWebhooks = true
+		cfg.DisableDeprecatedTests = true
+		cfg.DisableRunner = true
+		cfg.EnableK8sControllers = false
+		cfg.EnableCronJobs = "false"
+		cfg.DisableAPIServer = true
+		cfg.DisableEventSystem = true
+		cfg.DisableTestTriggers = true
+	} else if cfg.TestkubeProAPIKey != "" || cfg.TestkubeProAgentRegToken != "" {
```
admacleod, line 138: It can be difficult to work out the differences between these two because there's just a sea of booleans (also `EnableCronJobs` 👀 not a real boolean) and the configuration values usage is spread throughout the rest of the file so can have lots of different effects that may not be immediately obvious.

It would be a significant refactor but if each of the startup sections were in their own function then you could have a quite slim `main` function that dives into other functions that are specific for each "mode", something like:

```
func main() {
    // Initial setup, parse config etc.

    switch mode {
    case common.ModeListenerAgent:
        startListenerAgent(/* args */)
    case common.GitOpsAgent:
        startGitOpsAgent(/* args */)
    case common.ModeAgent:
        startSuperAgent(/* args */)
    default:
        startFoo(/* args */)
    }

    // Other common setup code, signal processing etc.
}

func startListenerAgent(/* args */) {
    startAPIServer(/* args */)
    startK8sEventForwarder(/* args */)
    // etc
}
```
that might make it clearer what is happening for each type of agent and potentially make it easier to update or migrate agent types into their own binaries in the future.

internal/common/mode.go
WitoDelnat: hmm, perhaps I misunderstand but you could still run in an `agent` mode and manually configure things? This is just a "preset" of configurations? Can you confirm as it completely invalidates my previous comment.

cmd/api-server/main.go
admacleod: Ignore this, I think we actually want to go a slightly different direction with configurability to make each module configurable and possibly move "modes" out into deployment manifests eventually. Although an approach similar to this could be used:
```
func main() {
    // Initial setup, parse config etc.

    if apiServerEnabled {
        startAPIServer(/* args */)
    }
    if k8sEventForwarderEnabled {
        startK8sEventForwarder(/* args */)
    }

    // Other common setup code, signal processing etc.
}
```

PR #6589, opened 2025-08-14
Rename the new ASSIGNED workflow status to PENDING

admacleod:
This name is more aligned with other new status names and this is the last opportunity to make this change without a significant breaking change.


admacleod:
> Why change?
> 
> For me assigned makes sense (the workflow is assigned to runner agent). While pending is unclear? What is pending?

The idea is to align it with the other new statuses `STARTING`, `PAUSING`, `RESUMING`, `STOPPING`.

`ASSIGNED` makes sense when viewed statically (which is why it was the original name), however in practice an execution in this state is awaiting pickup by its assigned runner, so more correctly it is `PENDING`.

It also adds some confusion when the state changes, if an `ASSIGNED` execution transitions to `STARTING` is it no longer assigned? The assigned-ness of an execution remains as long as it has a runner set, but the state may transition away from `ASSIGNED` so using a different term instead of `ASSIGNED` is preferable.

PR #6597, opened 2025-08-15
chore: Add logging to gRPC connection

admacleod:
By default logging will occur based on the default specified in the zap grpc package: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/zap#DefaultClientCodeToLevel Most things will log at Debug level with only really issues logging at a higher level, this is appropriate for out fault finding work without overloading a customer's log handling system.

### DIFF
--- a/pkg/agent/client/client.go
+++ b/pkg/agent/client/client.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"time"
 
+	grpczap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


pkg/agent/client/client.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 	"time"
 
+	grpczap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
```
ed382, line 11: I don't see this in the go.mod/sum - do we need to pull that in?

admacleod: No, we already had it but it was only being used over in the agent gRPC server https://github.com/kubeshop/testkube/blob/d96d51780017f8b7a82b27ef00bd5816082d02af/pkg/controlplane/server.go#L378-L396 we don't have this in the Control Plane or the client.

PR #6616, opened 2025-08-21
fix: Allow scheduling to be finished from multiple states

admacleod:
Runners should be able to finish an execution from some additional edge case states.

### DIFF
--- a/pkg/repository/testworkflow/mongo/mongo.go
+++ b/pkg/repository/testworkflow/mongo/mongo.go
@@ -492,10 +492,24 @@ func (r *MongoRepository) FinishResultStrict(ctx context.Context, id, runnerId s
 	res, err := r.Coll.UpdateOne(ctx, bson.M{"id": id,
 		"runnerid": runnerId,
 		"result.status": bson.M{"$in": bson.A{
+			// The expected statuses from which an execution may be finished:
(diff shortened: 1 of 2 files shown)


pkg/repository/testworkflow/mongo/mongo.go
```diff
@@ -492,10 +492,23 @@ func (r *MongoRepository) FinishResultStrict(ctx context.Context, id, runnerId s
 	res, err := r.Coll.UpdateOne(ctx, bson.M{"id": id,
 		"runnerid": runnerId,
 		"result.status": bson.M{"$in": bson.A{
+			// The expected statuses from which an execution may be finished:
+			// - Queued, before even being assigned to a runner it can be immediately finished.
+			// - Pending, once the execution has been assigned to a runner but has not yet been
```
vsukhin, line 498: Should we have pending const below?

vsukhin:
Postgres repo changes?

pkg/repository/testworkflow/mongo/mongo.go
admacleod: 😅 this is why writing comments is so useful!

admacleod:
> Postgres repo changes?

This function is not yet implemented in Postgres (simply returns an error).

vsukhin (approved):

PR #6617, opened 2025-08-22
feat: execution polling

admacleod:
Create a new protobuf definition for the new polling execution scheduler.

Updates the proto generation configuration to use up to date tool versions allowing use of the latest proto editions features.

Generation for older proto definitions is preserved using a separate configuration file and exclusion directives. These can be removed once we have migrated all proto definitions to the new structure and generation.

It also introduces a GitLab action to run `buf` commands when checking PRs. It will check for breaking changes, linting, and formatting. It does not validate generation output.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -26,6 +26,33 @@ jobs:
         with:
           definition-file: api/v1/testkube.yaml
 
+  lint-protobuf:
+    name: Lint Protocol Buffers
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Run buf checks
+        uses: bufbuild/buf-action@v1
+        with:
+          version: 1.56.0
+          # Prevent buf from attempting to create a PR comment.
+          pr_comment: false
+          # buf config files are in the ./proto directory.
+          input: proto
+          # This is required to avoid the breaking change check
+          # from failing due to the imports from the legacy
+          # proto definitions, it should still allow for valid
+          # breaking change detection of our own new protos.
+          exclude_imports: true
+          # Ignore the legacy protos from formatting, they are
+          # also excluded in the buf.yaml file for breaking and
+          # linting checks.
+          exclude_paths: |
+            proto/logs.proto
+            proto/service.proto
+
   lint-go:
     name: Lint Go
     runs-on: ubuntu-latest


proto/testkube/testworkflow/execution/v1/execution.proto
```diff
@@ -0,0 +1,81 @@
...
+  string execution_id = 1;
+  // transition_to is the state the execution should be transitioned into.
+  ExecutionState transition_to = 2;
+}
+
+// GetExecutionUpdatesResponse is the Control Plane response to a request for execution state
+// transitions that should be performed by the client.
+message GetExecutionUpdatesResponse {
+  // start may or may not be included and indicates an Execution that should be started.
+  // A maximum of one execution to start is expected for each call to the RPC, rapid calls
```
povilasv: Why only single execution? Maybe we could just send the whole batch of executions? if you have a single runner and a lot of executions, it would speed things up; agent can run them in parallel

proto/testkube/testworkflow/execution/v1/execution.proto
```diff
@@ -0,0 +1,81 @@
...
+
+// ExecutionStart contains fields used to initially start an Execution.
+message ExecutionStart {
+  // execution_id is the unique identifier of the execution to be started.
+  string execution_id = 1;
+  // environment_id is the unique identifier of the environment in which the execution should be started.
+  string environment_id = 2;
+  // execution_token is an authentication token that can be used to obtain any required
+  // additional information about an execution from the Control Plane.
+  optional string execution_token = 3;
```
povilasv: Shouldnt this have more data like the workflow to run? Execution labels; etc?

admacleod: Not currently, this is the same payload of data that is currently sent to an agent to trigger an execution.

proto/testkube/testworkflow/execution/v1/execution.proto
admacleod: Currently we only want to send a single execution at a time as the runner has no way to indicate how many executions it may be capable of receiving. For backwards compatibility in the future I could change this to `repeated` given that `repeated` implies `optional` as well, but we will still only send up to one execution for starting per request until we are able to communicate more intelligently with the Control Plane/scheduler.

admacleod: Changed to `repeated` and expanded the comment to explain.

proto/testkube/testworkflow/execution/v1/execution_start.proto
```diff
@@ -0,0 +1,14 @@
+edition = "2023";
+
+package testkube.testworkflow.execution.v1;
+
+// ExecutionStart contains fields used to initially start an Execution.
+message ExecutionStart {
```
povilasv, line 8: Shouldnt this contaiin all the execution data? What workflow to execute, what parameters, etc?

proto/testkube/testworkflow/execution/v1/get_execution_updates_response.proto
```diff
@@ -0,0 +1,30 @@
...
+  // up to one execution.
+  repeated ExecutionStart start = 1;
+  // update lists all Executions understood to be under the control of the client that should
+  // have their state transitioned. Subsequent calls to the RPC will yield duplicate updates
+  // if the Execution state transition successes have not been communicated to the Control
+  // Plane, however new state transition directives may be added in the interim so it is not
+  // expected that a client would wait to successfully transition all Executions before calling
+  // the RPC again, instead clients should interpret duplicate state transitions in a sensible
+  // manner to prevent overloading Execution state transition behaviour.
+  repeated ExecutionStateTransition update = 2;
```
povilasv, line 29: How big can this get?  Could we put some sensible limit?

proto/testkube/testworkflow/execution/v1/execution_start.proto
admacleod: No, as mentioned above this message currently is a copy of the message `RunnerRequest` (from https://github.com/kubeshop/testkube/blob/admacleod/feat/execution-polling/proto/service.proto#L375) adapted to be more future proof to adding additional information in the future if we wish to. That information is currently not required and so is not included in the new implementation either.

proto/testkube/testworkflow/execution/v1/get_execution_updates_response.proto
admacleod: This should never grow to be larger than the number of currently running executions on a client. It cannot be limited, that isn't how protobuf `repeated` fields works https://protobuf.dev/programming-guides/proto3/#field-labels

proto/testkube/testworkflow/execution/v1/get_execution_updates_response.proto
```diff
@@ -0,0 +1,30 @@
...
+import "testkube/testworkflow/execution/v1/execution_start.proto";
+import "testkube/testworkflow/execution/v1/execution_state_transition.proto";
+
+// GetExecutionUpdatesResponse is the Control Plane response to a request for execution state
+// transitions that should be performed by the client.
+message GetExecutionUpdatesResponse {
+  // start may or may not be included and indicates an Execution that should be started.
+  // A maximum of one execution to start is expected for each call to the RPC, rapid calls
+  // to the RPC without informing the Control Plane that the Execution is successfully started
+  // are expected to yield the same Execution and so clients SHOULD NOT call the RPC again
```
WitoDelnat: > and so clients SHOULD NOT call the RPC again until the Execution has been successfully started

I'm not sure this upholds - unless we split ExecutionSateTransition events from ExecutionStart (perhaps we should?).

proto/testkube/testworkflow/execution/v1/runner_target.proto
```diff
@@ -0,0 +1,48 @@
...
+  //   Given a set of runners thus:
+  //   - name: "A"
+  //     region: "eu"
+  //   - name: "B"
+  //     region: "us"
+  //   - name: "C"
+  //     region: "eu"
+  //   If an execution specifies a replicate on the "region" label then:
+  //   Two separate executions will be scheduled, one targeting the label
+  //   'region: "eu"' and another targeting 'region: "us"'.
```
WitoDelnat, line 46: Actual... documentation? 🤩

proto/testkube/testworkflow/execution/v1/get_execution_updates_response.proto
admacleod: Actually I think this message is really in error. Rather than requesting clients not call the RPC again, instead we should just advise that they may receive the same execution on repeated calls and so clients should ensure that they debounce executions based on their unique execution ID.

admacleod: Updated the comment

proto/testkube/testworkflow/execution/v1/test_workflow_execution_service.proto
```diff
@@ -0,0 +1,19 @@
...
+
+// TestWorkflowExecutionService contains RPCs related to TestWorkflowExecutions.
+service TestWorkflowExecutionService {
+  // GetExecutionUpdates retrieves state transition imperatives from the Control Plane for
+  // Executions that are understood by the Control Plane to be under the control of the
+  // calling client.
+  rpc GetExecutionUpdates(GetExecutionUpdatesRequest) returns (GetExecutionUpdatesResponse);
+  // ScheduleExecution enables a client to request that the Control Plane queue an execution
+  // to be run.
+  rpc ScheduleExecution(ScheduleExecutionRequest) returns (ScheduleExecutionResponse);
```
admacleod: ### For discussion

This isn't strictly _scheduling_ an execution, should we take this opportunity to name it something more descriptive.

@WitoDelnat this is related to our conversation today about the correct naming of some of the functions.

As an idea I quite like `EnqueueExecution` as really that is all that is actually happening, the execution is being pre-processed and added to the queue for the scheduler.

olensmar: So `AddExecutionToSchedulerQueue` would be even more descriptive!? (although that results in rather lengthy request/response definitions.. )

admacleod: Yes that is perhaps a touch too verbose, and I'm not sure that the name will hold to be true for a long period of time. The trick is to get something in that sweet spot of clear enough to be descriptive and true, whilst vague enough to survive future refactors.

admacleod: I've dropped this commit from the PR, it is too much of a sidequest to include here. It can be picked up separately.

proto/testkube/testworkflow/execution/v1/execution_start.proto
povilasv: I am not sure I follow the argument? I understand this is copy-paste from previous method but we are introducing new logic for future. 

So basically runner will get execution id and then it will be required to fetch execution via get execution grpc call.

Why do we need that?

Can't it just get executions from CP immediatelly and execute them?

admacleod: Yes we could. I wanted to avoid doing that work as part of this change because currently the `GetExecutionResponse` message is just a raw unstructured message which I refuse to use here.  Correctly implementing that message was something I had planned to leave for later.

admacleod: I've added that, it was not quite as much work as I had expected because I looked at how we are using it in https://github.com/kubeshop/testkube/blob/9814e5b8ec6f028bb829f8f83f6ea367c550f027/pkg/runner/agent.go#L353-L371 and was able to drop a whole load of fields which means we do not need to pass an entire execution object. The only change is that we need to look up the workflow but that should be fine and is much better than trying to pass an entire workflow using a `byte` field or something ridiculous like that.

PR #6638, opened 2025-09-01
fix: legacy warning

olensmar:
## Pull request description 

Adds a warning message to all commands related to deprecated functionality: 

image

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/common/cloudcontext.go
+++ b/cmd/kubectl-testkube/commands/common/cloudcontext.go
@@ -51,13 +51,13 @@ func UiCloudContextValidationError(err error) {
 }
 
 func UiContextHeader(cmd *cobra.Command, cfg config.Data) {
-	// Check for legacy resource type and show deprecation warning
-	CheckLegacyName(cmd.Name())
-
 	// only show header when output is pretty
 	if cmd.Flag("output") != nil && cmd.Flag("output").Value.String() != "pretty" {
 		return
 	}
+
+	// Check for legacy resource type and show deprecation warning
+	CheckLegacyName(cmd.Name())
 
 	header := "\n"
 	separator := "   "


vsukhin (approved):
cool!

olensmar:
@vsukhin please have another look - i've moved things around.. 🙏

cmd/kubectl-testkube/commands/download.go
```diff
@@ -161,6 +161,12 @@ func NewDownloadTestSuiteArtifactsCmd() *cobra.Command {
 		Use:   "testsuite-artifacts <executionName>",
 		Short: "download test suite artifacts",
 		Args:  validator.ExecutionName,
+
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			// Check for legacy resource type and show deprecation warning
+			common.CheckLegacyName(cmd.Name())
+		},
+
```
vsukhin: Because NewDownloadTestSuiteArtifactsCmd is part of Download command at line 48, UiContextHeader should be already called

cmd/kubectl-testkube/commands/common/cloudcontext.go
```diff
@@ -51,6 +51,9 @@ func UiCloudContextValidationError(err error) {
 }
 
 func UiContextHeader(cmd *cobra.Command, cfg config.Data) {
+	// Check for legacy resource type and show deprecation warning
+	CheckLegacyName(cmd.Name())
+
```
vsukhin: Might be to call it afer bellow check. Becuase we check if output is pretty - `// only show header when output is pretty`

olensmar: hm.. i put it here deliberately - so everyone sees it.. but perhaps it could mess with how users automate this command!?

vsukhin: so it depends, if we want to see them in any case, then we can leave it there, but it can break their existing flows (if ui is not expected)

PR #6648, opened 2025-09-03
[TKC-4107][TKC-4111] workflow tests - granular workflow suites (standard suite, small tool specific suites), missing timeouts added

tkonieczny:


test/suites/cron/dev/prod-healthcheck-workflow-with-cron.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
 spec:
   system:
     pureByDefault: true
+  job:
+    activeDeadlineSeconds: 180
```
WitoDelnat, line 12: Cool! I'm keen that we rely on Kubernetes instead of making custom stuff. I wonder though if we properly integrate with this? Did you test this? Can you link a resulting execution from such deadline being exceeded?

PR #6670, opened 2025-09-12
feat: event labels

ed382:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/triggers/matcher.go
+++ b/pkg/triggers/matcher.go
@@ -63,8 +63,11 @@ func (s *Service) match(ctx context.Context, e *watcherEvent) error {
 		selectorMatched := matchSelector(t.Spec.Selector, e, s.logger)
 		resourceSelectorMatched := matchResourceSelector(&t.Spec.ResourceSelector, t.Namespace, e, s.logger)
 
-		if !(((!selectorSpecified && resourceSelectorSpecified) || selectorMatched) &&
-			((!resourceSelectorSpecified && selectorSpecified) || resourceSelectorMatched)) {
+		if (selectorSpecified || !resourceSelectorSpecified) && !selectorMatched {
+			continue
+		}
+
+		if (resourceSelectorSpecified || !selectorSpecified) && !resourceSelectorMatched {
 			continue
 		}
 


pkg/triggers/matcher.go
```diff
@@ -44,7 +47,11 @@ func (s *Service) match(ctx context.Context, e *watcherEvent) error {
 		if !matchEventOrCause(string(t.Spec.Event), e) {
 			continue
 		}
-		if !matchSelector(&t.Spec.ResourceSelector, t.Namespace, e, s.logger) {
+		if !matchSelector(t.Spec.Selector, e, s.logger) {
+			continue
+		}
+		if !matchResourceSelector(&t.Spec.ResourceSelector, t.Namespace, e, s.logger) {
+			s.logger.Infoln("skipped because of resource selector")
```
vsukhin: will it be poluting the logs?

pkg/triggers/matcher.go
```diff
@@ -120,24 +128,55 @@ func matchEventOrCause(targetEvent string, event *watcherEvent) bool {
...
 		return k8sSelector.Matches(resourceLabelSet)
 	}
 
+	if strings.TrimSpace(selector.Name) == "" &&
+		strings.TrimSpace(selector.NameRegex) == "" &&
+		strings.TrimSpace(selector.Namespace) == "" &&
+		strings.TrimSpace(selector.NamespaceRegex) == "" {
+		// If selectors unspecified, match all
+		return true
+	}
```
vsukhin: does it change below logic?
	isSameTestTriggerNamespace = selector.Namespace == "" && selector.NamespaceRegex == "" && namespace == event.Namespace
	return isSameName && (isSameNamespace || isSameTestTriggerNamespace)

pkg/triggers/matcher.go
```diff
@@ -120,24 +128,55 @@ func matchEventOrCause(targetEvent string, event *watcherEvent) bool {
...
+	if selector == nil {
+		return true
+	}
+	k8sSelector, err := v1.LabelSelectorAsSelector(selector)
+	if err != nil {
+		logger.Errorf("error creating k8s selector from label selector: %v", err)
+		return false
+	}
+	mergedLabels := make(map[string]string)
+	maps.Copy(mergedLabels, event.resourceLabels)
```
vsukhin, line 158: so, we validate resource labells in 2 methods, is it intended?

pkg/triggers/matcher.go
ed382: nice catch, this was added for debugging at one point

pkg/triggers/matcher.go
ed382: yes, this is merged set of labels which we call event labels - hopefully we can later deprecate the `resourceSelector` and just have a `selector` for events.

pkg/triggers/matcher.go
ed382: Yes, this changes the behavior somewhat - if you don't specify any of the `resourceSelector` fields before it would not match, but for this particular matcher it would match all events.

This was the behavior which was agreed for the new event based `selector`, but on second thought maybe it is problematic as one could accidentally trigger tests on all events.

Also I need to think about how to make it possible for people to use the `selector` without specifying the `resourceSelector` and vice versa while keeping this backward compatible. So I guess I need some more complicated logic like this for the combined match from these two selectors: `((!selectorSpecified && resourceSelectorSpecified) || selectorMatched) && ((!resourceSelectorSpecified && selectorSpecified) || resourceSelectorMatched)` 😵‍💫

pkg/triggers/matcher.go
vsukhin: ok, got it

pkg/triggers/matcher.go
vsukhin: great!

pkg/triggers/event.go
```diff
@@ -27,16 +28,17 @@ type conditionsGetterFn func() ([]testtriggersv1.TestTriggerCondition, error)
...
-	labels           map[string]string
+	resource         testtrigger.ResourceType
+	resourceLabels   map[string]string
 	objectMeta       metav1.Object
 	Object           any `json:"object"`
 	eventType        testtrigger.EventType
 	causes           []testtrigger.Cause
 	conditionsGetter conditionsGetterFn
 	addressGetter    addressGetterFn
+	eventLabels      map[string]string
```
povilasv: Could you make it public and add json tag? Then it will be available for gotemplate / jsonpath methods

ed382: yes, addressed

pkg/triggers/matcher.go
ed382: addressed

pkg/triggers/matcher.go
```diff
@@ -44,9 +47,30 @@ func (s *Service) match(ctx context.Context, e *watcherEvent) error {
...
+		// eventually deprecating the resourceSelector the logic below toggles
+		// the matching based on which selectors are specified in the resource.
+		selectorSpecified := t.Spec.Selector != nil &&
+			(len(t.Spec.Selector.MatchLabels) > 0 || len(t.Spec.Selector.MatchExpressions) > 0)
+		resourceSelectorSpecified := (t.Spec.ResourceSelector.LabelSelector != nil &&
+			(len(t.Spec.ResourceSelector.LabelSelector.MatchLabels) > 0 || len(t.Spec.ResourceSelector.LabelSelector.MatchExpressions) > 0)) ||
+			(strings.TrimSpace(t.Spec.ResourceSelector.Name) != "" &&
+				strings.TrimSpace(t.Spec.ResourceSelector.NameRegex) != "" &&
+				strings.TrimSpace(t.Spec.ResourceSelector.Namespace) != "" &&
+				strings.TrimSpace(t.Spec.ResourceSelector.NamespaceRegex) != "")
```
vsukhin, line 62: are they all 4 should be not empty?

vsukhin (approved):

pkg/triggers/matcher.go
ed382: Yes, we want to check if the `resourceSelector` is at all specified.

This is so that we can enable the scenario where only one of the `resourceSelector` or `selector` can be specified individually while keeping the behavior that no selectors should not match anything.

vsukhin: yes, but the orginal idea was, that we either specify name or regex, like
((strings.TrimSpace(t.Spec.ResourceSelector.Name) != "" ||
				strings.TrimSpace(t.Spec.ResourceSelector.NameRegex) != "") &&
				(strings.TrimSpace(t.Spec.ResourceSelector.Namespace) != "" ||
				strings.TrimSpace(t.Spec.ResourceSelector.NamespaceRegex) != ""))
And Namespace could be also taken from installation one, if I rememeber it right

ed382: This should not change that behavior since one would have to specify one of those resource selector fields as far as I understand. And if any of those fields are specified it behaves just like it would before with a logical `AND` with the new selector matcher.

The behavior only changes if all resource selector fields are not specified and some of the selector fields are specified.

PR #6684, opened 2025-09-18
feat: docker, shttp and api-key auth for mcp server

olensmar:
- Updated docker-bake.hcl to include a new target for mcp-server.
- Enhanced mcp server to support
  - running the MCP server in environment variable mode, allowing context configuration via environment variables.
  - authentication via API-Key instead of OAuth
  - shttp protocol instead of just stdio 
- Added checks for required environment variables and improved user feedback for authentication requirements.
- Fixed telemetry to 
  - send before running the mcp-serve command since it won't run until the server is stopped gracefully (which it isn't when running under docker and the docker container is stopped)
  - send events for individual tools
- Fixed mcp server logging to use json format so it can be correctly captured by logging tools

The below config in vs-codes mcp.json will use the latest version - the api-key needs to be generated in the dashboard with desired permissions.

```json
{
  "servers": {
    "testkube": {
      "command": "docker",
      "args": [
        "run", "--rm", "-i",
        "-e", "TK_ACCESS_TOKEN=tkcapi_xxx",
        "-e", "TK_ORG_ID=tkcorg_xxx",
        "-e", "TK_ENV_ID=tkcenv_xxx",
        "testkube/mcp-server:latest",
        "mcp", "serve"
      ],
      "type": "stdio"
    }
  }
}
```

Updated Docs PR is at https://testkube-docs-git-olensmar-docsmcp-docker-testkube-cloud.vercel.app/articles/mcp-docker

- [X] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/mcp/README.md
+++ b/pkg/mcp/README.md
@@ -4,7 +4,9 @@ This package provides MCP integration for Testkube, enabling AI assistants to in
 
 ## Overview
 
-The MCP server is exposed via the `testkube mcp serve` CLI command, which leverages the existing OAuth authentication flow for security purposes. The server requires you to be logged in via `testkube login` so it can access your specific Testkube account, permissions, and context (organization and environment). This provides seamless setup since the CLI already has all the necessary configuration after login.
+The MCP server is exposed via the `testkube mcp serve` CLI command, which leverages the existing OAuth authentication flow for security purposes. The server requires you to be logged in via `testkube login` so it can access your specific Testkube account, permissions, and context (organization and environment). This provides seamless setup since the CLI already has all the necessary configuration after login. API-Key authentication is also supported, use `testkube set context` to set the API-Key used for authentication.
+
+The MCP Server supports both `stdio` and `shttp` transports - see configuration examples below.
 
 This implementation uses the [mcp-go](https://github.com/mark3labs/mcp-go) library, chosen for its proven usage in other projects like [GitHub's MCP server](https://github.com/github/github-mcp-server). The tool design patterns and helper functions draw inspiration from GitHub's implementation while being adapted for Testkube's specific needs.
 
@@ -39,14 +41,22 @@ The package uses an interface-based client design that supports multiple impleme
 
 This flexibility allows the same MCP tools to work in different deployment scenarios without code changes.
 
+### Docker Image
+
+There is a Docker image available for the MCP-Server on DockerHub - https://hub.docker.com/repository/docker/testkube/mcp-server - see build and usage instructions at [/build/mcp-server/README.md](../../build/mcp-server/README.md).
+
 ## Usage
+
+See extensive docs at https://docs.testkube.io/articles/mcp-overview.
 
 ### Starting the MCP Server
 
 #### Stdio Transport (Default)
+
 ```bash
 # Build the CLI
 make build-kubectl-testkube-cli
+
 # This also deletes the previously built cli
 make rebuild-kubectl-testkube-cli
 
@@ -61,6 +71,7 @@ make rebuild-kubectl-testkube-cli
 ```
 
 #### Streamable HTTP (SHTTP) Transport
+
 ```bash
 # Start MCP server with SHTTP transport on localhost:8080
 ./bin/app/kubectl-testkube mcp serve --transport=shttp
@@ -88,6 +99,7 @@ The debug mode enables detailed request/response logging for the API client, mak
 ### Example MCP configuration for VSCode
 
 #### Stdio Configuration
+
 ```json
 {
   "servers": {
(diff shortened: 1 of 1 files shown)


pkg/telemetry/mcp.go
```diff
@@ -0,0 +1,57 @@
...
+		}},
+	}
+}
+
+// getStatusFromDuration determines status based on execution duration
+func getStatusFromDuration(duration time.Duration) string {
+	if duration < time.Second {
+		return "fast"
+	} else if duration < 5*time.Second {
+		return "normal"
```
vsukhin: better to use const durationFast = time.Second, durationNormal = 5*time.Second

olensmar: thanks - my bad - removing this altogether - status is something else and we already have the duration

pkg/mcp/server.go
```diff
@@ -97,3 +104,76 @@ func ServeStdioMCP(cfg MCPServerConfig, client Client) error {
...
+	}()
+
+	// Log server startup information
+	protocol := "http"
+	if cfg.SHTTPConfig.EnableTLS {
+		protocol = "https"
+	}
+
+	if ui.IsVerbose() {
+		fmt.Fprintf(os.Stderr, "Testkube MCP server starting on %s://%s\n", protocol, addr)
```
vsukhin, line 145: why is it printed to StdErr, if it's just info?

pkg/mcp/server.go
```diff
@@ -85,6 +87,11 @@ func ServeStdioMCP(cfg MCPServerConfig, client Client) error {
 		}
 	}()
 
+	if ui.IsVerbose() {
+		fmt.Fprintf(os.Stderr, "Testkube MCP server started in stdio mode\n")
+		fmt.Fprintf(os.Stderr, "Use 'testkube mcp serve --help' for more options\n")
```
vsukhin, line 92: why is it printed to StdErr, if it's just info?

pkg/mcp/server.go
olensmar: so it doesn't interfere with the JSON protocol used over stdio - perhaps there is a better way to do this?

vsukhin: Might be, printing them as json data {"info": ""}, but not sure

pkg/mcp/server.go
```diff
@@ -97,3 +104,76 @@ func ServeStdioMCP(cfg MCPServerConfig, client Client) error {
...
+	// Wait for either server error or interrupt signal
+	select {
+	case <-sigCh:
+		// Signal received, shutdown gracefully
+		if ui.IsVerbose() {
+			fmt.Fprintf(os.Stderr, "\nShutting down MCP server...\n")
+		}
+
+		// Graceful shutdown handled by the mcp-go library
+		// The server will stop when the context is cancelled
```
vsukhin, line 159: do you need to cancel context, like  <-ctx.Done() ?

olensmar: I dunno - we didn't have this for the stdio server either - @devcatalin any thoughts on this?

pkg/mcp/README.md
devcatalin, line 1: It would be helpful to have links between this readme and the build/mcp-server readme

pkg/mcp/server.go
devcatalin: We'll have to keep this on StdErr, otherwise it can easily break tools like the MCP inspector which helps us debug the MCP server

pkg/mcp/server.go
devcatalin: Maybe we would need to cancel it like this instead of `<- ctx.Done()` ?
I'm not sure if we need it though, @vsukhin what are your thoughts?

```go
        // Shutdown HTTP server gracefully with timeout
		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
		defer cancel()
		
		if err := httpServer.Shutdown(shutdownCtx); err != nil {
			return fmt.Errorf("failed to shutdown HTTP server: %v", err)
		}
        return nil
```

pkg/mcp/server.go
vsukhin: Then, it's fine, if it's current convention in mcp

pkg/mcp/server.go
vsukhin: Yes, probably, it's an optional thing for mcp server, it doesn't have critical data for graceful shutdown. I just saw comments and was curious what they are for ) Up to you

pkg/mcp/README.md
olensmar: good idea - will add 👍

PR #6688, opened 2025-09-22
feat: add support for --variable flag in test workflows

dejanzele:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


dejanzele:
https://linear.app/kubeshop/issue/TKC-4160/runtime-overwrite-of-config-parameters-required-parity-with-tests-to

cmd/api-server/main.go
```diff
@@ -87,7 +87,9 @@ func init() {
 
 func main() {
 	startTime := time.Now()
+	fmt.Println("AUTORELOAD TEST: API Server starting with --variable support")
```
WitoDelnat: Are these comments leftover or intended?

internal/app/api/v1/testworkflows.go
```diff
@@ -465,6 +465,7 @@ func (s *TestkubeAPI) ExecuteTestWorkflowHandler() fiber.Handler {
 		ctx := c.Context()
 		name := c.Params("id")
 		selector := c.Query("selector")
+		s.Log.Infow("ExecuteTestWorkflowHandler START", "name", name, "selector", selector)
```
WitoDelnat: Debug log!

internal/app/api/v1/testworkflows.go
```diff
@@ -620,6 +628,13 @@ func (s *TestkubeAPI) ReRunTestWorkflowExecutionHandler() fiber.Handler {
 
 		scheduleExecution.Selector = &cloud.ScheduleResourceSelector{Name: name}
 		scheduleExecution.Config = request.Config
+		// Create runtime configuration if variables are provided
+		if len(request.Variables) > 0 {
+			scheduleExecution.Runtime = &cloud.TestWorkflowRuntime{
+				EnvVars: request.Variables,
+			}
+			s.Log.Infow("API: Received variables in rerun execution request", "name", name, "variables", request.Variables)
```
WitoDelnat: Intended?

cmd/api-server/main.go
vsukhin: is this debug output?

cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -78,6 +78,7 @@ func (e executionError) Unwrap() error {
 type RunOptions struct {
 	ExecutionName            string
 	Config                   map[string]string
+	Variables                []string
```
vsukhin, line 78: shouldn't it be map?

cmd/kubectl-testkube/commands/testworkflows/run.go
```diff
@@ -253,9 +258,21 @@ func loadConfig() testkubecfg.Data {
 	return cfg
 }
 
+// parseVariables converts variable array to map
+func parseVariables(variables []string) map[string]string {
```
vsukhin: corba can do it

vsukhin (commented):
do we need to support it in PostgreSQL?

cmd/kubectl-testkube/commands/testworkflows/run.go
dejanzele: Depends what we like more:
1. With array we can do `--variable KEY1=value1 --variable KEY2=value` but we lose the automatic parsing of key=value
2. With map we can use only one flag like `--variable KEY1=value1,KEY2=value2`, but we get automatic parsing of key=value

I don't have a strong opinion, what do you think @vsukhin ?

vsukhin: Up to you, honestly, what do you prefer, both should be okay

cmd/api-server/main.go
dejanzele: debug, removed

internal/app/api/v1/testworkflows.go
dejanzele: no

PR #6704, opened 2025-09-24
fix: Reconnect gRPC streams individually

admacleod:
Rather than having streams killed if one of them
needs to reconnect, instead streams should be
individually retried and reconnected so that
a single failure does not kill all of them.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/agent/agent.go
```diff
@@ -108,49 +109,54 @@ func NewAgent(logger *zap.SugaredLogger,
 }
 
 func (ag *Agent) Run(ctx context.Context) error {
-	for {
-		if ctx.Err() != nil {
-			return ctx.Err()
-		}
-		err := ag.run(ctx)
+	reconnectionLoopDelay := 5 * time.Second
```
vsukhin: might be to move to const

vsukhin (approved):
nice improvement

PR #6706, opened 2025-09-24
feat: MCP wait_for_executions tool

olensmar:
## Pull request description 

Adds a "wait_for_executions" tool to the MCP server that waits for the specified executions to finish, helpful when launching multiple executions and wanting the agent/llm to wait 

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/mcp/api.go
+++ b/pkg/mcp/api.go
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/kubeshop/testkube/pkg/mcp/tools"
 )
@@ -542,4 +543,91 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
 			"workflowName": workflowName,
 		},
 	})
+}
+
+func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string, timeoutMinutes int) (string, error) {
+	// Create a context with timeout
+	if timeoutMinutes > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(timeoutMinutes)*time.Minute)
+		defer cancel()
+	}
+
+	// Polling loop
+	ticker := time.NewTicker(5 * time.Second) // Check every 5 seconds
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return "", fmt.Errorf("timeout waiting for executions: %w", ctx.Err())
+		case <-ticker.C:
+			allComplete, results, err := c.checkExecutionStatuses(ctx, executionIds)
+			if err != nil {
+				return "", fmt.Errorf("failed to check execution status: %w", err)
+			}
+
+			if allComplete {
+				return results, nil
+			}
+		}
+	}
+}
+
+func (c *APIClient) checkExecutionStatuses(ctx context.Context, executionIds []string) (bool, string, error) {
+	var allComplete = true
+	var results []map[string]interface{}
+
+	// Final status values that indicate execution has completed
+	finalStatuses := []string{"passed", "failed", "aborted", "timeout", "skipped", "canceled"}
+
+	for _, executionId := range executionIds {
+		status, err := c.getExecutionStatus(ctx, executionId)
+		if err != nil {
+			return false, "", err
+		}
+
+		results = append(results, map[string]interface{}{
+			"executionId": executionId,
+			"status":      status,
+		})
+
+		// Check if execution is in a final state
+		if !slices.Contains(finalStatuses, status) {
+			allComplete = false
+		}
+	}
+
+	resultJSON, _ := json.Marshal(results)
+	return allComplete, string(resultJSON), nil
+}
+
+func (c *APIClient) getExecutionStatus(ctx context.Context, executionId string) (string, error) {
+	// Use the test-workflow-executions endpoint which only requires execution ID
+	response, err := c.makeRequest(ctx, APIRequest{
+		Method: "GET",
+		Path:   "/agent/test-workflow-executions/{executionId}",
+		Scope:  ApiScopeOrgEnv,
+		PathParams: map[string]string{
+			"executionId": executionId,
+		},
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to get execution: %w", err)
+	}
+
+	// Parse the response to extract status
+	var execInfo map[string]interface{}
+	if err := json.Unmarshal([]byte(response), &execInfo); err != nil {
+		return "", fmt.Errorf("failed to parse execution response: %w", err)
+	}
+
+	// Check if there's a result object with status
+	if result, ok := execInfo["result"].(map[string]interface{}); ok {
+		if status, ok := result["status"].(string); ok {
+			return status, nil
+		}
+	}
+
+	return "", fmt.Errorf("status not found in execution response")
 }
--- a/pkg/mcp/api.go
+++ b/pkg/mcp/api.go
@@ -553,6 +553,10 @@ func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string
 		defer cancel()
 	}
 
+	// Track completed executions to avoid re-checking them
+	completedExecutions := make(map[string]bool)
+	allResults := make(map[string]map[string]interface{})
+
 	// Polling loop
 	ticker := time.NewTicker(5 * time.Second) // Check every 5 seconds
 	defer ticker.Stop()
@@ -562,7 +566,7 @@ func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string
 		case <-ctx.Done():
 			return "", fmt.Errorf("timeout waiting for executions: %w", ctx.Err())
 		case <-ticker.C:
-			allComplete, results, err := c.checkExecutionStatuses(ctx, executionIds)
+			allComplete, results, err := c.checkExecutionStatuses(ctx, executionIds, completedExecutions, allResults)
 			if err != nil {
 				return "", fmt.Errorf("failed to check execution status: %w", err)
 			}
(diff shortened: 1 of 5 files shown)


pkg/mcp/api.go
```diff
@@ -543,3 +544,113 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+			}
+		}
+	}
+}
+
+func (c *APIClient) checkExecutionStatuses(ctx context.Context, executionIds []string, completedExecutions map[string]bool, allResults map[string]map[string]interface{}) (bool, string, error) {
+	var allComplete = true
+
+	// Final status values that indicate execution has completed
+	finalStatuses := []string{"passed", "failed", "aborted", "timeout", "skipped", "canceled"}
```
WitoDelnat: These contain invalid execution states? 🤔 Please consider constants like `ABORTED_TestWorkflowStatus` instead of magic string values.

pkg/mcp/api.go
```diff
@@ -543,3 +544,113 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string, timeoutMinutes int) (string, error) {
+	// Create a context with timeout
+	if timeoutMinutes > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(timeoutMinutes)*time.Minute)
+		defer cancel()
+	}
+
+	// Track completed executions to avoid re-checking them
+	completedExecutions := make(map[string]bool)
```
WitoDelnat, line 551: This is confusing and hard to follow code. I believe you intend to pass around these structs by pointer to be able to modify them within other functions. The `checkExecutionStatuses` both has explicit results and it does these side-effects on the structs. Please consider a different approach.

~I also doubt whether this works, because this map might not be passed as a pointer. Was it tested thoroughly? You are likely refetching everything all the time.~

olensmar: thanks - will try to simplify - it seemed to work when I tried it out - but that could have been "by mistake"..

admacleod: `map` is passed by reference once it is initialised so it will still work. But generally try to avoid mutating data, or relying on mutated data, it makes it quite challenging to debug.

WitoDelnat: Thanks for the clarification Ali. It's why I was doubting. [This article](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it) indeed explains why!

pkg/mcp/api.go
```diff
@@ -543,3 +544,113 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
 	})
 }
+
+func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string, timeoutMinutes int) (string, error) {
+	// Create a context with timeout
+	if timeoutMinutes > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(timeoutMinutes)*time.Minute)
+		defer cancel()
+	}
```
admacleod: I would lift this out of the function and instead just create the context with a timeout in the calling function in `pkg/mcp/tools/executions.go`.

The reason is that if you already know the timeout, you should implement the context cancellation, rather than passing the timeout value _and_ an uninstrumented context into a function for it to handle.

pkg/mcp/api.go
```diff
@@ -543,3 +544,113 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+
+	// Track completed executions to avoid re-checking them
+	completedExecutions := make(map[string]bool)
+	allResults := make(map[string]map[string]interface{})
+
+	// Polling loop
+	ticker := time.NewTicker(5 * time.Second) // Check every 5 seconds
+	defer ticker.Stop()
+
```
admacleod, line 556: Since Go 1.23 if you only want to use the ticker channel there is a simpler way to write this:

```suggestion
	// Polling loop
	ticker := time.Tick(5 * time.Second) // Check every 5 seconds
```

and then use it by simply listening on the returned value (which is a channel):

```
case <-ticker:
```

there is no longer a requirement to call `Stop` as the garbage collector will now correctly clean up tickers.

pkg/mcp/api.go
```diff
@@ -543,3 +544,113 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+			}
+
+			if allComplete {
+				return results, nil
+			}
+		}
+	}
+}
+
+func (c *APIClient) checkExecutionStatuses(ctx context.Context, executionIds []string, completedExecutions map[string]bool, allResults map[string]map[string]interface{}) (bool, string, error) {
```
admacleod: I would inline this function, it will make it easier to reason about the loop and make it easier to debug the `map` mutations that occur inside here.

pkg/mcp/api.go
```diff
@@ -543,3 +545,118 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+	var allComplete = true
+
+	// Final status values that indicate execution has completed
+	finalStatuses := []string{
+		string(testkube.PASSED_TestWorkflowStatus),
+		string(testkube.FAILED_TestWorkflowStatus),
+		string(testkube.ABORTED_TestWorkflowStatus),
+		string(testkube.CANCELED_TestWorkflowStatus),
+	}
```
admacleod: For a bit of future proofing you could use `testkube.TestWorkflowTerminalStatus` instead, it is also an array that contains these status values, but it is more likely to be updated in the future if we add/change those statuses that represent a terminal state for a workflow execution.

olensmar: even better - thanks!

pkg/mcp/api.go
```diff
@@ -543,3 +545,88 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
 		},
 	})
 }
+
+func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string, timeoutMinutes int) (string, error) {
```
admacleod: You can clean up the signature of this function now, it will prevent any future confusion.

```suggestion
func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string) (string, error) {
```

pkg/mcp/api.go
```diff
@@ -543,3 +545,88 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+	// Track completed executions to avoid re-checking them
+	completedExecutions := make(map[string]bool)
```
admacleod, line 551: This is totally fine, but worth remembering that there is an alternative approach that uses less memory by using an empty struct as the map field when all you care about is presence in the map:

```suggestion
	completedExecutions := make(map[string]struct{})
```

It isn't important for this because I don't think we're expecting the map to grow significantly and I don't think the performance of an MCP tool is ever going to be a bottleneck, its just worth knowing about.

pkg/mcp/api.go
```diff
@@ -543,3 +545,88 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+	allResults := make(map[string]map[string]interface{})
```
admacleod, line 552: Minor stylistic, but it is more idiomatic to use `any` instead of the empty interface `interface{}`. `any` is just syntactic sugar but it makes it a bit clearer that you expect _something_ to populate this value.

Lower down I can see that actually the values in here are all `string` so you _could_ change the type to `string` to make it more concrete.

pkg/mcp/api.go
```diff
@@ -543,3 +545,88 @@ func (c *APIClient) GetWorkflowMetrics(ctx context.Context, workflowName string)
...
+						results = append(results, result)
+					}
+				}
+
+				resultJSON, _ := json.Marshal(results)
+				return string(resultJSON), nil
+			}
+		}
+	}
+}
```
admacleod: Now that this block is inlined it is easier to determine ways that it can be improved. This current loop of loops is ok but we could make it easier to read by leveraging Go's concurrency to "fan out" and wait for all executions to complete.

My recommendation would be to have an inline function that simply polls periodically and then apply that function for each execution that you want to watch. By running each of these functions in a group and waiting for them to return you can then have a much simpler check implementation.

```suggestion
func (c *APIClient) WaitForExecutions(ctx context.Context, executionIds []string) string {
	pollEvery := 5 * time.Second
	type result struct {
		ExecutionId string
		Status      string
		Error       error
	}
	waitForExecution := func(ctx context.Context, executionId string, resultChan chan<- result) func() {
		request := APIRequest{
			Method: "GET",
			Path:   "/agent/test-workflow-executions/{executionId}",
			Scope:  ApiScopeOrgEnv,
			PathParams: map[string]string{
				"executionId": executionId,
			},
		}

		return func() {
			ticker := time.Tick(pollEvery)
			for {
				select {
				case <-ctx.Done():
					resultChan <- result{
						ExecutionId: executionId,
						Error:       fmt.Errorf("context cancelled before completing execution check: %w", ctx.Err()),
					}
					return
				case <-ticker:
				}
				response, err := c.makeRequest(ctx, request)
				if err != nil {
					// Could be a temporary issue, keep looping.
					continue
				}

				// Parse the response to extract status
				var execInfo testkube.TestWorkflowExecution
				if err := json.Unmarshal([]byte(response), &execInfo); err != nil {
					// Can't parse the response, this could be a temporary issue, keep going.
					continue
				}

				// Extract status
				var status string
				if execInfo.Result != nil && execInfo.Result.Status != nil {
					status = string(*execInfo.Result.Status)
				}
				if status == "" {
					// Can't evaluate this response, keep looping in case this is an unexpected response value.
					continue
				}

				// Check if execution is in a final state and return if so.
				if slices.Contains(testkube.TestWorkflowTerminalStatus, testkube.TestWorkflowStatus(status)) {
					// Store the result
					resultChan <- result{
						ExecutionId: executionId,
						Status:      status,
					}
					return
				}
			}
		}
	}

	resultChan := make(chan result)

	var wg sync.WaitGroup

	for _, executionId := range executionIds {
		wg.Go(waitForExecution(ctx, executionId, resultChan))
	}

	// Once all workers are complete, close the result channel.
	go func() {
		wg.Wait()
		close(resultChan)
	}()

	// Read results until the end.
	var results []result
	for r := range resultChan {
		results = append(results, r)
	}

	resultJSON, _ := json.Marshal(results)
	return string(resultJSON)
}
```

Of course you may want to do error handling slightly differently, and possibly lift the inline function, type definition, and tick constant out into the global scope, but generally when you're trying to do networky things it is better to try to do them concurrently otherwise one bad network call can ruin everything.

olensmar: Thanks Ali - i think i'm gonna defer this improvement for now but I'll make a note of it in the code with a link to this suggestion 🙏

PR #6720, opened 2025-09-30
fix: crash after retry when control plane is unavailable

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/commons/commons.go
+++ b/cmd/api-server/commons/commons.go
@@ -332,7 +332,7 @@ func ReadDefaultExecutors(cfg *config.Config) (executors []testkube.ExecutorDeta
 	return next, images, nil
 }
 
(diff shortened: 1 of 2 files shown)


cmd/api-server/commons/commons.go
```diff
@@ -376,7 +376,7 @@ func ReadProContext(ctx context.Context, cfg *config.Config, grpcClient cloud.Te
 	foundProContext, err := grpcClient.GetProContext(ctx, &emptypb.Empty{})
 	if err != nil {
 		log.DefaultLogger.Warnf("cannot fetch pro-context from cloud: %s", err)
-		return proContext
+		return proContext, fmt.Errorf("cannot get pro context: %v", err)
```
WitoDelnat, line 379: Before we would just continue as if this were an OSS agent which lead to unexpected behaviour.

PR #6722, opened 2025-10-01
feat: add support for silent executions

dejanzele:
## Pull request description 

Add support for `--silent` flag when running TestWorkflow executions via the CLI `testkube run testworkflow`.

When the flag is provided, the execution result will not be saved in Insights, will not get counted towards Workflow Health metrics, will not trigger any webhooks. and CDEvents.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/event/kind/webhook/listener.go
```diff
@@ -189,7 +189,8 @@ func (l *WebhookListener) Notify(event testkube.Event) (result testkube.EventRes
 		l.Log.With(event.Log()...).Debug("webhook listener is disabled for test suite execution")
 		result = testkube.NewSuccessEventResult(event.Id, "webhook listener is disabled for test suite execution")
 		return
-	case event.TestWorkflowExecution != nil && event.TestWorkflowExecution.DisableWebhooks:
+	case event.TestWorkflowExecution != nil && (event.TestWorkflowExecution.DisableWebhooks ||
+		(event.TestWorkflowExecution.SilentMode != nil && event.TestWorkflowExecution.SilentMode.Webhooks)):
```
vsukhin, line 193: does this flag duplicate DisableWebhooks?

pkg/agent/client/client.go
```diff
@@ -129,17 +129,14 @@ func NewGRPCConnectionWithTracing(
 		return client, fmt.Errorf("create new grpc client: %w", err)
 	}
 
-	// Wait for connection to go ready.
```
vsukhin, line 180: Is this your changes?

pkg/event/kind/webhook/listener.go
dejanzele: to some degree, yes.
--silence does much more than just webhooks, but it is backwards compatible

PR #6727, opened 2025-10-02
feat: new scheduling RPC

admacleod:
This is a new RPC but mostly copies the InitExecution
RPC, albeit with better type safety because I'm
not lazy.

This is being added to facilitate a simpler
implementation of the GetExecutionUpdates
client, and to begin the migration of RPCs
to the new style.


proto/testkube/testworkflow/execution/v1/test_workflow_execution_service.proto
```diff
@@ -4,11 +4,18 @@ package testkube.testworkflow.execution.v1;
...
 service TestWorkflowExecutionService {
   // GetExecutionUpdates retrieves state transition imperatives from the Control Plane for
   // Executions that are understood by the Control Plane to be under the control of the
   // calling client.
   rpc GetExecutionUpdates(GetExecutionUpdatesRequest) returns (GetExecutionUpdatesResponse);
+  // SetExecutionScheduling informs the server that the client has transitioned an execution
+  // to a SCHEDULING state. This should be initiated by a client immediately after an execution
+  // has been successfully applied to Kubernetes in order to prevent the execution from being
+  // sent in subsequent calls to GetExecutionUpdates.
+  rpc SetExecutionScheduling(SetExecutionSchedulingRequest) returns (SetExecutionSchedulingResponse);
```
WitoDelnat, line 20: Does this mean we skip setting `Starting` and immediately inform the control plane of `Scheduling`?

admacleod: No, because `Starting` is set by the Control Plane to indicate that a runner has been selected and that the runner has been informed of the execution. Whereas `Scheduling` is set by the Runner to indicate that it has begun the process of spinning up the containers to execute the workflow.

The statuses indicate a "success" from either side of a network connection so we need to ensure that both are at least attempted.

Note that this behaviour already exists, I've just changed the naming here to be clearer with our new more accurate model of how everything works.

WitoDelnat: Right. Starting just means some runner has polled and received the event. Assigned & Starting happen at the "same" time and scheduling comes after. Thanks for the clarification!

PR #6731, opened 2025-10-03
feat: webhook syncing, also switched to maintained mockgen version

ed382:
## Pull request description 
depends on https://github.com/kubeshop/testkube-operator/pull/344

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


proto/service.proto
```diff
@@ -74,8 +73,16 @@ service TestKubeCloudAPI {
...
+    // -- Webhook
+    rpc GetWebhook(GetWebhookRequest) returns (GetWebhookResponse);
+    rpc ListWebhooks(ListWebhooksRequest) returns (stream WebhookListItem);
+    rpc ListWebhookLabels(ListWebhookLabelsRequest) returns (ListWebhookLabelsResponse);
+    rpc CreateWebhook(CreateWebhookRequest) returns (CreateWebhookResponse);
+    rpc UpdateWebhook(UpdateWebhookRequest) returns (UpdateWebhookResponse);
+    rpc DeleteWebhook(DeleteWebhookRequest) returns (DeleteWebhookResponse);
+    rpc DeleteAllWebhooks(DeleteAllWebhooksRequest) returns (DeleteAllWebhooksResponse);
+    rpc DeleteWebhooksByLabels(DeleteWebhooksByLabelsRequest) returns (DeleteWebhooksByLabelsResponse);
+    rpc WatchWebhookUpdates(WatchWebhookUpdatesRequest) returns (stream WebhookUpdate);
```
povilasv, line 85: Do we want to add another grpc streaming endpoint?

pkg/controlplaneclient/webhooks.go
```diff
@@ -0,0 +1,166 @@
...
+				Resource:  &resource,
+			})
+		}
+		if errors.Is(err, io.EOF) {
+			err = nil
+		}
+		watcher.Close(err)
+	}()
+	return watcher
+}
```
povilasv, line 166: This creates grpc streaming connection, right?
Do we want to do that? Or maybe instead just use polling?

proto/service.proto
ed382: Yes, but this is how the other CRs are synced at the moment - removing streaming RPCs is not really in the scope of this PR.

admacleod, line 85: Ideally no new proto definitions should be added to this file, it should be considered deprecated. Instead messages and RPCs should be added in a new package. Perhaps something like `proto/testkube/webhook/v1`?

Please try to follow best practices there (one file per enum/message/service, do not send `bytes` as a message field rather than actually adding the required fields to be passed around, using up to date protobuf specification `edition = "2023";` rather than `syntax = "proto3";`). The linter will help to pick up any issues (which will be ignored if you add them here and then some poor contributor is going to have to clean up in a month or two's time).

admacleod: Agreed, this is fairly easy to spot as a long lived stream. Instead consider polling `ListWebhooks`.

admacleod: @ed382 removing streaming RPCs is not in scope, but not adding new ones any more is everyone's responsibility. Please be a responsible contributor and avoid introducing known tech debt.

ed382: I did not get some of these memos.

PR #6738, opened 2025-10-06
fix: event loop unimplemented backoff

admacleod:
In the event that the Control Plane does not
support the event stream endpoint, we should not
exit and cause a reconnection attempt.

Instead we should backoff and try again at a
later time. The time to backoff is set to start
at a 1 day growing to a maximum of 7 days.


pkg/agent/agent.go
```diff
@@ -184,21 +185,31 @@ func (ag *Agent) runEventsReaderLoop(ctx context.Context) (err error) {
 		return errors.Wrap(err, "failed to setup events stream")
 	}
 
+	// Backoff a day at a time up to a week.
+	// This backoff is for retrying against a Control Plane that does not support
+	// The required endpoint, so if we don't want to wait then instead we can just be restarted.
+	b := backoff.New(7*24*time.Hour, 24*time.Hour)
```
vsukhin: move them to const

vsukhin (approved):

PR #6740, opened 2025-10-06
chore: merge helm-charts repo into testkube repo

WitoDelnat:
This PR closes https://linear.app/kubeshop/issue/TKC-4316/merge-helm-charts-repo-into-testkube-repo

⚠️ ~do not merge this without taking the latest changes from helm-charts repository!~ updated

### DIFF
--- a/k8s/helm/testkube/Chart.yaml
+++ b/k8s/helm/testkube/Chart.yaml
@@ -5,7 +5,7 @@ type: application
 version: 2.3.0
 dependencies:
   - name: testkube-operator
-    version: 2.1.154
+    version: 2.x.x
     repository: "file://../testkube-operator"
   - name: mongodb
     condition: mongodb.enabled
@@ -20,13 +20,8 @@ dependencies:
     version: 1.2.6-3
     repository: "file://./charts/nats"
   - name: testkube-api
-    version: 2.3.0
-    #repository: https://kubeshop.github.io/helm-charts
+    version: 2.x.x
     repository: "file://../testkube-api"
-  - name: testkube-logs
-    version: 0.2.0
-    repository: "file://../testkube-logs"
-    condition: global.features.logsV2
   - name: global
-    version: 0.1.3
+    version: 0.x.x
     repository: "file://../global"
--- a/k8s/helm/testkube-runner/Chart.yaml
+++ b/k8s/helm/testkube-runner/Chart.yaml
@@ -6,5 +6,5 @@ version: 2.3.1
 appVersion: 2.3.0
 dependencies:
 - name: global
-  version: 0.1.3
+  version: 0.x.x
   repository: "file://../global"


.github/workflows/lint.yaml
```diff
@@ -14,26 +12,27 @@ permissions:
   pull-requests: read
 
 jobs:
```
WitoDelnat, line 14: All these jobs run <50s. I merged them together in one job which runs <2m which is still faster than our tests; so not on the critical path. This gives less job bloat when looking at GitHub PRs.

.github/workflows/new-build.yaml
```diff
@@ -114,3 +130,75 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - id: helm-publish
+        name: Publish Helm Chart
+        run: |
+          REGISTRY=registry-1.docker.io
+          REGISTRY_GAR=${{ env.GAR_LOCATION }}-docker.pkg.dev
+          
+          echo ${{ secrets.DOCKERHUB_TOKEN }} | helm registry login ${REGISTRY} --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+          
+          # testkube chart
```
WitoDelnat, line 162: This no longer publishes sub-charts like testkube-operator and testkube-api directly. I don't think anyone uses them so this should be fine and will be a stepping stone to removing them later onwards.

.github/workflows/release.yaml
WitoDelnat, line 1: The CLI is now build together with all other Docker Images, this will give us confidence that main has no broken CLI builds.

- release.yaml will still use GoReleaser to build and distribute for different platforms (MacOs, Windows, Linux, ...). 
- release-dev.yaml has been removed as it had been creating Github Releases which was creating lots of bloat.

.github/workflows/release.yaml
```diff
@@ -146,79 +80,10 @@ jobs:
             ${{ matrix.name }}/testkube_${{ matrix.name }}_*
           retention-days: 1
 
-  build_agent_image:
```
WitoDelnat, line 149: I moved this to the bottom of the file. It should probably be its own GitHub Action as it has nothing to do with publishing the CLI, but for now I've just moved it a bit out of the way.

.github/workflows/release.yaml
```diff
@@ -419,25 +282,73 @@ jobs:
           ### Publish repo
           curl --fail-with-body -u ${USERNAME}:${PASSWORD} -X PUT -H 'Content-Type: application/json' --data '{"Snapshots": [{"Component": "main", "Name": "testkube-'${VERSION}'"}]}'}], http://repo.testkube.io:8080/api/publish/:linux/linux
 
-  trigger-argocd-image-build:
```
WitoDelnat, line 422: This job was legacy that was no longer doing anything

k8s/helm/testkube/Chart.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v2
+name: testkube
+description: Testkube is an open-source platform that simplifies the deployment and management of automated testing infrastructure.
+type: application
+version: 2.2.10
+appVersion: 2.2.10
+dependencies:
+  - name: global
```
WitoDelnat: All these charts are copy pasted so you can pretty much ignore reviewing them. Only in this file I made some changes. The versions for our own sub-charts are wildcarded so that they no longer need to be updated all the time. I've also added the appVersion as this was missing (probably because in the past app version of operator and api was different so no unified app version exists - no longer the case now).

Makefile
```diff
@@ -435,7 +435,7 @@ generate-sqlc: ## Generate sqlc package with sql queries
 
 .PHONY: generate-crds
 generate-crds: ## Generate Kubernetes CRDs from kubebuilder Golang structs.
-	go tool controller-gen crd:allowDangerousTypes=true object paths="./api/..." output:crd:dir=k8s/crd
+	go tool controller-gen crd:allowDangerousTypes=true object paths="./api/..." output:crd:dir=k8s/crd && node js/scripts/crd-postprocess.js
```
WitoDelnat: In the past we'd have to go to kubeshop/helm-charts and manually start adding the CRDs in between the Helm templating syntax, this completely automates this 🎉

.builds-darwin.goreleaser.yml
```diff
@@ -12,7 +12,6 @@ builds:
       - darwin
     goarch:
       - amd64
-      - arm64
```
admacleod, line 15: Is this the wrong way around? I thought we did want to build `darwin/arm64` as that represents an Apple Silicon target?

.builds-linux.goreleaser.yml
```diff
@@ -23,9 +11,7 @@ builds:
     goos:
       - linux
     goarch:
-      - amd64
```
admacleod, line 26: I think we probably do still want to support `linux/amd64`, you can probably drop `linux/arm64` though, I don't think we have many people trying to use the CLI on Linux ARM.

.builds-windows.goreleaser.yml
```diff
@@ -11,7 +11,6 @@ builds:
     goos:
       - windows
     goarch:
-      - amd64
```
admacleod, line 14: On the third one of these and I'm thinking that either I am confused by what these files are for or you got `amd64` and `arm64` the wrong way around as I'm pretty sure we want to keep `windows/amd64` support.

Makefile
admacleod: Rather than using `&&` in Make recipes it is probably better to split it across two lines, unless you really do want to use `&&` here for some reason.

cmd/kubectl-testkube/commands/agent.go
```diff
@@ -23,11 +25,18 @@ func NewAgentCmd() *cobra.Command {
 
 			ui.Logo()
 			ui.Info("Client Version", common.Version)
-			ui.Info("Server Version", info.Version)
-			ui.Info("Commit", common.Commit)
-			ui.Info("Built by", common.BuiltBy)
-			ui.Info("Build date", common.Date)
-
+			if info.Version != "" && !strings.Contains(info.Version, "invalid character") {
```
admacleod: Can you add a comment here explaining why this build variable might contain the text "invalid character" and why we would choose to ignore such cases rather than fix the source of the issue?

.goreleaser.yml
```diff
@@ -3,17 +3,30 @@ builds:
...
       - v1
     prebuilt:
       path: "{{ .Os }}/testkube_{{ .Os }}_{{ .Arch }}{{ with .Amd64 }}_{{ . }}{{ end }}/kubectl-testkube{{ .Ext }}"
     binary: kubectl-testkube
+  - builder: prebuilt
+    goos:
+      - windows
+    goarch:
+      - arm64
+      - 381
```
admacleod:
```suggestion
      - 386
```

.goreleaser.yml
```diff
@@ -3,17 +3,30 @@ builds:
...
     goamd64:
       - v1
     prebuilt:
       path: "{{ .Os }}/testkube_{{ .Os }}_{{ .Arch }}{{ with .Amd64 }}_{{ . }}{{ end }}/kubectl-testkube{{ .Ext }}"
     binary: kubectl-testkube
+  - builder: prebuilt
+    goos:
+      - windows
+    goarch:
+      - arm64
```
admacleod:
```suggestion
      - amd64
```

.goreleaser.yml
```diff
@@ -3,17 +3,30 @@ builds:
...
+    prebuilt:
+      path: "{{ .Os }}/testkube_{{ .Os }}_{{ .Arch }}{{ with .Amd64 }}_{{ . }}{{ end }}/kubectl-testkube{{ .Ext }}"
+    binary: kubectl-testkube
+  - builder: prebuilt
+    goos:
       - darwin
     goarch:
       - amd64
```
admacleod:
```suggestion
      - arm64
```

.goreleaser.yml
```diff
@@ -3,17 +3,30 @@ builds:
   - builder: prebuilt
     goos:
       - linux
-      - windows
+    goarch:
+      - arm64
```
admacleod:
```suggestion
      - amd64
```

cmd/kubectl-testkube/commands/agent.go
WitoDelnat: I've removed the check. I'll just let it be like it is and we can look at the root cause.

Makefile
WitoDelnat: Updated. It was due to a lack of understanding how makefile works.

admacleod (approved):
Looks ok to me for the parts that I understand enough to comment on. Definitely needs a review from @kubeshop/testkube-devops in here.

ypoplavs (approved):

PR #6754, opened 2025-10-14
feat: add NewArchitecture scheduling to standalone deployment

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -99,7 +99,7 @@ func main() {
 	features := commons.MustGetFeatureFlags()
 
 	mode := common.ModeAgent
-	if cfg.TestkubeProAPIKey == "" || cfg.TestkubeProAgentRegToken == "" {
+	if cfg.TestkubeProAPIKey == "" && cfg.TestkubeProAgentRegToken == "" {
 		mode = common.ModeStandalone
 		cfg.TestkubeProURL = fmt.Sprintf("%s:%d", cfg.APIServerFullname, cfg.GRPCServerPort)
 		cfg.TestkubeProTLSInsecure = true
@@ -209,7 +209,7 @@ func main() {
 	grpcClient := cloud.NewTestKubeCloudAPIClient(grpcConn)
 
 	// If we don't have an API key but we do have a token for registration then attempt to register the runner.
-	if cfg.TestkubeProAPIKey == "" || cfg.TestkubeProAgentRegToken != "" {
+	if cfg.TestkubeProAPIKey == "" && cfg.TestkubeProAgentRegToken != "" {
 		runnerName := cfg.RunnerName
 		if runnerName == "" {
 			// Fallback to a set name, but this is unlikely to be unique.


cmd/api-server/commons/commons.go
```diff
@@ -349,9 +349,9 @@ func ReadProContext(ctx context.Context, cfg *config.Config, grpcClient cloud.Te
 		LogStreamWorkerCount:                cfg.TestkubeProLogStreamWorkerCount,
 		Migrate:                             cfg.TestkubeProMigrate,
 		DashboardURI:                        cfg.TestkubeDashboardURI,
-		NewArchitecture:                     grpcClient == nil,
-		CloudStorage:                        grpcClient == nil,
-		CloudStorageSupportedInControlPlane: grpcClient == nil,
+		NewArchitecture:                     false,
```
WitoDelnat: This was a very strange one…. the `grpcClient` is always initialised.  Furthermore, the argument`grpcClient cloud.TestKubeCloudAPIClient` is not a pointer so this is never nil. I've set it to `false` as this more clearly is the intend.

pkg/agent/client/client.go
```diff
@@ -110,32 +109,32 @@ func NewGRPCConnectionWithTracing(
 	// All of step 2 should be removed in the future when Control Planes are required to create TLS connections.
 
 	// Default credentials using the system CAs to verify server certificates.
-	certPool, err := x509.SystemCertPool()
```
WitoDelnat, line 118: Woeps, this is to be reverted! build-in control plane will always take this ping and it makes starts quite slow. I did not want to bother with changes here now so I had temporarily commented it.

pkg/controlplane/scheduling/enqueuer.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+	}
+	// END POST-COMMIT VALIDATION
+
+	executions, err := e.persistExecution(commitContext, intermediateExecutions, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO WITO: Should Standalone agents dispatch events over embedded NATS?
```
WitoDelnat: Remove "TODO WITO". This is to be addressed before merging.

cmd/api-server/main.go
```diff
@@ -751,7 +730,14 @@ func main() {
 		deprecatedSystem.API.Init(httpServer)
 	}
 
+	isStandalone := mode == common.ModeStandalone
+	var executionController scheduling.Controller
+	if isStandalone && controlPlane != nil {
+		executionController = controlPlane.ExecutionController
```
WitoDelnat, line 731: This is not ideal, but it's temporary. Once the PRO control plane is the source of truth, the HTTP Server can be moved into the OSS and this breaking of encapsulation disappears. Also see [this comment](https://github.com/kubeshop/testkube/pull/6754/files#diff-0dca34dfa6f7a870d7a571a312eb1cac9a9fd2a4a2efd0f49cec5b10a95a5e6aR135)

internal/common/mode.go
```diff
@@ -5,6 +5,19 @@ import "errors"
 const (
 	ModeStandalone = "standalone"
 	ModeAgent      = "agent"
+
+	// note: Testkube Agent should not really have knowledge about organizations and environments.
+	// It appears that over time this has seeped into this codebase. Often empty strings are used
+	// to make function works. These constants make usages explicit to determine if we can and should get rid of this.
+
+	// WARNING: Be extremely cautious renaming this from an empty string to anything else. I've discovered that
```
WitoDelnat: I can remove this warning. I have removed all places where we did this.

sqlc.yaml
```diff
@@ -15,3 +15,103 @@ sql:
             go_type:
               import: "github.com/google/uuid"
               type: "UUID"
+          - column: "blub.bar"
+            go_type:
+              type: "Bar"
+              slice: true
+  # Use SQLC for jsonb parsing instead of doing it all manually
+  - engine: "postgresql"
```
WitoDelnat, line 23: The default way we've been handling PSQL is by using AI to generate dedicated. The AI generated code is enormous and I genuinely worry about maintaining it. For example, adding a field means we have to update every query, but its so much code that I doubt it will not fit in an AI context window so you have to start manually verifying everything or risk some query to be overlooked.

This approach leverages SLQC to do these things for us through its `sqlc.embed` and using type overrides features which will respectively allow code reuse and automated parsing of `jsonb` columns. 

I'm here introducing a second client to (1) experiment with this approach and (2) refactoring all other PSQL related code would be too much change for this already _enormous_ PR.

cmd/api-server/main.go
```diff
@@ -97,13 +98,16 @@ func main() {
...
 	features := commons.MustGetFeatureFlags()
 
+	mode := common.ModeStandalone
+	if cfg.TestkubeProAPIKey != "" || cfg.TestkubeProAgentRegToken != "" {
+		mode = common.ModeAgent
+	} else {
+		cfg.TestkubeProURL = fmt.Sprintf("%s:%d", cfg.APIServerFullname, cfg.GRPCServerPort)
+		cfg.TestkubeProTLSInsecure = true
+	}
+
```
admacleod: I know this is just the same code moved but I really hate an `else` statement, we could invert the logic here:

```suggestion
	mode := common.ModeAgent
	if cfg.TestkubeProAPIKey == "" && cfg.TestkubeProAgentRegToken == "" {
		mode = common.ModeStandalone
		cfg.TestkubeProURL = fmt.Sprintf("%s:%d", cfg.APIServerFullname, cfg.GRPCServerPort)
		cfg.TestkubeProTLSInsecure = true
	}

```

internal/config/config.go
```diff
@@ -225,12 +225,11 @@ type Config struct {
 	TransferEnvVariables            []string `envconfig:"TRANSFER_ENV_VARS" default:"GRPC_ENFORCE_ALPN_ENABLED"`
 	EnableK8sEvents                 bool     `envconfig:"ENABLE_K8S_EVENTS" default:"true"`
 	TestkubeDockerImageVersion      string   `envconfig:"TESTKUBE_DOCKER_IMAGE_VERSION" default:""`
-	DisableDeprecatedTests          bool     `envconfig:"DISABLE_DEPRECATED_TESTS" default:"false"`
+	DisableDeprecatedTests          bool     `envconfig:"DISABLE_DEPRECATED_TESTS" default:"true"`
```
admacleod: Does this change mean that users will have to manually set this flag to "false" in order to be able to run Tests and TestSuites?

If so...why didn't we make this change months ago!?

pkg/agent/client/client.go
```diff
@@ -129,7 +129,7 @@ func NewGRPCConnectionWithTracing(
 		}
 	}
 
-	// Attempt to use a TLS connection.
+	//Attempt to use a TLS connection.
```
admacleod: Weird change, probably just undo this particular file.

pkg/runner/grpc/client.go
```diff
@@ -57,29 +57,36 @@ type Client struct {
...
 func NewClient(conn grpc.ClientConnInterface, logger *zap.SugaredLogger, r runner, apiToken, organisationId string, controlPlane testworkflowconfig.ControlPlaneConfig, workflows workflowStore) Client {
 	client := executionv1.NewTestWorkflowExecutionServiceClient(conn)
+
+	opts := []grpc.CallOption{
+		// In the event of a transient failure on the server wait for it to come back rather than
+		// failing immediately.
+		grpc.WaitForReady(true),
+	}
+
+	// Standalone deployment does not have an API Token for now
```
admacleod, line 67: Have you tested that this still works? I think it might be ok provided that standalone uses either a unix socket or sets skipverify for connections.

cmd/api-server/main.go
WitoDelnat: I'd go a step further. I do not believe that the mode should be derived from the existence of a key. Before going into a rabbit hole I just copy pasted it. I'll take in your suggestion though as there is no harm 👍

internal/config/config.go
WitoDelnat: I've reverted this. It's overridden by default in the Helm Chart but not in the service, and initially I was running this locally and not on Kubernetes. Imho, we do _way_ too much in the Helm Chart and our defaults should be revisited at some point.

pkg/runner/grpc/client.go
WitoDelnat: I've tested both standalone and commercial deployment. It works unless I missed something in the tests.

As for the recommendation, I'd avoid using unix sockets as Execution Pods also use the gRPC Server (it should be using Execution Tokens for this which are also not there yet for now). Once the server is exposed, it's better to just incorporate proper credentials. It also keeps the system unified.

cmd/api-server/main.go
```diff
@@ -97,13 +98,15 @@ func main() {
 	cfg := commons.MustGetConfig()
 	features := commons.MustGetFeatureFlags()
 
+	mode := common.ModeAgent
+	if cfg.TestkubeProAPIKey == "" || cfg.TestkubeProAgentRegToken == "" {
```
admacleod: If we want to maintain the previous behaviour then this should be `&&`. Good ol' De Morgan!

WitoDelnat: Fixed 🤦

PR #6756, opened 2025-10-15
feat: Add RPC definitions for sync functionality

admacleod:
Whilst I hate that these RPCs use `bytes` to
carry structured data, my attempts to generate
or manually copy these data structures into
proto definitions were fruitless so until we
have some sort of CRD generation that can
create protobuf definitions for us this is the
best we can hope for. I did try `go-to-protobuf`
which is what the Kubernetes project uses, and
whilst it was mostly fine it would have required
some wider changes to the repository to make it
work nicely with the other protobuf changes that
we have here, so maybe in the future for a v2
of these RPCs we can implement them, but for
now this is the best we can hope for.


proto/testkube/testworkflow/template/v1/test_workflow_template_service.proto
```diff
@@ -0,0 +1,23 @@
...
+
+package testkube.testworkflow.template.v1;
+
+import "testkube/testworkflow/template/v1/delete_request.proto";
+import "testkube/testworkflow/template/v1/delete_response.proto";
+import "testkube/testworkflow/template/v1/update_or_create_request.proto";
+import "testkube/testworkflow/template/v1/update_or_create_response.proto";
+
+// TestWorkflowTemplateService contains RPCs related to manipulation of Test Workflow Template objects.
+service TestWorkflowTemplateService {
```
WitoDelnat: I see that you decided to go for a service per entity. I would perhaps have considered a `SyncService` myself or something like that. Or alternatively when splitting it up more, TestWorkflows do not work without TestWorkflowTemplates so they seem so intertwined that having separate service feels a bit off imho. I worry we will lead to the equivalent of Anemic Domain Models but for our interfaces.

Penny for your thoughts on the matter?

---

update 1: Since CR are passed as `bytes`, perhaps there is even little point in splitting them up? Perhaps an `UpdateOrCreateRequest { apiVersion string; kind string; yaml string; }` is sufficient here and easier to add extend this in the future?

update 2: All of this would also be so much easier if agents could simply call REST API with proper auth support for Agent Tokens. Imho the "ideal" solution here is that none of this should need to exist to begin with. We're making so many handlers duplicate for little reason. Perhaps the design here matters little and is just a short in-between for merging Agent and REST API as we discussed in the past.

admacleod: That is a good point, I had considered it but what I want to avoid is something that looks like:
```
service SyncService {
  rpc UpdateOrCreateTestWorkflow;
  rpc UpdateOrCreateTestWorkflowTemplate;
  rpc UpdateOrCreateWebhook;
  // etc.
```
it doesn't feel quite right to do that, I do understand the concern. My thinking was that in each object's "vertical" that there is a specific CRD, controller, RPC, storage backend. Each with its own data structure and potentially different logic. But perhaps that can be represented differently. I'll do some reading up and see if there is a nicer way to represent this.

And no, do not be distracted by the use of `bytes`, please treat each one as its own separate data type. Although, we potentially could achieve it with a `oneof`, that actually may be a better way to combine these and represent the different data types in a single RPC service 🤔

WitoDelnat: Perhaps a thought experiment. What API would you come up with if the objective was to design REST endpoints that extend Testkube's default REST API? Either reuse anything that exists, or create something new that would be under some `/agent` prefix, etc.

admacleod: I actually think that may be the thinking that led me down this incorrect path. Really RPCs are about actions performed, rather than objects they are performed upon. See the latest changes for an alternative, cleaner approach, it keeps the service definition nice and tight but allows for easy expansion to cover additional objects without causing breaking changes in the future.

proto/testkube/sync/v1/update_or_create_request.proto
```diff
@@ -0,0 +1,30 @@
...
+import "testkube/sync/v1/test_trigger.proto";
+import "testkube/sync/v1/test_workflow.proto";
+import "testkube/sync/v1/test_workflow_template.proto";
+import "testkube/sync/v1/webhook.proto";
+import "testkube/sync/v1/webhook_template.proto";
+
+// UpdateOrCreateRequest is the message sent by a client when
+// requesting the update or creation of a Testkube object.
+message UpdateOrCreateRequest {
+  // Payload is the protobuf represenation of the object that
```
WitoDelnat: Nitpick: represenation -> representation ;p

admacleod: Don't know why my spell checker wasn't picking that up 😅

PR #6757, opened 2025-10-15
feat: webhooks to control plane

ed382:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/repository/leasebackend/k8s/k8s_test.go
+++ b/pkg/repository/leasebackend/k8s/k8s_test.go
@@ -2,8 +2,8 @@ package k8s
 
 import "testing"
 
-func TestLeaseName_Default(t *testing.T) {
-	b := NewK8sLeaseBackend(nil, "")
+func TestLeaseName_Prefix(t *testing.T) {
+	b := NewK8sLeaseBackend(nil, "testkube-triggers-lease", "")
 
 	name := b.leaseName("cluster-a")
 
@@ -13,7 +13,7 @@ func TestLeaseName_Default(t *testing.T) {
 }
 
 func TestLeaseName_Override(t *testing.T) {
-	b := NewK8sLeaseBackend(nil, "", WithLeaseName("custom-lease"))
+	b := NewK8sLeaseBackend(nil, "lease-prefix", "", WithLeaseName("custom-lease"))
 
 	name := b.leaseName("cluster-b")
 


cmd/api-server/main.go
```diff
@@ -157,6 +156,7 @@ func main() {
 	log.DefaultLogger.Infow("connected to Kubernetes cluster successfully", "namespace", cfg.TestkubeNamespace)
 
 	var eventsEmitter *event.Emitter
+	// TODO(emil): what is this lazy emitter for?
```
WitoDelnat: I had to figure this out for LazyRunner too - which is a bit more obvious. There we basically want to Dependency Inject the Runner into a service which is defined before the Runner is created. This is because there is a cyclic dependency during creation time:
 
- Creating standalone control plane required a runner; ❌ cannot construct such runner!
- Creating gRPC connection required a control plane;
- Creating a Runner required a gRPC connection.

So in summary, Dawid introduced this Lazy concept to inject this dependency and initialise properly, as long as the lazy instance is initialised before runtime ever gets to it, it solves this problem:

- Create LazyRunner with empty reference
- Creating standalone control plane requires a lazy runner; ✅  no more problem now!
- Creating gRPC connection required a control plane;
- Creating a Runner required a gRPC connection.
- LazyRunner.Runner = runner
- ControlPlane.execute -> works because lazyRunner has a reference

I guess LazyEmiter had a similar problem, but it's no longer relevant. [In my PR I have removed it](https://github.com/kubeshop/testkube/pull/6754/files#diff-9a806b47e2c7033d8cd996647dab85acf24df2026be1ffe009df197a35956002L160).

cmd/api-server/main.go
```diff
@@ -208,28 +211,24 @@ func main() {
...
-	if cfg.TestkubeProAPIKey == "" && cfg.TestkubeProAgentRegToken != "" {
+	// Agents should be able to make an initial registration with their
+	// registration token and subsequent registrations with their API key to
+	// enable them to update their set of capabilities.
+	// To transition super agents to the new capabilities based registration,
+	// the charts have been changed to set the old super agent API key value as
+	// their registration token and registration has been enabled with the use
+	// of the API key. After intial self registration the super agent will be
+	// able to use the API key in the generated secret to reregister.
+	if cfg.TestkubeProAPIKey != "" || cfg.TestkubeProAgentRegToken != "" {
```
WitoDelnat, line 234: Let me try to rephrase this to see if I understand the flow correctly. Please correct me if there is a misunderstanding!

- The new (i.e. >v2.5) agent charts will be changed where the API KEY in values.yaml will not be a assigned to the registration token environment variable; and not the agent token environment variable.
- The super agent registers as a normal agent; it will receive a normal agent key which is stored in a Kubernetes secret.
- The agent now operates as this normal agent; on reboots it will use this Kubernetes secrets

Can you elaborate what happens when that Kubernetes secret gets removed, or the Helm chart gets uninstalled and reinstalled in a new namespace, etc?

ed382: > The new (i.e. >v2.5) agent charts will be changed where the API KEY in values.yaml will not be a assigned to the registration token environment variable; and not the agent token environment variable.

The old super agent API key will be _swapped_ into the new registration token environment variable. The super agent will now be able to register with its old super agent API key until we decide to deprecate that in the future.

> Can you elaborate what happens when that Kubernetes secret gets removed, or the Helm chart gets uninstalled and reinstalled in a new namespace, etc?

As one can now use the old super agent key or the new registration token to _update_ the registration for the agent; the dependency on the secret creation/updating has been removed.

If one is unable to create/update the secret a warning is shown that says that the registration token (which can be the old super agent key will be reused on the next registration); preferably we want to perform a proper registration and get an API key back and save it in a secret, because registration tokens could be rotated/expired frequently and the super agent API keys are going to be deprecated and eventually removed as a registration mechanism.

Note, the agent registration is unique by runner name within the organization. Super agents receive the environment ID name for there runner name on the backend to make sure their registration is unique across the organization.

WitoDelnat: Thank you Emil. Does this mean that given a valid registration token, an agent can update and hi-jack any other agent simply by spoofing its name?

ed382: The current join token currently allows creating agent registrations within an environment - these changes simply allows the join token to also update registrations within the environment.

Not sure, I would call this "hi-jacking" as this is simply the scope of the token. We have options in our charts which allow people to disable secret creation, so it is hard for us to make the seamless (just chart update, a primary requirement for this work) transition without removing the dependence on secrets during self registration. Additionally, it is possible that requiring the secret to be present after registration will make the agent registration less robust, creating support load.

Perhaps, in future work though, we should look into the scope and validity of these join tokens. Maybe they should only create/update runners with a certain name prefix and have shorter lifetimes (~90 days).

What is your concern with giving the join token _update_ permissions?

pkg/event/emitter.go
```diff
@@ -36,184 +42,228 @@ type Interface interface {
...
+		name := listeners[i].Name()
+		if e.listenerExists[kind] == nil {
+			e.listenerExists[kind] = make(map[string]struct{})
+		}
+		if _, exists := e.listenerExists[kind][name]; !exists {
+			// Mark as existing and append to listeners array
+			e.listenerExists[kind][name] = struct{}{}
+			e.listeners = append(e.listeners, listeners[i])
+		}
+	}
```
admacleod, line 73: The use of mutex locks in this object are quite awkward and could lead to future developers creating either race conditions or deadlocks.

You could simplify this by removing the `listenerExists` map entirely and replacing it with a simple slice lookup when adding listeners:
```suggestion
	e.mutex.Lock()
	defer e.mutex.Unlock()
	for _, listener := range listeners {
		if !slices.ContainsFunc(e.listeners, func(l common.Listener) bool {
			return l.Kind() == listener.Kind() && l.Name() == listener.Name()
		}) {
			e.listeners = append(e.listeners, listener)
		}
	}
```
by moving the mutex in to this function as well it can help reduce the chance of an incorrect usage of the mutex. Of course there is still the chance that someone introduces a deadlock by locking and then calling this function, if we want to guard against that then it would make sense to wrap `common.Listeners` in our own type in here so that we can ensure that it is safe to use concurrently:
```
type safeListeners struct {
	mutex     sync.Mutex
	listeners common.Listeners
}

func (s *safeListeners) Dump() []any {
	s.mutex.Lock()
	defer s.mutex.Unlock()
	return s.listeners.Dump()
}

func (s *safeListeners) Add(listener common.Listener) {
	s.mutex.Lock()
	defer s.mutex.Unlock()
	s.listeners = append(s.listeners, listener)
}

func (s *safeListeners) Copy() common.Listeners {
	s.mutex.Lock()
	defer s.mutex.Unlock()
	listeners := make([]common.Listener, len(s.listeners))
	copy(listeners, s.listeners)
	return listeners
}

func (s *safeListeners) Listeners() common.Listeners {
	s.mutex.Lock()
	defer s.mutex.Unlock()
	return s.listeners
}
```

ed382: I'm not sure how many listeners we are talking about here in a cloud setting, but using the map is computationally more efficient here.

Looking at the usages of this function again though, I think I can move the lock acquisition directly into it to simplify things.

ed382: Maybe this is better: https://github.com/kubeshop/testkube/pull/6757/commits/9b3ecb3bf767e845d081b008e157234b1fd16159

PR #6759, opened 2025-10-16
feat: Implement sync gRPC client

admacleod:
This implementation adds a layer of type
specificity on top of the RPC so that it is
easier to implement the controllers for
individual types.

These are not yet being initialised as the
controller implementations will need to be
completed first.

### DIFF
--- a/internal/sync/grpc/client.go
+++ b/internal/sync/grpc/client.go
@@ -1,0 +1,110 @@
+// Package grpc provides simple to use functions that call Contrl Plane gRPC endpoints
+// for updating the Control Plane about changes made to synchronised objects.
+package grpc
+
+import (
+	"context"
+	"time"
+
+	"github.com/cloudflare/backoff"
+	"go.uber.org/zap"
+	"golang.org/x/oauth2"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+
+	syncv1 "github.com/kubeshop/testkube/pkg/proto/testkube/sync/v1"
+)
+
+const defaultCallTimeout = time.Second * 30
+
+type Client struct {
+	OrganisationId string
+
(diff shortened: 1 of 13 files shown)


internal/sync/grpc/client.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		// Execute with our own call timeout context to prevent stalling out.
+		callCtx, cancel := context.WithTimeout(ctx, c.callTimeout)
+		// Add metadata to the call.
+		callCtx = metadata.AppendToOutgoingContext(callCtx, "organisation-id", c.OrganisationId)
+		// Attempt to call delete with a workflow that should not exist.
+		// This means that we should expect to receive a NotFound response
+		// from the server, and any other response represents an error of
+		// some form.
+		testId := "id-that-should-not-exist-vgyxwqavpd"
```
vsukhin, line 78: random or uid?

vsukhin: Or might be to have a sepate method to retrun CP version, it can be useful for the future changes

admacleod: I'm not a fan of using a randomly generated string here, it would make it much harder to track down a collision. I could have generated a UUID and put that in here but it would be much less descriptive in an error message as to the nature of what happened.

Having a separate method to return the Control Plane version isn't really appropriate because that would require every gRPC service that we create to re-implement the same call. Really the only way to check whether the server has an implementation, and that everything is functioning is to try a call and see what happens.

Normally such checking isn't really necessary and clients should be able to gracefully handle such situations. Unfortunately because we have been allowing insecure connections to Control Plane from Agents we have to perform this check, which is essentially a "have you correctly deployed your Agent" check, for each Service that wants to have a secure credentials transport.

Alternatively this check could be removed and when I pull this into `main.go` then I can further gate the behaviour behind a check that `TESTKUBE_PRO_SKIP_VERIFY` AND `TESTKUBE_PRO_TLS_INSECURE` are false. I'll have a think and circle back when I set that up.

vsukhin: ok, take your time, definitely something to think about

PR #6762, opened 2025-10-16
feat: Implement sync Kubernetes controllers

admacleod:
These controllers listen for changes on objects
that we wish to sync and then update an
abstract store interface depending on the
observed state of the object.

That abstract store interface is implemented by
the grpc client so these can all be neatly
aligned with the previous work.

### DIFF
--- a/internal/sync/controller/testtrigger.go
+++ b/internal/sync/controller/testtrigger.go
@@ -1,0 +1,81 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	testtriggersv1 "github.com/kubeshop/testkube/api/testtriggers/v1"
+)
+
+type testTriggerStore interface {
+	UpdateOrCreateTestTrigger(context.Context, testtriggersv1.TestTrigger) error
+	DeleteTestTrigger(context.Context, string) error
+}
+
+func NewTestTriggerSyncController(mgr ctrl.Manager, store testTriggerStore) error {
+	if err := ctrl.NewControllerManagedBy(mgr).
+		For(&testtriggersv1.TestTrigger{}).
+		Complete(testTriggerSyncReconciler(mgr.GetClient(), store)); err != nil {
+		return fmt.Errorf("create new sync controller for TestTrigger: %w", err)
+	}
+	return nil
+}
+
+func testTriggerSyncReconciler(client client.Reader, store testTriggerStore) reconcile.Reconciler {
+	return reconcile.Func(func(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+		var trigger testtriggersv1.TestTrigger
+		err := client.Get(ctx, req.NamespacedName, &trigger)
+		switch {
+		case errors.IsNotFound(err):
+			// Deleted, request deletion from store.
+			// Passing the name here rather than the namespaced name as generally we refer to objects
+			// purely by their name.
+			if err := store.DeleteTestTrigger(ctx, req.Name); err != nil {
+				// Unable to delete for some reason, request a retry.
+				// We might want to selectively handle different errors here, but ideally they should
+				// be handled in the store implementation. If we return abstracted error messages from
+				// the store then we should handle them here.
+				return ctrl.Result{}, fmt.Errorf("delete TestTrigger %q from store: %w", req.Name, err)
+			}
+			return ctrl.Result{}, nil
+		case err != nil:
+			return ctrl.Result{}, fmt.Errorf("retrieve TestTrigger %q from Kubernetes: %w", req.NamespacedName, err)
+		}
+
+		// Resource has been marked for deletion, we may not get an event when it finally goes so this
+		// is the moment when we should update the Control Plane.
+		// Kubernetes is a funny thing, when a resource is marked for deletion then the DeletionTimestamp
+		// is set, but the resource is not yet removed, giving a chance for controllers to do their thing
+		// run finalizers etc. before the resources is removed entirely. Once DeletionTimestamp is set
+		// there is no going back so we know this resource is about to be deleted.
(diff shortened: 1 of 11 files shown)


internal/sync/controller/testtrigger.go
```diff
@@ -0,0 +1,79 @@
...
+			// Passing the name here rather than the namespaced name as generally we refer to objects
+			// purely by their name.
+			if err := store.DeleteTestTrigger(ctx, req.Name); err != nil {
+				// Unable to delete for some reason, request a retry.
+				// We might want to selectively handle different errors here, but ideally they should
+				// be handled in the store implementation. If we return abstracted error messages from
+				// the store then we should handle them here.
+				return ctrl.Result{}, fmt.Errorf("delete TestTrigger %q from store: %w", req.Name, err)
+			}
+		}
```
topliceanurazvan, line 68: Should return `ctrl.Result{}, nil` after successfully deleting the trigger. Without this, execution continues to line 68 where `UpdateOrCreateTestTrigger()` is called, recreating the trigger that was just deleted.

(This same issue exists in all 5 controllers in this package)

internal/sync/controller/testtrigger.go
```diff
@@ -0,0 +1,79 @@
...
+			// Deleted, request deletion from store.
+			// Passing the name here rather than the namespaced name as generally we refer to objects
+			// purely by their name.
+			if err := store.DeleteTestTrigger(ctx, req.Name); err != nil {
+				// Unable to delete for some reason, request a retry.
+				// We might want to selectively handle different errors here, but ideally they should
+				// be handled in the store implementation. If we return abstracted error messages from
+				// the store then we should handle them here.
+				return ctrl.Result{}, fmt.Errorf("delete TestTrigger %q from store: %w", req.Name, err)
+			}
```
topliceanurazvan, line 44: Should return `ctrl.Result{}, nil` after successfully deleting from the store. Without this, execution continues to line 68, where `UpdateOrCreateTestTrigger()` is called with an uninitialized zero-value `trigger` object (since `client.Get()` returned NotFound and never populated it).

(This same issue exists in all 5 controllers in this package)

admacleod: Good catch! That was a subtle one. Updated on all controllers.

internal/sync/controller/testtrigger.go
admacleod: Updated on all controllers. Thanks!

internal/sync/controller/testworkflow.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := ctrl.NewControllerManagedBy(mgr).
+		For(&testworkflowsv1.TestWorkflow{}).
+		Complete(testWorkflowSyncReconciler(mgr.GetClient(), store)); err != nil {
+		return fmt.Errorf("create new sync controller for TestWorkflow: %w", err)
+	}
+	return nil
+}
+
+func testWorkflowSyncReconciler(client client.Reader, store testWorkflowStore) reconcile.Reconciler {
+	return reconcile.Func(func(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
WitoDelnat, line 30: I'm not sure how this works under the hood. Is it like a stream of events from the kube-api-server and we respond to this? Hypothetically, what would happen if agent is temporarily down (let's say a crash or whatever), a Workflow is deleted, and agent restarts. Will (1) this be handled correctly and lead to the workflow being deleted, or (2) be overlooked - and we'd need to implement a background process for error corrections?

Doesn't have to be implemented now, I just want to understand the limits of our system.

admacleod: Yes this is correct, the events will be missed whilst the controller is down. The only way around this is to maintain some state in the agent (or retrieve it from the control plane) of what it last saw, and then on startup get the current state of everything from the cluster, it can then diff those and take actions as required.

I've purposefully excluded that behaviour from the agent at the moment as there is no predicate filtering, so it would be fairly trivial to generate a change event from a sync. The only thing we would miss is a deletion, but having a workflow in the Control Plane that no longer exists in the cluster is a behaviour that we sort of support.

It will be interesting to note whether any users require/request this behaviour.

internal/sync/controller/testworkflow.go
```diff
@@ -0,0 +1,81 @@
...
+		// is the moment when we should update the Control Plane.
+		// Kubernetes is a funny thing, when a resource is marked for deletion then the DeletionTimestamp
+		// is set, but the resource is not yet removed, giving a chance for controllers to do their thing
+		// run finalizers etc. before the resources is removed entirely. Once DeletionTimestamp is set
+		// there is no going back so we know this resource is about to be deleted.
+		if !workflow.DeletionTimestamp.IsZero() {
+			// About to be deleted, request deletion from store.
+			// Passing the name here rather than the namespaced name as generally we refer to objects
+			// purely by their name.
+			if err := store.DeleteTestWorkflow(ctx, req.Name); err != nil {
```
WitoDelnat, line 60: More thinking out loud, but it could be interesting to have configurability for soft-delete instead of hard delete. I worry for data loss in the control plane because somebody accidentally kubectl'd something incorrectly or something like that.

Anyway.. not for now

admacleod: Interestingly I discovered we're kind of saved from this by our Control Plane implementation. Because related artefacts aren't cleaned up when the parent workflow is removed you can simply recreate the workflow with the same name and it will inherit all of the previous execution history 😆 . So if you're doing GitOps and simply revert the change with the incorrect deletion then everything should just go back to how it was.

vsukhin: It migth be a regression, even for Tests we deleted old executions and reset the exdcution counter, not sure about latest Test Workflow implementation, however

internal/sync/controller/testworkflow.go
vsukhin: Controller should be idempotemt by design, expect to receive the same event multiple times from informer cache, for example on controller reload, For example, for execution CR we checked this out

internal/sync/controller/testworkflow.go
admacleod: I don't think it is a regression because there is a checkbox in the UI when you delete a workflow to choose not to delete the execution history, so it is a decision to allow the ability to not totally purge executions when deleting a workflow.

PR #6768, opened 2025-10-20
feat: [TKC-4269] add running context builtin variables

povilasv:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/testworkflow-init/data/global.go
+++ b/cmd/testworkflow-init/data/global.go
@@ -16,8 +16,12 @@ func GetBaseTestWorkflowMachine() expressions.Machine {
 		wd = "/"
 	}
 	fileMachine := libs.NewFsMachine(os.DirFS("/"), wd)
-	state := GetState() // load state
+	GetState() // load state
+	return expressions.CombinedMachines(EnvMachine, StateMachine, fileMachine)
+}
+
+func ExecutionMachine() expressions.Machine {
+	state := GetState()
 	executionCfg := state.InternalConfig.Execution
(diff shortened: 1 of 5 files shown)


cmd/testworkflow-init/data/global.go
```diff
@@ -15,6 +16,8 @@ func GetBaseTestWorkflowMachine() expressions.Machine {
 		wd = "/"
 	}
 	fileMachine := libs.NewFsMachine(os.DirFS("/"), wd)
-	GetState() // load state
-	return expressions.CombinedMachines(EnvMachine, StateMachine, fileMachine)
+	state := GetState() // load state
+	executionCfg := state.InternalConfig.Execution
+	executionMachine := testworkflowconfig.CreateExecutionMachine(&executionCfg)
```
vsukhin: do we need to add it there, because previously execution machine was added separately, like

		data.GetBaseTestWorkflowMachine(),
		testworkflowconfig.CreateCloudMachine(&cfg.ControlPlane, orgSlug, envSlug),
		testworkflowconfig.CreateExecutionMachine(&cfg.Execution),
		testworkflowconfig.CreateWorkflowMachine(&cfg.Workflow),
		credentials.NewCredentialMachine(data.Credentials()),

povilasv: Thanks for review! 

I rewired it instead of using global

PR #6784, opened 2025-10-24
feat: add support for new credential types in expressions

dejanzele:
## Pull request description 

Add support for `encrypted`, `variable` and `vault` expression functions.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/credentials/expressions.go
+++ b/pkg/credentials/expressions.go
@@ -8,22 +8,11 @@ import (
 )
 
 func NewCredentialMachine(repository CredentialRepository, observers ...func(name string, value string)) expressions.Machine {
-	return expressions.NewMachine().RegisterFunction("credential", func(values ...expressions.StaticValue) (interface{}, bool, error) {
-		computed := false
-		if len(values) == 2 {
-			if values[1].IsBool() {
-				computed, _ = values[1].BoolValue()
-			} else {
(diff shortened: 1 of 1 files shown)


pkg/credentials/expressions.go
```diff
@@ -8,22 +8,11 @@ import (
...
-			} else {
-				return nil, true, fmt.Errorf(`"credential" function expects 2nd argument to be boolean, %s provided`, values[1].String())
-			}
-		} else if len(values) != 1 {
-			return nil, true, fmt.Errorf(`"credential" function expects 1-2 arguments, %d provided`, len(values))
-		}
-
-		name, _ := values[0].StringValue()
+	// Helper function to handle credential fetching logic
+	fetchCredential := func(name string, computed bool) (interface{}, error) {
```
vsukhin, line 12: all four types are fetched using the same repo method?

dejanzele: in a later point it is analyzed is it a `credential()` (kept for backwards compatibility), `encrypted()` (same as credential but a better/newer/more intuitive approach), `variable()` or `vault()`

PR #6786, opened 2025-10-24
feat: [TKC-4384] introduce leases-based leader election

povilasv:
## Pull request description 

  - Introduced a generic leader coordinator that runs shared subsystems (agent, cron scheduler, telemetry heartbeat, GitOps sync, event emitter, triggers) under a single lease so the API server can safely scale horizontally.
  - Reworked trigger service to consume the new coordinator and moved lease access into a reusable  component.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/k8s/helm/testkube-api/templates/role.yaml
+++ b/k8s/helm/testkube-api/templates/role.yaml
@@ -2,6 +2,29 @@
 apiVersion: {{ include "global.capabilities.rbac.apiVersion" . }}
 kind: Role
 metadata:
+  name: {{ printf "%s-lease-access" (include "testkube-api.fullname" .) }}
+  labels: {{- include "testkube-api.labels" . | nindent 4 }}
+    {{- if .Values.global.labels }}
+    {{- include "global.tplvalues.render" ( dict "value" .Values.global.labels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.global.annotations}}
+  annotations: {{- include "global.tplvalues.render" ( dict "value" .Values.global.annotations "context" $ ) | nindent 4 }}
+  {{- end }}
+rules:
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - get
+      - create
+      - update
+
+---
+
+apiVersion: {{ include "global.capabilities.rbac.apiVersion" . }}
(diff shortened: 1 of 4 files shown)


cmd/api-server/services/telemetry.go
```diff
@@ -28,21 +28,27 @@ func HandleTelemetryHeartbeat(ctx context.Context, clusterId string, configMapCo
 	}
 
 	ticker := time.NewTicker(heartbeatInterval)
+	defer ticker.Stop()
```
admacleod, line 31: You don't need to do this anymore https://pkg.go.dev/time#NewTicker
> As of Go 1.23, the garbage collector can recover unreferenced tickers, even if they haven't been stopped. The Stop method is no longer necessary to help the garbage collector. (Code may of course still want to call Stop to stop the ticker for other reasons.)

admacleod: As a simplification for this you can now use https://pkg.go.dev/time#Tick

pkg/coordination/leader/coordinator.go
```diff
@@ -0,0 +1,176 @@
...
+		c.mu.Unlock()
+		return
+	}
+	c.leader = false
+	c.active = nil
+	c.mu.Unlock()
+
+	c.logger.Infow("leader coordinator: releasing lease", "identifier", c.identifier, "clusterId", c.clusterID)
+
+	group.cancel()
```
vsukhin, line 179: is it worth to check for nil?

cmd/api-server/services/telemetry.go
povilasv: I think it doesnt work for my case

 Switching to time.Tick here would be a regression. time.Tick only exposes a receive-only channel and gives you no way to halt the  underlying ticker. In this loop we already defer ticker.Stop(), so the goroutine is correctly cleaned up when the context ends. 
 
 This can happen when we are no longer leader and we need stop sending heartbeats.

pkg/coordination/leader/coordinator.go
povilasv: Done :)

PR #6796, opened 2025-10-28
feat: [TKC-4233] workflow tests - playwright with video recording enabled

tkonieczny:


test/playwright/playwright-project/playwright.config.js
```diff
@@ -49,10 +49,10 @@ module.exports = defineConfig({
...
     },
 
-    {
-      name: 'firefox',
-      use: { ...devices['Desktop Firefox'] },
-    },
+    // {
+    //   name: 'firefox',
+    //   use: { ...devices['Desktop Firefox'] },
+    // },
```
WitoDelnat: Intended? Better to remove it if so

tkonieczny: That's how playwright generates this config by default - updated with cleaned up version (with all comments removed)

PR #6797, opened 2025-10-28
feat: enable new sync agent

admacleod:


k8s/helm/testkube-api/templates/deployment.yaml
```diff
@@ -245,12 +245,6 @@ spec:
               value: "{{ not .Values.next.legacyTests.enabled }}"
             - name: GITOPS_KUBERNETES_TO_CLOUD_ENABLED
               value: "{{ .Values.next.gitops.syncKubernetesToCloud }}"
-            - name: GITOPS_CLOUD_TO_KUBERNETES_ENABLED
```
WitoDelnat, line 248: So nice that we can update and review all of this in one PR 🥳

cmd/api-server/main.go
```diff
@@ -698,17 +536,49 @@ func main() {
 		})
 		commons.ExitOnError("creating kubernetes controller manager", err)
 
-		// Initialise controllers
-		err = controller.NewTestWorkflowExecutionExecutorController(mgr, testWorkflowExecutor)
-		commons.ExitOnError("creating TestWorkflowExecution controller", err)
+		// Create Sync Controllers
```
WitoDelnat, line 539: Could it be possible to pull this out of the main func?

admacleod: Potentially, I'd like to spend some time chopping up the entire main function as a separate piece of work once we've finished the work separating the different agent types as it would be nice to compartmentalise each of those groups so that it is much clearer.

PR #6799, opened 2025-10-28
fix: issues around credential expressions

dejanzele:
## Pull request description 

Add support for `credential()` and `vault()` expressions and a better way to distinguish them in Control Plane

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/credentials/expressions.go
```diff
@@ -7,10 +7,14 @@ import (
 	"github.com/kubeshop/testkube/pkg/expressions"
 )
 
+const (
+	SourceCredential = "credential"
+	SourceVault      = "vault"
```
WitoDelnat, line 12: Can you tell me more about this differentiation? Intuitively, I would expect a workflow author to refer a credential by its name `credential("foo")`. The source of the credential feels like a detail that can be resolved automatically?

dejanzele: Idea is to also give the option to reference vault credentials directly from vault.
I expect not always users would want to create creds via Credentials API, instead have existing ones in Vault.

PR #6814, opened 2025-11-06
fix: missing return

admacleod:
Not returning after an error
attempting to execute the workflow
here can lead to a nil pointer
dereference and panic.

### DIFF
--- a/pkg/runner/grpc/client.go
+++ b/pkg/runner/grpc/client.go
@@ -249,6 +249,7 @@ func (c Client) executeResponse(ctx context.Context, response *executionv1.GetEx
 				c.logger.Errorw("Failed to start execution.",
 					"executionId", start.GetExecutionId(),
 					"error", err)
+				return
 			}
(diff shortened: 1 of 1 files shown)


vsukhin (approved):

pkg/runner/grpc/client.go
```diff
@@ -249,6 +249,7 @@ func (c Client) executeResponse(ctx context.Context, response *executionv1.GetEx
 				c.logger.Errorw("Failed to start execution.",
 					"executionId", start.GetExecutionId(),
 					"error", err)
+				return
```
povilasv, line 252: Could we add a test here somehow?

admacleod: We could, but I'm working on a mechanism for reporting the failure back to the Control Plane which will change this behaviour again. When that is in I will add the test to ensure that the Control Plane is correctly updated about the failure.

PR #6824, opened 2025-11-07
fix: simplify and fix pro context

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/testworkflow-toolkit/env/client.go
+++ b/cmd/testworkflow-toolkit/env/client.go
@@ -18,19 +18,16 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/config"
 	config2 "github.com/kubeshop/testkube/cmd/testworkflow-toolkit/env/config"
-	"github.com/kubeshop/testkube/internal/common"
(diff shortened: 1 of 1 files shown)


cmd/testworkflow-toolkit/env/client.go
```diff
@@ -40,100 +37,53 @@ import (
...
-	if isExternalStorageCache == nil && cfg.Worker.FeatureFlags[testworkflowconfig.FeatureFlagCloudStorage] != "true" {
-		isExternalStorageCache = common.Ptr(false)
-	}
+func loadProContext() (*cloud.ProContextResponse, error) {
+	proContextMu.Lock()
+	defer proContextMu.Unlock()
 
-	// Do not check Cloud support if its already predefined
-	if isExternalStorageCache != nil {
-		return nil
```
WitoDelnat, line 97: This was the bug. In the past there used to be some NewArchitecture condition which accidentally always made us fetch this. Currently it would always make an early return and silently not fetch anything. 

This is quite error prone and just premature optimisation. There is little harm in getting the pro context from the control plane. It's still cached so we only do this at most once.

admacleod (approved):

PR #6828, opened 2025-11-07
feat: workflow retrieval rpc

admacleod:
This replaces passing workflows directly in the execution start message and provides a new rpc to retrieve a "resolved" workflow directly.


proto/testkube/testworkflow/execution/v1/get_execution_workflow_request.proto
```diff
@@ -0,0 +1,18 @@
...
+// GetExecutionWorkflowRequest is a call to a server to request
+// an enriched workflow specifically for the execution that a
+// client wishes to execute.
+message GetExecutionWorkflowRequest{
+  // execution_id is the unique identifier of the execution for
+  // which the workflow should be retrieved.
+  string execution_id = 1;
+  // workflow_name is the name of the workflow that should be
+  // retrieved for execution.
+  string workflow_name = 2;
```
WitoDelnat: Is the execution identifier by itself not sufficient? The environment should be handled by the agent token; the workflow name does not matter much as an execution has a one to one relationship with a workflow. It complicates usage of the client without providing much added value.

admacleod: Thats fair, I guess there isn't a case where an execution can have multiple workflows to execute because they would all have to get bundled into one single workflow.

EnvironmentID I was planning to include because the execution could be for execution in a different environment, but actually that could have been a hack to implement multi-environment runners, in the new RPC the expectation is that you would either poll separately for each environment, or better deploy a different runner for each environment.

admacleod: Ah, no environment ID is required here, because it is not required for the runner to connect to the control plane.

PR #6832, opened 2025-11-11
feat: [TKC-4507] add runner mode and labels when auto-registering through grpc

caiomede-tk:
## Pull request description 

Include runner mode and labels when auto-registering through grpc

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- add runner mode and labels when auto-registering through grpc

## Fixes

-

### DIFF
--- a/k8s/helm/testkube-api/templates/_helpers.tpl
+++ b/k8s/helm/testkube-api/templates/_helpers.tpl
@@ -114,6 +114,8 @@ Define API environment in agent mode
   value: "self-registration"
 - name: SECRET_CREATION_PREFIX
   value: "{{ include "testkube-api.fullname" . }}-"
+- name: RUNNER_IS_GLOBAL
+  value: "true"
 {{- if .Values.cloud.url }}
 - name: TESTKUBE_CLOUD_URL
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -170,6 +170,9 @@ type RunnerConfig struct {
 	RunnerName                string            `envconfig:"RUNNER_NAME" default:""`
 	FloatingRunner            bool              `envconfig:"FLOATING_RUNNER" default:"false"`
 	EventLabels               map[string]string `envconfig:"EVENT_LABELS" default:""`
+	IsGlobal				  bool			    `envconfig:"RUNNER_IS_GLOBAL" default:"false"`
```
ed382, line 173: This should be true for existing super agents, otherwise, they won't schedule unless targeted, right? I don't see changes to the Helm charts to make that the default, but maybe I'm missing something.

caiomede-tk: Thanks for reporting that @ed382 ! please review the last commit if you think that covers the default set up for super agents

ed382 (approved):

PR #6833, opened 2025-11-11
chore: clean up unused protobufs and RPCs

admacleod:
With recent changes and additions of new features it is possible to deprecate some no longer used RPCs and remove their associated code.


WitoDelnat (approved):
LGTM. You checked that pulling this update into testkube-cloud-api will allow compilation?

vsukhin (approved):

PR #6837, opened 2025-11-11
fix: updating/deleting loaded listeners

ed382:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/event/emitter.go
+++ b/pkg/event/emitter.go
@@ -57,7 +57,7 @@ type Emitter struct {
 }
 
 // uniqueListeners keeps a unique set of listeners by kind and name.
-// The last listeners for each kind and name combination takes precedence.
+// The last listener for each kind and name combination takes precedence.
 func uniqueListeners(listeners []common.Listener) []common.Listener {
 	set := make(map[string]struct{})
 	unique := make(common.Listeners, 0, len(listeners))


pkg/event/emitter.go
```diff
@@ -44,38 +44,57 @@ type Interface interface {
...
+func uniqueListeners(listeners []common.Listener) []common.Listener {
+	exists := make(map[string]map[string]common.Listener)
+
+	// Determine latest version for each listener by kind and name
 	for i := range listeners {
 		kind := listeners[i].Kind()
 		name := listeners[i].Name()
-		if e.listenerExists[kind] == nil {
-			e.listenerExists[kind] = make(map[string]struct{})
+		if exists[kind] == nil {
```
admacleod: You want map access checking here:

```suggestion
		if _, ok := exists[kind]; !ok {
```
normally I would use the variable name `exists` here rather than `ok`, but that won't work nicely because the map of maps is called that.

pkg/event/emitter.go
```diff
@@ -44,38 +44,57 @@ type Interface interface {
...
+	}
+	// Generate unique listeners slice
+	unique := make(common.Listeners, 0, len(listeners))
+	for _, kindMap := range exists {
+		for _, listener := range kindMap {
+			unique = append(unique, listener)
 		}
 	}
+	return unique
+}
```
admacleod, line 73: This can be achieved much faster, more readably, and with a lower memory footprint by reducing the number of loops here:

```suggestion
func uniqueListeners(listeners []common.Listener) []common.Listener {
	var unique common.Listeners
	var uniqSet map[string]struct{}
	for _, listener := range slices.Backward(listeners) {
		if _, exists := uniqSet[listener.Name()+listener.Kind()]; exists {
			continue
		}
		uniqSet[listener.Name()+listener.Kind()] = struct{}{}
		unique = append(unique, listener)
	}

	return unique
}
```

admacleod: If you care about the ordering then you can reverse the slice again at the end with https://pkg.go.dev/slices#Reverse. You could also preallocate the `unique` slice but I do not believe it will be necessary and it usually is a premature optimisation that can harm readability.

ed382: This is not quite the same logic, this takes the first instance of the listener, but this need to take the last one.
I did not want to use a concatenation as the unique key because it not exactly the constraint being applied here, perhaps if a delimiter is added it is a safer assumption to make, but I'm not sure this is more readable.

ed382: I see you made an edit, and yes with reverse this is a much better implementation - I just will add some delimeters. This part had some iterations on it, and somethings stuck around from the previous ones.

admacleod: No, it reads the slice backwards using https://pkg.go.dev/slices#Backward so that you get the most recently added unique listener.

I do not believe that you require a delimiter because `Kind` is a fixed value (it should be being represented by an enum but it isn't for some reason). So the only time you get a collision is if you have something weird like:
- Name: `"FooBar"` Kind: `""`
- Name: `"Foo"` Kind: `"Bar"`

in the same set, which I very much doubt we're going to see.

ed382: Yes, I was replying to an old version of the code. About delimiter, it doesn't hurt though.

admacleod: The code always had `Backward` in, but no matter. The delimiter shouldn't hurt true, you can always just chuck something fun in like an emoji just to make really sure someone isn't going to put that in the event name 😛

ed382: Addressed here: https://github.com/kubeshop/testkube/pull/6837/commits/6b315e98f080f505fef907632febe42fa8e44452
Thanks for the feedback.

PR #6839, opened 2025-11-12
chore: refactor cronjobs

admacleod:
This looks like a set of meaty changes but is in fact a fairly simple refactoring of the existing monolithic scheduled execution system to be more pluggable.

Agent specific implementations are moved to `internal` as we are not expecting them to be reused outside of this repository. The exposed `cronjob` package is reduced to a more reusable API which we can share with the Control Plane to manage workflow execution schedules in a common way even when the implementation specifics are different.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -170,8 +170,8 @@ type RunnerConfig struct {
 	RunnerName                string            `envconfig:"RUNNER_NAME" default:""`
 	FloatingRunner            bool              `envconfig:"FLOATING_RUNNER" default:"false"`
 	EventLabels               map[string]string `envconfig:"EVENT_LABELS" default:""`
-	IsGlobal				  bool			    `envconfig:"RUNNER_IS_GLOBAL" default:"false"`
-	RunnerGroup				  string			`envconfig:"RUNNER_GROUP" default:""`
+	IsGlobal                  bool              `envconfig:"RUNNER_IS_GLOBAL" default:"false"`
+	RunnerGroup               string            `envconfig:"RUNNER_GROUP" default:""`
 	RunnerLabelsPrefix        string            `envconfig:"RUNNER_LABELS_PREFIX" default:"runner.testkube.io/"`
 }
 


internal/cronjob/robfig/manager.go
```diff
@@ -0,0 +1,136 @@
...
+				Selector: &cloud.ScheduleResourceSelector{Name: workflow},
+				Config:   config.Config,
+				Targets:  targets,
+			}},
+		}
+
+		// Pro edition only (tcl protected code)
+		if m.proModEnabled {
+			request.RunningContext, _ = testworkflowexecutor.GetNewRunningContext(cronjobtcl.GetRunningContext(cronSpec), nil)
+		}
```
admacleod, line 95: I'm not sure if this is either still required or if it should be enabled by default. The check here is essentially `proContext.APIKey != ""` which I have moved up to `main.go`. Perhaps once the functionality is migrated to the Control Plane this would never be true as this manager would only be used on OSS where pro mode is never enabled.

@WitoDelnat @vsukhin

cmd/api-server/commons/commons.go
```diff
@@ -488,13 +483,7 @@ func CreateImageInspector(cfg *config.ImageInspectorConfig, configMapClient conf
...
 }
 
-func CreateCronJobScheduler(cfg *config.Config,
-	testWorkflowClient testworkflowclient.TestWorkflowClient,
-	testWorkflowTemplateClient testworkflowtemplateclient.TestWorkflowTemplateClient,
-	testWorkflowExecutor testworkflowexecutor.TestWorkflowExecutor,
-	logger *zap.SugaredLogger,
-	proContext *config.ProContext,
-) cronjob.Interface {
+func CronJobsEnabled(cfg *config.Config) bool {
```
admacleod, line 486: Functionality here is changed, perhaps we would want to log in `main.go` whether or not the cronjobs are enabled to somewhat preserve similar functionality (but really killing the whole application because we cannot read an optional piece of configuration feels like a poor choice).

admacleod: Added a log message https://github.com/kubeshop/testkube/pull/6839/files#diff-9a806b47e2c7033d8cd996647dab85acf24df2026be1ffe009df197a35956002R853 that indicates that cron job management is not enabled.

cmd/api-server/main.go
```diff
@@ -816,47 +815,42 @@ func main() {
...
-				u,
-				k8sctrlclient.InNamespace(cfg.TestkubeNamespace),
-				k8sctrlclient.MatchingLabelsSelector{Selector: labels.NewSelector().Add(reqs...)},
-			); err != nil {
-				log.DefaultLogger.Errorw("unable to delete legacy cronjobs", "error", err, "label", "testkube="+resource, "namespace", cfg.TestkubeNamespace)
-				continue
-			}
-		}
-
+	if commons.CronJobsEnabled(cfg) {
```
WitoDelnat, line 818: This would generally only ever be enabled for standalone deployment right?

I'd prefer to move this into the Standalone build-in Control Plane. It'll be much easier to reason about our system if there is a clear symmetry between running with an PRO or OSS deployment. Perhaps it'll also allow our main.go to be a bit less humongous.

admacleod: Eventually yes, but right now the Control Plane doesn't handle the cronjobs so this is enabled in a fun and weird way (part of the migration path from `testkube-operator` into the agent) because it is also enabled for enterprise deployments too.

admacleod: Oh, actually, I get it, if we move this into `controlplane` now, then when Super Agent concept is gone (we will have migrated handling to the Control Plane for enterprise deployments) then it won't have to be moved again?

admacleod: That isn't currently possible, because we currently make use of the `proContext` which is all set up after the standalone control plane. So once handling is migrated to the Control Plane this whole setup will need to be refactored again to be more amenable to running purely in Standalone mode.

WitoDelnat: This all ties into how we want to migrate source of truth which we discussed yesterday a bit. We could start making separate feature flags for this too, or just have a single superagent migration at agent boot that is required to succeed or we bail. (What's the point in many flags for cron/webhook/etc if we all want to flip em over at once). That way only control plane needs to be backwards compatible with older agents whereas the agent can have legacy moved / removed earlier.

So for this; this migration would enable crons in the control plane. We can already move OSS crons inside OSS Control Plane because it can simply never boot beyond this point. We can still do this within this release, but maybe let's first already get the control plane in place.

PR #6842, opened 2025-11-13
chore: remove legacy

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/cloud/data/testresult/testresult.go
+++ b/pkg/cloud/data/testresult/testresult.go
@@ -19,6 +19,8 @@ type CloudRepository struct {
 	executor executor.Executor
 }
 
+var _ testresult.Repository = (*CloudRepository)(nil)
+
 func NewCloudRepository(client cloud.TestKubeCloudAPIClient, proContext *config.ProContext) *CloudRepository {
 	return &CloudRepository{executor: executor.NewCloudGRPCExecutor(client, proContext)}
 }


pkg/executor/client/.keep
admacleod, line 1: Why do we want to keep an empty directory?

pkg/cloud/data/testresult/testresult.go
```diff
@@ -15,8 +15,6 @@ import (
 	"github.com/kubeshop/testkube/pkg/repository/testresult"
 )
 
-var _ testresult.Repository = (*CloudRepository)(nil)
```
admacleod, line 18: Does this mean that there are functions in this file that are no longer required/this entire file is no longer required?

pkg/executor/client/.keep
WitoDelnat: we're not right? This also seems outdated and not sure why

pkg/cloud/data/testresult/testresult.go
WitoDelnat: hmm, so this all likely should be removed, but I saw a good opportunity to leave this code without changing. I'm not sure how this line got removed. Perhaps I did it earlier. I've reverted this, let's see if that works 👍

WitoDelnat: I reverted this but not on the exact line of code so diff does not update. But you can check the file.

admacleod (approved):
This was a tough one because GitHub didn't like this size of diff much and JetBrains was even stumbling over parts of it but as far as I can tell this is all good deletions!

PR #6843, opened 2025-11-13
feat: replace procontext in webhooks

ed382:
The `proContext` used in webhooks is being refactored and replaced with a version that can operate in a multi-tenant environment where one will need to match events to listeners by group and be able to load listeners with the same name and kind across different groups.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -170,8 +170,8 @@ type RunnerConfig struct {
 	RunnerName                string            `envconfig:"RUNNER_NAME" default:""`
 	FloatingRunner            bool              `envconfig:"FLOATING_RUNNER" default:"false"`
 	EventLabels               map[string]string `envconfig:"EVENT_LABELS" default:""`
-	IsGlobal				  bool			    `envconfig:"RUNNER_IS_GLOBAL" default:"false"`
-	RunnerGroup				  string			`envconfig:"RUNNER_GROUP" default:""`
+	IsGlobal                  bool              `envconfig:"RUNNER_IS_GLOBAL" default:"false"`
+	RunnerGroup               string            `envconfig:"RUNNER_GROUP" default:""`
 	RunnerLabelsPrefix        string            `envconfig:"RUNNER_LABELS_PREFIX" default:"runner.testkube.io/"`
 }
 
--- a/pkg/api/v1/testkube/model_event_extended.go
+++ b/pkg/api/v1/testkube/model_event_extended.go
@@ -14,7 +14,6 @@ const (
 	TestStopSubject  = "agentevents.test.stop"
 )
 
-
 // check if Event implements model generic event type
 var _ Trigger = Event{}
 


k8s/helm/testkube-api/templates/_helpers.tpl
```diff
@@ -77,7 +77,7 @@ Define API image
 {{- $registryName := .Values.image.registry -}}
 {{- $repositoryName := .Values.image.repository -}}
 {{- $tag := default .Chart.AppVersion .Values.image.tag | toString -}}
-{{- $tagSuffix := .Values.image.tagSuffix -}}
+{{- $tagSuffix := default "" .Values.image.tagSuffix -}}
```
ed382: fixed elsewhere now

api/v1/testkube.yaml
```diff
@@ -7868,6 +7868,9 @@ components:
         streamTopic:
           type: string
           description: stream topic
+        groupId:
+          description: ID for event group
+          type: string
```
admacleod, line 7873: I'm a little unclear here as to why removing the proContext requires this change?

api/executor/v1/webhook_types.go
```diff
@@ -184,6 +184,8 @@ type Webhook struct {
 
 	Spec   WebhookSpec   `json:"spec,omitempty"`
 	Status WebhookStatus `json:"status,omitempty"`
+
+	ContextMeta map[string]string `json:"-"`
```
admacleod: I'm not sure why this change is in this PR? I can see that you're accessing this field but you're only reading it in this change. Where will it be written? Is that part of a different PR?

pkg/event/kind/webhook/loader.go
```diff
@@ -28,6 +27,11 @@ type WebhookClient interface {
 	List(selector string) (*executorv1.WebhookList, error)
 }
 
+const (
+	ContextKeyOrganizationId string = "organization-id"
+	ContextKeyEnvironmentId  string = "environment-id"
```
admacleod: Naming here is a little confusing because I normally would expect `ContextKeyXxx` in Go to be referring to the key used to store values in a `context.Context` which isn't the case here.

I'm also wondering why, if you know the structure of the data you intend to store, why you're defining `ContextMeta` as a `map[string]string` rather than as a concrete `struct`, something like:
```
type TenancyId struct {
    OrganizationId string
    EnvironmentId  string
}
```
and use the type system to your advantage?

api/executor/v1/webhook_types.go
admacleod: Found it in https://github.com/kubeshop/testkube-cloud-api/pull/2665/files#r2541102869 I see these things are sort of connected.

api/v1/testkube.yaml
ed382: Procontext is being replaced, and in the multi-tenant environment one of its fields (env id) will be used to group listeners and events. Listeners are now unique within this group (i.e. allowing multiple listeners with the same name/kind to be loaded as long as they are in different groups); then events are also emitted with this group and matched by listeners with the same group.

Named it groupId to keep it generic as other events could be unique at a different level (i.e. org).

pkg/event/kind/webhook/loader.go
ed382: > I'm also wondering why, if you know the structure of the data you intend to store, why you're defining ContextMeta as a map[string]string rather than as a concrete struct, something like:

I did not really want to add so much of cloud-specific details in the CRD structs, so I chose to go with a generic context to pass some metadata back with the clients.

> and use the type system to your advantage?

These values are just strings so the type system is still being utilized here, but I guess it is about taste when we are talking about how to name and access these variables. As mentioned above I was just trying to keep the metadata field generic and outside the CRD schema.

admacleod: Stringly typed data structures are not "using the type system". The problem with generic metadata fields is that it is far too easy for someone to start stuffing any data in there, which probably isn't what we want to happen here.

Why add this to the CRD in the first place? Is there a way you can pass this information in code without having to pass it via the CRD? As far as I can see this data is only written in the Control Plane, which has no way to make its way back into an actual Kubernetes CRD.

ed382: > Why add this to the CRD in the first place? Is there a way you can pass this information in code without having to pass it via the CRD? As far as I can see this data is only written in the Control Plane, which has no way to make its way back into an actual Kubernetes CRD.

This is being set on the CRD struct, because both the agent and control plane need to use a client with the same interface. The agent already has a client that returns back a slice of CRD structs. These CRD structs could be wrapped or extended. I chose to simply extend them with a generic metadata map representing a context not so unlike the `Annotations` field in the `metav1.ObjectMeta` defined a few lines above the `ContextMeta`, except that it does not marshal into (or out of) the actual resource definition.

admacleod: My recommendations here, pick any of them to unblock this, in descending order of preference:
1. Wrap the CRD struct in the Control Plane only, requiring no modifications to OSS code for purely Control Plane functionality.
2. Define the actual shape of `ContextMeta` using fixed fields so that any attempt to add more fields to it in the future is clear in a PR.
3. Document the purpose and intended use of `ContextMeta` so that it is _very_ clear what it is being used for.
4. Rename `ContextMeta` to something that isn't overloading multiple terms already used and clarifies its intended usage.

ed382: > Wrap the CRD struct in the Control Plane only, requiring no modifications to OSS code for purely Control Plane functionality.

I'm not sure what you mean by this, the control plane and agent implement the same interface for the client. The interface has a list method returning a struct. How would I wrap this struct only on the control plane?

admacleod: You would have to create a new interface in the Control Plane that is used on the Control Plane only, that could wrap around the implementation from OSS, and provides different data types that are enriched.

Take this (very simplified) code example of how that may look:
```
// OSS types that need to be wrapped.
type CRD struct {
	Foo string
}

type OSS interface {
	Get() CRD
}

// New wrapping types.
type EnrichedCRD struct {
	CRD
	
	Bar string
}

type ControlPlane interface {
	Get() EnrichedCRD
}

// Example implementation with static enriching data.
type Implementation struct {
	oss OSS
	bar string
}

func (i Implementation) Get() EnrichedCRD {
	crd := i.oss.Get()
	return EnrichedCRD{
		CRD: crd,
		Bar: i.bar,
	}
}
```

ed382: Maybe I'm not understanding your proposal, I have one interface for both agent and control plane implementations:

```
type WebhookClient interface {
	List(selector string) (*executorv1.WebhookList, error)
}
```

What does this look like your proposal?

admacleod:
```
type Webhook struct {
	executorv1.Webhook

	OrgId string
	EnvId string
}

type WebhookList struct {
	metav1.TypeMeta
	metav1.ListMeta
	Items []Webhook
}

type WebhookClient interface {
	List(selector string) (WebhookList, error)
}
```
you probably could just return `[]Webhook` directly because you'll be getting the items like this:
```
type Client struct {
	client webhookClient.WebhookClient
	orgId string
	envId string
}

func (c Client) List(selector string) (WebhookList, error) {
	list, _ := c.client.List(selector)
	ret := WebhookList{
		TypeMeta: list.TypeMeta,
		ListMeta: list.ListMeta,
	}
	for _, webhook := range list.Items {
		ret.Items = append(ret.Items, Webhook{
			Webhook: webhook,
			OrgId:   c.orgId,
			EnvId:   c.envId,
		})
	}
	return ret, nil
}
```

ed382: I would have to wrap the agent client just to return the wrapped struct. I think it was cleaner to add a generic context to these structs, but this is just a matter of taste.

admacleod: Thanks to Go's struct embedding you don't have to overload every function though, only the webhook listing function. This is also a reason why I keep telling everyone not to rely on these massive interfaces that are defined alongside the implementations, instead consumers should define the narrow interfaces that they want to consume, doing so would make this change much easier than it currently would be.

admacleod: Looking in the Control Plane code this is even easier than I was expecting. You don't need to wrap everything at all, all you need is to create a wrapper around the loader that the client is used in, which would be even simpler than wrapping the client or trying to pass values via the CRD.

ed382: > instead consumers should define the narrow interfaces that they want to consume

I have actually done this here.

admacleod: Please see my comment here https://github.com/kubeshop/testkube-cloud-api/pull/2665#discussion_r2542311228 for an example of a way that this change can be made with minimal code changes to this repository and without modifying the CRD schema.

PR #6849, opened 2025-11-14
fix(TKC-4536): authz agent must have gitops capability to sync

ed382:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -261,7 +261,7 @@ func main() {
 				capabilities = append(capabilities, cloud.AgentCapability_AGENT_CAPABILITY_WEBHOOKS)
 			}
 		}
-		if cfg.EnableK8sControllers || cfg.GitOpsSyncKubernetesToCloudEnabled {
+		if cfg.GitOpsSyncKubernetesToCloudEnabled {
 			capabilities = append(capabilities, cloud.AgentCapability_AGENT_CAPABILITY_GITOPS)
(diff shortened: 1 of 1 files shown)


cmd/api-server/main.go
```diff
@@ -261,6 +261,9 @@ func main() {
 				capabilities = append(capabilities, cloud.AgentCapability_AGENT_CAPABILITY_WEBHOOKS)
 			}
 		}
+		if cfg.EnableK8sControllers || cfg.GitOpsSyncKubernetesToCloudEnabled {
```
admacleod: `EnableK8sControllers` is not part of the logic for this, it is only used with an `||` in the main function because it is used to guard the initialisation of kubernetes controllers that are used for various things.

This should either:
- Check only for `cfg.GitOpsSyncKubernetesToCloudEnabled`.
- Check for all of the parameters that would affect the ability to do the sync which would be `proContext.CloudStorageSupportedInControlPlane && cfg.GitOpsSyncKubernetesToCloudEnabled && !cfg.TestkubeProTLSInsecure && !cfg.TestkubeProSkipVerify`

ed382: Addressed, kept it simple with the first option as this is just the registration.

PR #6855, opened 2025-11-17
chore: add official template wrapper client

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/k8s/helm/testkube-api/templates/deployment.yaml
+++ b/k8s/helm/testkube-api/templates/deployment.yaml
@@ -156,7 +156,7 @@ spec:
               value:  "{{ .Values.podStartTimeout }}"
             - name: CDEVENTS_TARGET
               value: "{{ .Values.cdeventsTarget }}"
-            {{- if not .Values.global.testWorkflows.createOfficialTemplate }}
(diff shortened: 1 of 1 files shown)


WitoDelnat:
hmm. Seems like the commercial control plane is implemented completely different instead of using this interface which would allow decorating it. I expected this code to be simple to reuse but that's not the case at all. Will have to dig a bit deeper into the other side as this might need additional changes.

k8s/helm/testkube-api/templates/deployment.yaml
```diff
@@ -156,6 +156,10 @@ spec:
               value:  "{{ .Values.podStartTimeout }}"
             - name: CDEVENTS_TARGET
               value: "{{ .Values.cdeventsTarget }}"
+            {{- if not .Values.global.testWorkflows.createOfficialTemplate }}
```
povilasv:
```suggestion
            {{- if not .Values.global.testWorkflows.createOfficialTemplates }}
```

Our values.yaml var is :

global.testWorkflows.createOfficialTemplates

PR #6857, opened 2025-11-17
fix(TKC-4532): secret manager missing prefixes

ed382:
## Pull request description 

It seems some of the existing secret manager methods where broken they were missing a name prefix that is added on creation of the secrets.

Merge this in first so the created secrets have the runner name:
https://github.com/kubeshop/testkube-cloud-api/pull/2669

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/pkg/secretmanager/manager.go
+++ b/pkg/secretmanager/manager.go
@@ -34,6 +34,7 @@ var (
 )
 
 type GetOptions struct {
+	// Bypass allows bypassing the permissions provided to the secret manager
 	Bypass bool
 }
 
@@ -41,12 +42,14 @@ type CreateOptions struct {
 	Type   string
 	Labels map[string]string
 	Owner  *metav1.OwnerReference
+	// Bypass allows bypassing the permissions provided to the secret manager
 	Bypass bool
 }
 
 type UpdateOptions struct {
 	Labels map[string]string
 	Owner  *metav1.OwnerReference
+	// Bypass allows bypassing the permissions provided to the secret manager
 	Bypass bool
 }
 


pkg/secretmanager/manager.go
```diff
@@ -33,6 +33,10 @@ var (
 	ErrNotControlled      = errors.New("secret is not controlled by Testkube")
 )
 
+type GetOptions struct {
+	Bypass bool
```
admacleod, line 38: Could you add a comment here explaining what the purpose of setting `Bypass` would be?

pkg/secretmanager/manager.go
```diff
@@ -126,13 +130,13 @@ func (s *secretManager) List(ctx context.Context, namespace string, all bool) ([
 	return results, nil
 }
 
-func (s *secretManager) Get(ctx context.Context, namespace, name string) (testkube.Secret, error) {
-	if !s.config.List {
+func (s *secretManager) Get(ctx context.Context, namespace, name string, opts GetOptions) (testkube.Secret, error) {
+	if !s.config.List && !opts.Bypass {
```
admacleod, line 137: Is this logic correct? As I read it, unless `Bypass` is set to `true` then secret management is disabled? So the purpose of `Bypass` is to permit us to ignore a user request to not store secrets in their cluster? I understand that this can cause issues with self-registration, but that is a user's choice to shoot themselves in the foot, overriding the user's configuration in certain edge cases seems even more unexpected than the functionality failing to work correctly.

ed382: > Is this logic correct? As I read it, unless Bypass is set to true then secret management is disabled?

Yes, it is correct (implemented just like the other preexisting methods). If bypass is `true` set we ignore the config to disable a particular action such as `List` in this case. This is how the others are implemented all that was done here is extend it to the `Get` command.

> So the purpose of Bypass is to permit us to ignore a user request to not store secrets in their cluster?

Not exactly this bypasses the config passed to the secret manager which also includes a different boolean `cfg.EnableSecretsEndpoint`:

```
	secretConfig := testkube.SecretConfig{
		Prefix:     cfg.SecretCreationPrefix,
		List:       cfg.EnableSecretsEndpoint,
		ListAll:    cfg.EnableSecretsEndpoint && cfg.EnableListingAllSecrets,
		Create:     cfg.EnableSecretsEndpoint && !cfg.DisableSecretCreation,
		Modify:     cfg.EnableSecretsEndpoint && !cfg.DisableSecretCreation,
		Delete:     cfg.EnableSecretsEndpoint && !cfg.DisableSecretCreation,
		AutoCreate: !cfg.DisableSecretCreation,
	}
	secretManager := secretmanager.New(clientset, secretConfig)
```

> I understand that this can cause issues with self-registration, but that is a user's choice to shoot themselves in the foot, overriding the user's configuration in certain edge cases seems even more unexpected than the functionality failing to work correctly.

The new registration process still abides by the configuration to avoid creating secrets, but it does not have issues with it as we register on every deployment.

In cases where it cannot create a secret, it writes a warning log saying that because secret creation is disabled the registration token will be reused in subsequent deployments which is not recommended as the registration token should probably be rotated relatively frequently.

pkg/secretmanager/manager.go
ed382: Sure, addressed here: https://github.com/kubeshop/testkube/pull/6857/commits/bc9b8623a755118350db950053fca64ce0232d4d

PR #6864, opened 2025-11-19
feat: add GetLatestExecutionsByWorkflow repository method

topliceanurazvan:
Adds support for fetching the latest execution per workflow with filtering.

- New repository interface method: `GetLatestExecutionsByWorkflow`
- MongoDB implementation with optimized aggregation pipeline
- PostgreSQL implementation with proper SQL query
- Utilizes existing index: `workflow.name + scheduledat`
- Supports all existing filters (status, tags, labels, date ranges)
- Performance tested at 1M+ executions (<20ms response time)

### DIFF
--- a/pkg/repository/testworkflow/postgres/postgres_test.go
+++ b/pkg/repository/testworkflow/postgres/postgres_test.go
@@ -54,6 +54,11 @@ func (m *MockTestWorkflowExecutionQueriesInterface) GetLatestTestWorkflowExecuti
 	return args.Get(0).([]sqlc.GetLatestTestWorkflowExecutionsByTestWorkflowsRow), args.Error(1)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/repository/testworkflow/postgres/postgres.go
```diff
@@ -2381,3 +2381,285 @@ func populateConfigParams(resolvedWorkflow *testkube.TestWorkflow, configParams
...
+			return nil, err
+		}
+
+		result[i] = *execution
+	}
+
+	return result, nil
+}
+
+const getLatestTestWorkflowExecutionsByWorkflowQuery = `-- name: GetLatestTestWorkflowExecutionsByWorkflow :many
```
vsukhin: It would be nce to move it to executions.sql and generate SQL method, it will be hard to support it in case of any changes

topliceanurazvan: Thanks, made the refactor

PR #6874, opened 2025-11-22
chore: Add cleanup of dirs created during tests

afterlook:
os package does not handle cleanup of temporary diretories, so users are expected to cleanup after they are done.

It's a non functional change, but IMO it's beneficial to clean either whatever go tmp is set to or `/tmp` that pollutes RAM space in most default installations.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vsukhin (approved):
lgtm, thank you @afterlook for your contribution

PR #6878, opened 2025-11-25
fix: [TKC-4581] Gradle and Maven - example projects restored from contrib - moved to test directory

tkonieczny:
- example gradle and maven projects restored
- workflows updated


WitoDelnat (approved):
LGTM

PR #6883, opened 2025-11-26
chore: prepare source of truth migration

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


proto/service.proto
```diff
@@ -197,7 +197,7 @@ message ProContextEnvironment {
 message ProContextAgent {
   string id = 1;
   string name = 2;
-  string type = 3;
+  string type = 3 [deprecated = true]; // Superseded by agent capabilities
```
WitoDelnat, line 200: TODO: I believe I'll need to run codegen for the generated code which is also commited afaik

povilasv (approved):

PR #6895, opened 2025-12-04
feat: super agent migration

admacleod:
This PR provides all of the necessary client side ground work to enable Agents to migrate away from Super Agents to regular Agent types.

Due to the way that it is implemented, it will not currently perform any migration actions as the Control Plane ProContext gRPC implementation does not yet return the `isSuperAgent` value.

The expected flow for a migrating Super Agent is:
- Get ProContext.
- If Control Plane supports Source of Truth, and considers Agent to still be a Super Agent, then:
  - Snapshot all syncable objects in the current Namespace.
  - Send all syncable objects from the current Namespace in a request to migrate to the Control Plane.
  - Control Plane will store the passed objects in its Source of Truth store.
  - Control Plane will mark the Agent as no longer a Super Agent.
  - Control Plane will report successful migration to the Agent.
  - Agent will restart, causing it to get new ProContext informing it that it is no longer considered to be a Super Agent.

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cloudflare/backoff"
 	"github.com/go-logr/zapr"
 	"github.com/gofiber/fiber/v2/middleware/cors"
 	"golang.org/x/sync/errgroup"
@@ -21,7 +20,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	executorv1 "github.com/kubeshop/testkube/api/executor/v1"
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -239,6 +239,7 @@ type Config struct {
 
 	FeatureCloudStorage     bool `envconfig:"FEATURE_CLOUD_STORAGE" default:"false"`
 	TestTriggerControlPlane bool `envconfig:"TEST_TRIGGER_CONTROL_PLANE" default:"false"`
+	ForceSuperAgentMode     bool `envconfig:"WARNING_UNSAFE_FORCE_SUPERAGENT_MODE" default:"false"`
```
WitoDelnat: Default to `true` while testing? That way we can merge stuff to main and other devs are not impacted while we're ironing this out.

cmd/api-server/main.go
```diff
@@ -313,6 +316,74 @@ func main() {
...
 
+	// SUPER AGENT DEPRECATION MIGRATION
+	// "Super" Agents are deprecated, instead they are being migrated to a more generic Agent with Capabilities.
+	// The migration can only occur if:
+	// - The Control Plane supports being a Source of Truth.
+	// - The current Agent is still considered to be a Super Agent by the Control Plane.
+	// - The current Agent is not being held back as a Super Agent by an override.
+	// Once Super Agent migration has been completed across all clients then this entire block can be removed.
+	if proContext.CloudStorageSupportedInControlPlane && proContext.Agent.IsSuperAgent && !cfg.ForceSuperAgentMode {
+		// Run migration in a separate goroutine so that it does not affect other operations.
```
WitoDelnat: Isn't the whole point of One-Shot migration that we do it blocking and then it's done with. What I mean with this is that we could start looking in removing superagent specific code instead of needing to keep this backwards compatible for 3-6 months.

cmd/api-server/main.go
```diff
@@ -313,6 +316,74 @@ func main() {
...
+					// On a failure log and retry with a backoff just in case.
+					retryAfter := b.Duration()
+					log.DefaultLogger.Errorw("Failed to migrate SuperAgent, will retry after backoff.",
+						"backoff", retryAfter,
+						"error", err)
+					time.Sleep(retryAfter)
+					continue
+				}
+				log.DefaultLogger.Infow("migrated super agent successfully, agent will now restart in normal agent mode.")
+				os.Exit(0)
```
WitoDelnat: Add a comment why we do this

proto/service.proto
```diff
@@ -906,3 +916,28 @@ message WebhookUpdate {
...
 }
+
+// MigrateSuperAgentRequest contains all of the migration information required for
+// a Super Agent to cede control to a Control Plane as a Source of Truth.
+// This data should be stored within the Control Plane to be later accessed by Agents
+// using the Control Plane as the Source of Truth.
+message MigrateSuperAgentRequest {
+  // test_trigger contains the following Go struct as JSON encoded data:
+  //   github.com/kubeshop/testkube/api/testtriggers/v1.TestTrigger
+  repeated bytes test_trigger = 1;
```
WitoDelnat: Interesting. I had expected that one-shot migration would call GitOps endpoints to sync them one by one. I wonder whether we will hit gRPC message limits here? There are some customers with 500s of workflows and workflows tend to get big.

cmd/api-server/main.go
admacleod: My concern is that this process could take some time, do we really want to stall out Super Agents (which in some cases are a customer's only agent) just to make our lives a bit easier?

WitoDelnat: For database migrations we also wait for completion before starting the service. As long as readiness-check does not go green the previous pod will not be killed and they notice little. I also doubt this will take longer than 2 minutes - perhaps we could benchmark with 1k Custom Resources?

proto/service.proto
WitoDelnat: If you have concerns about the length of this process, then this remark becomes more important.

admacleod: Good point, the default gRPC limit for the receiver is 4MB (there isn't a limit for sending) which is pretty low for this amount of data, especially as we're just barfing a JSON blob across rather than using defined protos which would help reduce the size somewhat. I suspect even with some compression we would probably hit 4MB quite fast.

As previously discussed we don't want to enforce the TLS of the new sync endpoints for this migration as it will delay migrations whilst some customers sort out their certificates, although if we want to block whilst migration happens anyway as per https://github.com/kubeshop/testkube/pull/6895#discussion_r2589492774 then we probably could just use this as a hard "you must configure certificates from this point onwards" cut-over.
The other alternative is to add a deprecated sync endpoint to the old mono-Service which we can simply remove once we're done?

cmd/api-server/main.go
```diff
@@ -313,6 +316,74 @@ func main() {
...
+				// WebhookTemplates
+				webhookTemplateList := executorv1.WebhookTemplateList{}
+				_ = kubeClient.List(ctx, &webhookTemplateList, client.InNamespace(cfg.TestkubeNamespace))
+				for _, t := range webhookTemplateList.Items {
+					data, _ := json.Marshal(t)
+					req.WebhookTemplate = append(req.WebhookTemplate, data)
+				}
+
+				// Send sync snapshot in request to migrate away from Super Agent.
+				if _, err := grpcClient.MigrateSuperAgent(ctx, &req); err != nil { //nolint:staticcheck // Marked as deprecated so nobody else is tempted to use it.
```
WitoDelnat: I wonder what happens if multiple agents update rather simultaneously? Do you believe there might be a need to safeguard our control plane from to many concurrent migrations?

proto/service.proto
WitoDelnat: You suggest that we automatically migrate once TLS is configured and otherwise they remain a superagent? Then they have \~2-3 months to do this which is how long we want to keep superagent code intact on agent repository? That's something to discuss with Ole and Javier. I'm not against it. 

I'm not too keen on TLS configured triggering a tricky migration - but then again, is it really so tricky? It does not do much in the end.

cmd/api-server/main.go
admacleod: I doubt it, I don't believe we have any customers running some multiplicative number of Agents and syncable resources that would cause significant load for the Control Plane. Even in our SaaS, if every customer updated all at exactly the same time, this load should be more than manageable, we can of course do some performance testing to validate this but I would be surprised if we ran into any issues.

cmd/api-server/main.go
admacleod: Ok, I'll remove the goroutine and then we can test to see if it does cause an issue and then just add it back if it is.

proto/service.proto
admacleod: Yes, in this implementation the migration is actually fairly simple and there is little to really go wrong. I do agree that it would be a little weird that configuring TLS would suddenly trigger this migration to begin, but again if we're blocking startup without migration anyway then we can simply check for TLS and report an error message and block forever.

admacleod: Ok, I've updated the implementation accordingly, this is all now much more "belt and braces" because all the syncing _must_ succeed before we can call the migration endpoint to finalise the migration, which has bloated the implementation somewhat and will definitely cause some fun loops in certain scenarios, but this has to all go through in order for the migration to be finalised.

cmd/api-server/main.go
```diff
@@ -313,6 +315,209 @@ func main() {
...
+	// - The Control Plane supports being a Source of Truth.
+	// - The current Agent is still considered to be a Super Agent by the Control Plane.
+	// - The current Agent is not being held back as a Super Agent by an override.
+	// Once Super Agent migration has been completed across all clients then this entire block can be removed.
+	if proContext.CloudStorageSupportedInControlPlane && proContext.Agent.IsSuperAgent && !cfg.ForceSuperAgentMode {
+		// If the sync store is a NoOpStore then TLS is not enabled and migration cannot progress.
+		if _, ok := syncStore.(syncagent.NoOpStore); ok {
+			log.DefaultLogger.Error("Unable to perform Super Agent migration when TLS is not configured. Please configure TLS and restart the Agent to perform migration and enable Agent functionality.")
+			// Block forever as this Agent should be migrating but is inappropriately configured to do so.
+			select {}
```
povilasv: Should we Exit(1) here instead of blocking?

I think people notice when pods are crashlooping, when they are stuck running they think everything is fine

admacleod: Thats a good question. My thinking was that I want the log message before this to be really obvious, which I often find that in a crashloop isn't necessarily the easiest thing to debug. Perhaps we could emit a Kubernetes event here as well as the log message and then die in order to make it very clear to a cluster operator what is going wrong.

admacleod: Actually I think the correct way to achieve this is to use a termination log. That is simple enough and should provide some additional context for a human operator.

admacleod: Added a write to the termination log and changed this to `os.Exit(1)`.

cmd/api-server/main.go
```diff
@@ -313,6 +315,212 @@ func main() {
 	proContext, err := commons.ReadProContext(ctx, cfg, grpcClient)
 	commons.ExitOnError("cannot connect to control plane", err)
 
+	// Configure SyncStore here as it is required for the SuperAgent migration.
```
WitoDelnat, line 316: I wonder how difficult it is to pull `MigrateSuperAgent` to a separate function? This is like 200 LoC in main

admacleod: Yeah, it is pretty chunky now. I think I'll pull it into a separate file and function so that removal can be clean and easy.

cmd/api-server/main.go
```diff
@@ -313,6 +315,212 @@ func main() {
...
+	// Once Super Agent migration has been completed across all clients then this entire block can be removed.
+	if proContext.CloudStorageSupportedInControlPlane && proContext.Agent.IsSuperAgent && !cfg.ForceSuperAgentMode {
+		// If the sync store is a NoOpStore then TLS is not enabled and migration cannot progress.
+		if _, ok := syncStore.(syncagent.NoOpStore); ok {
+			// Attempt to write to the termination log to make cluster operators' lives easier when working out why
+			// the Agent is dying. Errors here are ignored as this is a nice to have and we're about to die so there
+			// isn't any relevant error handling to perform here.
+			_ = os.WriteFile(cfg.TerminationLogPath, []byte("Insecure TLS settings configured"), os.ModePerm)
+			log.DefaultLogger.Error("Unable to perform Super Agent migration when TLS is not configured. Please configure TLS and restart the Agent to perform migration and enable Agent functionality.")
+			os.Exit(1)
```
povilasv: commons.ExitOnError(...)

admacleod: No, that function is just a wrapper around the behaviour that is already explicitly here but for people who don't like Go's error handling mechanisms:
```
func ExitOnError(title string, err error) {
	if err != nil {
		log.DefaultLogger.Errorw(title, "error", err)
		os.Exit(1)
	}
}
```

cmd/api-server/main.go
admacleod: Moved it to its own file, it should make clean-up easier.

cmd/api-server/main.go
admacleod: Actually it is different behaviour, the error log message I have does not contain an "error" field because there is no `error` to be reported. To use `commons.ExitOnError` I would have to create a non-nil error to force it to make this behaviour.

This is why I dislike these kinds of "helpers" that hide some key logic.

PR #6902, opened 2025-12-05
fix: do not enforce super agent migration when TLS is not configured

admacleod:


povilasv (approved):

cmd/api-server/superagentmigration.go
```diff
@@ -58,11 +58,15 @@ func migrateSuperAgent(ctx context.Context, log superAgentMigrationLogger, cfg s
...
+			log.Errorw("Unable to perform Super Agent migration when TLS is not configured. Please configure TLS and restart the Agent to perform migration.")
+			// Currently do not enforce TLS (do not exit the program if TLS is not configured but migration is desired).
+			// Once we want to force TLS and SuperAgent migration then this return can be removed and the logging
+			// and exiting behaviour can be used instead.
+			return
 			// Attempt to write to the termination log to make cluster operators' lives easier when working out why
 			// the Agent is dying. Errors here are ignored as this is a nice to have and we're about to die so there
 			// isn't any relevant error handling to perform here.
 			_ = os.WriteFile(cfg.terminationLogPath, []byte("Insecure TLS settings configured"), os.ModePerm)
-			log.Errorw("Unable to perform Super Agent migration when TLS is not configured. Please configure TLS and restart the Agent to perform migration and enable Agent functionality.")
```
vsukhin, line 65: Unreachable  code!

vsukhin (commented):
will not compile

cmd/api-server/superagentmigration.go
admacleod: Yes, as per the comment this is purposefully left to aid future changes. We do not have a linter that checks for unreachable code, if we did I would have added a `nolint` comment here to exclude this usage.

PR #6908, opened 2025-12-09
feat: add super agent rollback

admacleod:
Allow the ability for agents to rollback to become super agents. This is going to cause any data that was added to the Control Plane to become inaccessible, at least until the agent migrates back, but this is a disaster recovery rollback mode anyway.


cmd/api-server/superagentmigration.go
```diff
@@ -238,5 +243,29 @@ func migrateSuperAgent(ctx context.Context, log superAgentMigrationLogger, cfg s
 			log.Infow("migrated super agent successfully, agent will now restart in normal agent mode.")
 			os.Exit(0)
 		}
+	} else if !cfg.proContextAgentIsSuperAgent && cfg.forceSuperAgentMode {
+		log.Infow("Rolling back agent to super agent. This may result in data added to the Control Plane no longer being accessible!")
+		// If the agent is not currently considered to be a super agent, but the user has requested
+		// that super agent mode is forced, then rollback the super agent migration.
+		b := backoff.New(0, 0)
```
WitoDelnat, line 250: What backoff time will `New(0,0)` lead to?

WitoDelnat (approved):

cmd/api-server/superagentmigration.go
admacleod: As per [`backoff.New`](https://pkg.go.dev/github.com/cloudflare/backoff#New):
> Zero values may be used to use the default values.

[Which are](https://pkg.go.dev/github.com/cloudflare/backoff#pkg-variables):
- `DefaultInterval = 5 * time.Minute`
- `DefaultMaxDuration = 6 * time.Hour`

which I think is a _fine_ starting point.

PR #6918, opened 2025-12-11
chore: add more migration logging

WitoDelnat:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/api-server/superagentmigration.go
+++ b/cmd/api-server/superagentmigration.go
@@ -70,10 +70,14 @@ func migrateSuperAgent(ctx context.Context, log superAgentMigrationLogger, cfg s
 
 	isOrWasSuperAgent := strings.HasPrefix(cfg.agentId, "tkcroot_")
 	if !isOrWasSuperAgent || !cfg.proContextControlPlaneHasSourceOfTruthCapability {
+		log.Infow("agent is ineligible to migrate environment to source of truth",
+			"no-superagent", !isOrWasSuperAgent,
+			"no-capability", !cfg.proContextControlPlaneHasSourceOfTruthCapability)
 		return
 	}
 
 	if cfg.proContextAgentIsSuperAgent && !cfg.forceSuperAgentMode {
+		log.Infow("starting migration of super agent, agent will sync resources and afterwards restart in normal agent mode.")
 		// If the sync store is a NoOpStore then TLS is not enabled and migration cannot progress.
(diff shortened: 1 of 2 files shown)


cmd/api-server/superagentmigration.go
```diff
@@ -70,10 +70,14 @@ func migrateSuperAgent(ctx context.Context, log superAgentMigrationLogger, cfg s
 
 	isOrWasSuperAgent := strings.HasPrefix(cfg.agentId, "tkcroot_")
 	if !isOrWasSuperAgent || !cfg.proContextControlPlaneHasSourceOfTruthCapability {
+		log.Infow("agent is ineligible to migrate environment to source of truth",
+			"no-superagent", !isOrWasSuperAgent,
```
admacleod, line 74:
```suggestion
			"not-superagent", !isOrWasSuperAgent,
```

admacleod: I think this makes more sense when reading it out.

cmd/api-server/superagentmigration.go
```diff
@@ -280,5 +284,7 @@ func migrateSuperAgent(ctx context.Context, log superAgentMigrationLogger, cfg s
 			log.Infow("Rolled back super agent successfully, agent will now restart in super agent mode.")
 			os.Exit(0)
 		}
+	} else {
+		log.Infow("agent has migrated", "was_superagent", true, "requires_rollback", false)
```
admacleod, line 288: It is perhaps worth clarifying what the state is here to better structure this message:
|                    | `isSuperAgent` | `!isSuperAgent` |
|--------------------|----------------|-----------------|
| `forceSuperAgent`  |                | handled         |
| `!forceSuperAgent` | handled        |                 |

So we could have two different cases here:
- `isSuperAgent && forceSuperAgent` meaning that the agent is currently a Super Agent and will remain that way.
- `!isSuperAgent && !forceSuperAgent` meaning that the agent is not currently a Super Agent and either migrated, or never was a Super Agent.

I think then that a better way to frame this log message is:


```suggestion
		log.Infow("no agent migration will be performed, agent is remaining in current state",
			"isSuperAgent", cfg.proContextAgentIsSuperAgent,
			"forcedSuperAgent", cfg.forceSuperAgentMode,
			"wasSuperAgent", isOrWasSuperAgent)
```

PR #6919, opened 2025-12-12
fix: standalone decline execution

admacleod:
This should be no-op rather than unimplemented.

### DIFF
--- a/pkg/controlplane/agent_grpc_runner.go
+++ b/pkg/controlplane/agent_grpc_runner.go
@@ -56,6 +56,37 @@ func (s *Server) AcceptExecution(ctx context.Context, req *executionv1.AcceptExe
 	return &executionv1.AcceptExecutionResponse{}, nil
 }
 
+func (s *Server) DeclineExecution(ctx context.Context, req *executionv1.DeclineExecutionRequest) (*executionv1.DeclineExecutionResponse, error) {
+	// Running in Standalone mode so the only option here is to immediately enter an ABORTED state without passing through other transitional states.
+	execution, err := s.resultsRepository.GetWithRunner(ctx, req.GetExecutionId(), common.StandaloneRunner)
+	if err != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "execution %q could not be retrieved: %v", req.GetExecutionId(), err)
+	}
+
+	result := execution.Result
+	if result == nil {
+		// This shouldn't happen but just in case we can set something to avoid nil panics.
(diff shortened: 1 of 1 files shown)


pkg/controlplane/agent_grpc_runner.go
```diff
@@ -56,6 +56,11 @@ func (s *Server) AcceptExecution(ctx context.Context, req *executionv1.AcceptExe
 	return &executionv1.AcceptExecutionResponse{}, nil
 }
 
+func (s *Server) DeclineExecution(_ context.Context, _ *executionv1.DeclineExecutionRequest) (*executionv1.DeclineExecutionResponse, error) {
+	// Execution to remain queued so no action should be taken.
+	return &executionv1.DeclineExecutionResponse{}, nil
```
WitoDelnat, line 87: Should this not abort the execution or risk it staying in queue indefinitely? I'm having difficulty recalling when we are calling this.

admacleod: Ah yes you're right, I've just seen @tkonieczny's comment about the previous behaviour being to abort the execution.

admacleod: Ok, I've changed it so that it will immediately abort now, this has to be done mostly manually so please check this, we don't want to go via any intermediate steps as it will just get stuck in the `ABORTING` state until it gets cleaned up (which I don't think even happens here).

WitoDelnat: I'm wondering whether this needs something similar to [createCancelExecutionSteps](https://github.com/kubeshop/testkube-cloud-api/blob/af830b2228d1d51db60eaa4bf75a9090cf79f14e/internal/executor/mongo/controller_pipelines.go#L19) in [ForceCancelExecution](https://github.com/kubeshop/testkube-cloud-api/blob/e6f876e3df56d75941bea8502e0878c2cdba09ea/internal/executor/mongo/controller.go#L235)? I checked in control plane and there we also do not do this though.

FYI: We should also have access to ExecutionController  in server; but I think this is fine for now.

admacleod: We do have access to `ExecutionController` however the `Abort()` function there sends an execution into the `ABORTING` state, which needs to be explicitly avoided on the standalone executor (this case here is another example of how having things go via the gRPC server, with its own version of business logic is a bad idea).

I don't think we need to cancel all the steps because this execution is being declined, meaning it has never been executed at all.

PR #6928, opened 2025-12-18
feat: add eager assignment

WitoDelnat:
The OSS deployment now also uses a polling scheduler to work similar to the PRO offering. Currently assignments happen "Just in time" on every poll. The problem is this poll happens every 1s and therefore we are limited to 60 executions per second. Within the PRO offering, you can easily scale horizontally by adding more runners. For the OSS Standalone offering you cannot, which leads to a backlog to build up for high volume OSS (likely rather short-running) workloads.

This fix moves assignment from the poll handler to the the enqueuer. This is possible since for OSS there is only one runner which is known in advance (i.e. the standalone runner). Policies are also unsupported on this flavour and therefor do not need to be considered.

One caveat is that the previous implementation throttled requests to one execution per second, whereas now if at the clock 100 executions get started, the runner will immediately start one after the other. As far as I understand, this is also how it used to work in the past so should not cause problems.

This PR also optimised poller from five separate to a single more performant query.

https://github.com/user-attachments/assets/87123a6c-9f37-43d9-b733-48ab384c9de8

### DIFF
--- a/pkg/controlplane/agent_grpc_general.go
+++ b/pkg/controlplane/agent_grpc_general.go
@@ -260,7 +260,9 @@ func (s *Server) GetLogsStream(srv cloud.TestKubeCloudAPI_GetLogsStreamServer) e
 func (s *Server) ScheduleExecution(req *cloud.ScheduleRequest, srv cloud.TestKubeCloudAPI_ScheduleExecutionServer) error {
 	executions, err := s.enqueuer.Execute(srv.Context(), req)
 	if err != nil {
-		return status.Error(codes.Internal, "cannot enqueue execution")
+		errMsg := "cannot enqueue execution"
+		log.Errorw(errMsg, "error", err)
+		return status.Error(codes.Internal, errMsg)
 	}
 
 	for _, execution := range executions {


pkg/controlplane/scheduling/enqueuer.go
```diff
@@ -124,6 +124,12 @@ func (e *Enqueuer) Execute(ctx context.Context, req *cloud.ScheduleRequest) ([]t
 	}
 	// END POST-COMMIT VALIDATION
 
+	// Workaround
+	err = e.assignExecutions(commitContext, intermediateExecutions)
```
WitoDelnat, line 132: Below we dispatch execution events. I wanted to add dispatch an "assigned" event, but to my surprise this does not seem to exist.

pkg/controlplane/agent_grpc_execution_updates.go
```diff
@@ -109,27 +133,17 @@ func (s *Server) getNextExecution(ctx context.Context, log *zap.SugaredLogger, i
 		variableOverrides = exe.Runtime.Variables
 	}
 
-	// Mark the execution as starting
-	if err := s.ExecutionController.StartExecution(ctx, exe.Id); err != nil {
-		log.Warnw("error marking execution as starting", "err", err)
-	}
-
-	// Dispatch event for WebHooks and friends
-	s.emitter.Notify(testkube.NewEventStartTestWorkflow(&exe))
```
WitoDelnat, line 118: I believe this is a bug. We query both assigned and previously started executions. This might lead to duplicate start events. Below I handle it more correctly. This is likely also a bug in the control plane!

pkg/controlplane/agent_grpc_execution_updates.go
```diff
@@ -81,18 +81,42 @@ func (s *Server) getExecutionUpdates(ctx context.Context, log *zap.SugaredLogger
 	return updates
 }
 
-func (s *Server) getNextExecution(ctx context.Context, log *zap.SugaredLogger, info scheduling.RunnerInfo) []*executionv1.ExecutionStart {
-	// Run THE schedule query that will atomically assign one or no execution to this runner.
-	exe, ok, err := s.scheduler.ScheduleExecution(ctx, info)
-	if err != nil {
-		log.Warnw("error scheduling execution", "err", err)
-		return nil
+func (s *Server) getNextExecutions(ctx context.Context, log *zap.SugaredLogger, info scheduling.RunnerInfo) []*executionv1.ExecutionStart {
```
WitoDelnat: Polling so often becomes much less important since it can start all requested executions at once. The poll interval should still be low enough to respond promptly to transitioning events (e.g. pausing).

pkg/controlplane/agent_grpc_execution_updates.go
```diff
@@ -25,7 +25,7 @@ func (s *Server) GetExecutionUpdates(ctx context.Context, _ *executionv1.GetExec
 	// separate functions that do not error (logging is doing a lot of heavy lifting here).
 	log := log2.DefaultLogger.With("runner id", info.Id, "runner name", info.Name)
 	update := s.getExecutionUpdates(ctx, log)
-	start := s.getNextExecution(ctx, log, info)
+	start := s.getNextExecutions(ctx, log, info)
```
WitoDelnat: tbh this is very inefficient what we are doing. We are querying the database for each separate state. Instead if would be better to do one single query checking for all states of interest (pausing, resuming, aborting, canceling, starting and assigned).

PR #6932, opened 2025-12-19
fix: add mode=replace query param for TestTrigger full replacement

topliceanurazvan:
## Problem

When updating TestTriggers via the cloud dashboard's YAML Definition tab, deleting fields (like `conditionSpec`) doesn't work. The cloud-api parses the YAML but always forwards requests to the agent as JSON, so the agent's YAML Content-Type detection never triggers.

## Solution

Add `mode=replace` query parameter support. Cloud-api can pass `?mode=replace` when the original frontend request was YAML, signaling the agent to use full replacement semantics even though the forwarded content is JSON.

### DIFF
--- a/api/v1/testkube.yaml
+++ b/api/v1/testkube.yaml
@@ -263,6 +263,19 @@ paths:
       parameters:
         - $ref: "#/components/parameters/ID"
         - $ref: "#/components/parameters/Namespace"
+        - name: mode
+          in: query
+          description: |
+            Update mode for the trigger. When set to "replace", the entire trigger 
+            is replaced with the request body (full replacement semantics). When not 
+            set or set to any other value, merge semantics are used where only 
+            non-null fields from the request are applied to the existing trigger.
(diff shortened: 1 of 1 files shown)


internal/app/api/v1/testtriggers.go
```diff
@@ -83,6 +83,8 @@ func (s *TestkubeAPI) UpdateTestTriggerHandler() fiber.Handler {
 		errPrefix := "failed to update test trigger"
 		var request testkube.TestTriggerUpsertRequest
 		isYAML := string(c.Request().Header.ContentType()) == mediaTypeYAML
+		// Check for replace mode query parameter (used by cloud-api for YAML updates)
+		isReplaceMode := c.Query("mode") == "replace"
```
vsukhin, line 87: do we have it in api spec?

topliceanurazvan: Thanks for the catch Vlad! Updated

PR #6997, opened 2026-01-27
chore: add silent_mode column to postgres

florinspataru1:
filter executions fir silent mode

### DIFF
--- a/pkg/repository/testworkflow/filter.go
+++ b/pkg/repository/testworkflow/filter.go
@@ -7,26 +7,26 @@ import (
 )
 
 type FilterImpl struct {
-	FName           string
-	FNames          []string
-	FLastNDays      int
-	FStartDate      *time.Time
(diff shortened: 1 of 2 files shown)


pkg/repository/testworkflow/interface.go
```diff
@@ -91,6 +91,8 @@ type Filter interface {
 	InitializedDefined() bool
 	HealthRanges() [][2]float64
 	HealthRangesDefined() bool
+	SkipSilentMode() bool
```
dejanzele: Why do we need this?

dejanzele:
SilentMode is a boolean as far as I know, why do we need SkipSilentMode additionally?

pkg/repository/testworkflow/interface.go
florinspataru1: This is for the filter in the request so we can filter out executions with silentMode

PR #7025, opened 2026-02-09
refactor: pass request context to downstream client calls

### LINKED ISSUE #253: Pass context from requests to clients calls
**Why should it be implemented?**
Currently request when cancelled will not affect execution.

**Describe the improvement**
Pass context from fibers `c.Context()` to clients who use contexts on the end

mgazza:
## Summary

- Replace `context.Background()` with the HTTP request context (`c.Context()`) in API handlers so cancelled requests properly propagate cancellation to downstream operations
- Thread context through the `LabelSource` interface to avoid creating a disconnected background context in the labels handler
- Remove unused `"context"` import from `testworkflows.go`

Fixes #253

## Details

Several API handlers were using `context.Background()` instead of the request context when calling downstream clients. This meant that if a client cancelled an HTTP request, the underlying operations (deletes, updates, label lookups) would continue running instead of being cancelled.

The client/repository interfaces already accept `context.Context` as their first parameter, so this change simply passes through the correct context at each call site.

**Files changed:**
- `internal/app/api/v1/testworkflows.go` — 6 call sites fixed in delete, create, and update handlers
- `internal/app/api/v1/labels_tags.go` — `LabelSource` interface updated to accept context

**Not changed (intentionally):** WebSocket handlers in `testworkflowexecutions.go` correctly use `context.WithCancel(context.Background())` since WebSocket connections don't have an HTTP request context — their cancel-on-close pattern is appropriate.

## Test plan

- [ ] Verify existing tests pass
- [ ] Verify delete workflow operations respect request cancellation
- [ ] Verify label listing passes request context correctly

### DIFF
--- a/internal/app/api/v1/testworkflows.go
+++ b/internal/app/api/v1/testworkflows.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


internal/app/api/v1/testworkflows.go
```diff
@@ -78,11 +77,11 @@ func (s *TestkubeAPI) DeleteTestWorkflowHandler() fiber.Handler {
 		}
 		skipExecutions := c.Query("skipDeleteExecutions", "")
 		if skipExecutions != "true" {
-			err := s.TestWorkflowOutput.DeleteOutputByTestWorkflow(context.Background(), name)
+			err := s.TestWorkflowOutput.DeleteOutputByTestWorkflow(ctx, name)
```
povilasv: We shouldn’t propagate request context into persistence operations.
If the request is cancelled mid-delete, we can leave the system in a partially deleted, inconsistent state.

PR #7026, opened 2026-02-09
chore: add noctx linter and fix violations

mgazza:
## Summary
- Add `noctx` linter to `.golangci.yml` and fix all violations
- Fix `staticcheck` SA5011 issues (unreachable code after `ui.Failf`)

All violations of `noctx` (bare `http.Get`/`http.Post`/`http.NewRequest`, `net.Dial`/`net.Listen`, `exec.Command` without context) have been replaced with their context-aware equivalents.

The more invasive `contextcheck` linter (requires function signature changes) is tracked separately in #7043.

## Test plan
- [x] `golangci-lint run` passes with `noctx` enabled
- [x] Unit tests pass
- [x] No functional behaviour changes — only context propagation

### DIFF
--- a/cmd/kubectl-testkube/cloudlogin/login.go
+++ b/cmd/kubectl-testkube/cloudlogin/login.go
@@ -32,7 +32,7 @@ func getServerAddress(port int) string {
 
 func checkPortAvailable(port int) error {
 	addr := fmt.Sprintf("127.0.0.1:%d", port)
-	l, err := net.Listen("tcp", addr)
+	l, err := (&net.ListenConfig{}).Listen(context.Background(), "tcp", addr)
 	if err != nil {
 		return fmt.Errorf("port %d is not available: %w", port, err)
 	}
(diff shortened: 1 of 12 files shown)


cmd/kubectl-testkube/cloudlogin/login.go
```diff
@@ -253,7 +253,12 @@ func exchangeCodeForTokens(apiBaseURL, code, codeVerifier string, port int) (Tok
 		return Tokens{}, fmt.Errorf("failed to marshal request: %w", err)
 	}
 
-	resp, err := http.Post(tokenURL, "application/json", strings.NewReader(string(jsonBody)))
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodPost, tokenURL, strings.NewReader(string(jsonBody)))
```
dejanzele, line 256: Shouldn't we propagate the context from the cobra command?

The function doesn't accept it, but in this case, `http.Post` doesn't differ from `http.NewRequestWithContext` if it uses the background context

mgazza: Good point — the function signature doesn't accept a context today, so `context.Background()` is effectively what `http.Post` was using internally anyway. Threading the cobra command context through these call sites is worth doing, but I'd rather keep this PR scoped to the linter addition and tackle that as a follow-up.

dejanzele (approved):

PR #7034, opened 2026-02-10
feat: workflow tests - config override, global env - extended with non-global env case

tkonieczny:


test/special-cases/special-cases.yaml
```diff
@@ -12,20 +12,36 @@ spec:
...
-      use:
-      - name: distribute/evenly
+    service:
       image: alpine:3.22.0
       logs: always
-      count: 2
-      shell: '[ "$EXAMPLE_GLOBAL_ENV" = "GLOBAL_ENV_VALUE" ] && { echo ENV value OK; exit 0; } || { echo ENV validation failed; exit 1; }'
+      count: 1
+      shell: |
+        if [ "$EXAMPLE_GLOBAL_ENV" = "GLOBAL_ENV_VALUE" ] && \
```
mgazza: nitpick: "$EXAMPLE_GLOBAL_ENV" should be "${EXAMPLE_GLOBAL_ENV}"

PR #7043, opened 2026-02-12
chore: add contextcheck linter

mgazza:
## Summary
Enable the `contextcheck` linter and fix context propagation issues across the codebase.

This PR enables the `contextcheck` linter in `.golangci.yml` and takes a mixed approach:
- **Fixes straightforward violations** where context can be threaded through directly (e.g. replacing `context.Background()` with the available `ctx`, switching to context-aware API variants)
- **Suppresses remaining violations with `//nolint:contextcheck`** where the fix requires upstream API changes or where context is intentionally not propagated (fire-and-forget goroutines, background cleanup, APIs that don't accept context yet)

The `nolint` comments include explanations and serve as markers for follow-up work. Split from #7026 which handles `noctx`.

## What's fixed
- `cmd/api-server/commons/commons.go` — `MustGetPostgresDatabase` now uses the caller's `ctx` instead of `context.Background()`
- `pkg/mcp/tools/query.go` — switched from `jsonpath.Query` to `jsonpath.QueryWithContext`
- `pkg/runner/runner.go` — `abortExecution` now passes `ctx` to `GetExecution`
- `pkg/triggers/executor.go` — threads `ctx` through `getTestWorkflows`

## What's suppressed (for follow-up)
These require upstream API changes or are intentional design choices:
- **Telemetry calls** — `SendServerStartEvent`, `SendHeartbeatEvent` APIs don't accept context
- **Fire-and-forget goroutines** — telemetry middleware, runner request handling, monitoring reattach
- **Background cleanup** — `Destroy` and `Deploy` intentionally use `context.Background()` to complete regardless of caller cancellation
- **Watcher `cycle()` methods** — use internal context derived from `parentCtx`, opaque to the linter
- **`MustGetMinioClient`/`Abort`** — interface methods that don't accept context

## Dead code `return` after `Failf`
`cmd/kubectl-testkube/commands/agents/delete.go` and `install.go` add `return` statements after `ui.Failf()` calls. `Failf` calls `os.Exit(1)` so the `return` is unreachable at runtime, but the linter can't see through `os.Exit` and flags subsequent code as reachable without proper context. The `return` satisfies static analysis and makes control flow explicit for readers.

## CI status
- [x] Lint passes
- [x] Unit tests pass
- [x] Integration tests pass
- [x] Build passes
- [x] CodeQL passes

Related: #7025, #7026

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
     - staticcheck
     - unused
     - ineffassign
+    - contextcheck
   exclusions:
     generated: lax
     paths:
--- a/cmd/api-server/commons/commons.go
+++ b/cmd/api-server/commons/commons.go
@@ -153,7 +153,7 @@ func runMongoMigrations(ctx context.Context, db *mongo.Database) error {
 
 func MustGetMongoDatabase(ctx context.Context, cfg *config.Config, secretClient secret.Interface, migrate bool) *mongo.Database {
 	mongoSSLConfig := getMongoSSLConfig(cfg, secretClient)
-	db, err := storage.GetMongoDatabase(cfg.APIMongoDSN, cfg.APIMongoDB, cfg.APIMongoDBType, cfg.APIMongoAllowTLS, mongoSSLConfig)
+	db, err := storage.GetMongoDatabase(cfg.APIMongoDSN, cfg.APIMongoDB, cfg.APIMongoDBType, cfg.APIMongoAllowTLS, mongoSSLConfig) //nolint:contextcheck // GetMongoDatabase does not accept a context parameter
 	ExitOnError("getting mongo database", err)
 	if migrate {
 		if err = runMongoMigrations(ctx, db); err != nil {
@@ -202,7 +202,7 @@ func getMongoSSLConfig(cfg *config.Config, secretClient secret.Interface) *stora
 
 func MustGetPostgresDatabase(ctx context.Context, cfg *config.Config, migrate bool) *pgxpool.Pool {
 	// Connect to PostgreSQL
-	pool, err := pgxpool.New(context.Background(), cfg.APIPostgresDSN)
+	pool, err := pgxpool.New(ctx, cfg.APIPostgresDSN)
 	ExitOnError("Getting Postgres database", err)
 
 	if migrate {
--- a/cmd/kubectl-testkube/commands/agents/delete.go
+++ b/cmd/kubectl-testkube/commands/agents/delete.go
@@ -115,6 +115,7 @@ func UiDeleteAgent(cmd *cobra.Command, name string, uninstall, deleteAgent bool)
 		}
 		if kubernetesAgent == nil {
 			ui.Failf("kubernetes agent not found: namespaces: %s", strings.Join(nses, ", "))
+			return
 		}
 
 		spinner := ui.NewSpinner("Running Helm command...")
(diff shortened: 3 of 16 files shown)


cmd/kubectl-testkube/commands/agents/delete.go
```diff
@@ -115,6 +115,7 @@ func UiDeleteAgent(cmd *cobra.Command, name string, uninstall, deleteAgent bool)
 		}
 		if kubernetesAgent == nil {
 			ui.Failf("kubernetes agent not found: namespaces: %s", strings.Join(nses, ", "))
+			return
```
vsukhin, line 118: actually, it's reverse, we need to delete `return` after Failf, because it contains os.Exit

cmd/api-server/commons/commons.go
```diff
@@ -153,7 +153,7 @@ func runMongoMigrations(ctx context.Context, db *mongo.Database) error {
 
 func MustGetMongoDatabase(ctx context.Context, cfg *config.Config, secretClient secret.Interface, migrate bool) *mongo.Database {
 	mongoSSLConfig := getMongoSSLConfig(cfg, secretClient)
-	db, err := storage.GetMongoDatabase(cfg.APIMongoDSN, cfg.APIMongoDB, cfg.APIMongoDBType, cfg.APIMongoAllowTLS, mongoSSLConfig)
+	db, err := storage.GetMongoDatabase(cfg.APIMongoDSN, cfg.APIMongoDB, cfg.APIMongoDBType, cfg.APIMongoAllowTLS, mongoSSLConfig) //nolint:contextcheck // GetMongoDatabase does not accept a context parameter
```
vsukhin, line 156: why do we need it? code looks ugly

.golangci.yml
```diff
@@ -8,6 +8,7 @@ linters:
     - staticcheck
     - unused
     - ineffassign
+    - contextcheck
```
vsukhin, line 11: not sure about it, so many necessary nolint comments make code dirty

vsukhin (commented):
see in the code

vsukhin (changes requested):
Confidence Score: 3/5

.golangci.yml
mgazza: We should address this by fixing the linting errors; turning the linter off makes the code ugly. The nolint comments highlight the existing issues to be addressed.

cmd/kubectl-testkube/commands/agents/delete.go
mgazza: You're right that `Failf` calls `os.Exit(1)`, so the `return` is technically dead code at runtime. However, the linter can't see through `os.Exit` — without the `return`, `contextcheck` flags the subsequent code as reachable without proper context propagation. The `return` is there purely to satisfy static analysis. An alternative would be to annotate `Failf` so the linter understands it never returns (e.g. via a `//nolint` on the call site instead), but the explicit `return` makes the control flow clearer for readers too.

cmd/api-server/commons/commons.go
mgazza: `GetMongoDatabase` (in `pkg/repository/storage/mongo.go:37`) doesn't accept a `context.Context` — it creates its own `context.Background()` with a 10s timeout internally. So we can't thread context through here without changing the function signature first. The nolint comment marks this for follow-up. There's no existing PR addressing it yet, but it's a straightforward change — add a `ctx` param and use it instead of `context.Background()`. Could be part of a follow-up PR that tackles the remaining contextcheck suppressions.

mgazza:
@greptile-apps review

mgazza:
Closing — abandoning this approach in favour of addressing contextcheck violations incrementally.

PR #7046, opened 2026-02-12
fix: retry log stream on kubelet proxy errors (502 Bad Gateway)

mgazza:
## Summary
- Retry `getContainerLogsStream` when the API server proxy returns a 502 Bad Gateway reaching the kubelet
- Exponential backoff: 500ms initial, 5s max, 10 attempts before giving up
- Refactor retry loop to use `select` on `ctx.Done()` so all retries (connection lost, TLS, proxy, waiting-to-start) respect context cancellation

## Problem
Intermittent 502 errors from the kubelet proxy cause execution aborts even though the pod ran successfully. The log stream fails, the controller never receives the final instruction hints, and the execution is marked as aborted with "Job has been aborted".

Error:
```
error while fetching container logs: Get "https://<node>:10250/containerLogs/...": proxy error from 127.0.0.1:9345, code 502: Bad Gateway
```

The existing retry logic handled `connection lost` and `tls: internal error` but not proxy errors.

## Test plan
- [x] New test: `TestWatchContainerLogsProxyErrorPropagates` verifies proxy errors reach the channel
- [x] All existing controller tests pass
- [ ] Verify on cluster with intermittent 502s that executions no longer abort

### DIFF
--- a/pkg/testworkflows/executionworker/controller/logs_test.go
+++ b/pkg/testworkflows/executionworker/controller/logs_test.go
@@ -16,6 +16,16 @@ import (
 	"github.com/kubeshop/testkube/cmd/testworkflow-init/instructions"
 )
 
+const (
+	testIdleTimeoutShort   = 50 * time.Millisecond
+	testIdleTimeoutDefault = 500 * time.Millisecond
+	testDeadlineShort      = 500 * time.Millisecond
+	testDeadlineMedium     = 2 * time.Second
+	testDeadlineLong       = 5 * time.Second
+	testBufferSizeSmall    = 1
+	testBufferSizeLarge    = 5
+)
+
 func Test_ReadTimestamp_UTC_Initial(t *testing.T) {
 	reader := newTimestampReader()
 	prefix := "2024-06-07T12:41:49.037275300Z "
@@ -85,9 +95,8 @@ func TestWatchContainerLogsIdleTimeoutCancelsWhenDone(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	t.Cleanup(cancel)
 
-	idleTimeout := 50 * time.Millisecond
 	ch := watchContainerLogsWithStream(
 		ctx,
 		func(ctx context.Context, _ kubernetes.Interface, _, _, _ string, _ func() bool, _ *time.Time) (io.Reader, error) {
@@ -97,13 +106,13 @@ func TestWatchContainerLogsIdleTimeoutCancelsWhenDone(t *testing.T) {
 		"default",
 		"pod",
 		"container",
-		1,
+		testBufferSizeSmall,
 		func() bool { return true },
 		func(*instructions.Instruction) bool { return false },
-		idleTimeout,
+		testIdleTimeoutShort,
 	)
 
-	deadline := time.NewTimer(500 * time.Millisecond)
+	deadline := time.NewTimer(testDeadlineShort)
 	defer deadline.Stop()
 
 	var gotErr bool
@@ -128,7 +137,7 @@ func TestWatchContainerLogsReopensOnEOF(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	t.Cleanup(cancel)
 
 	var calls int32
 	line := "2024-06-07T12:41:49.037275300Z hello\n"
@@ -145,13 +154,13 @@ func TestWatchContainerLogsReopensOnEOF(t *testing.T) {
 		"default",
 		"pod",
 		"container",
-		5,
+		testBufferSizeLarge,
 		func() bool { return false },
 		func(*instructions.Instruction) bool { return false },
-		500*time.Millisecond,
+		testIdleTimeoutDefault,
 	)
 
-	deadline := time.NewTimer(2 * time.Second)
+	deadline := time.NewTimer(testDeadlineMedium)
 	defer deadline.Stop()
 
 	var gotLog bool
@@ -179,7 +188,7 @@ func TestWatchContainerLogsProxyErrorPropagates(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	t.Cleanup(cancel)
 
 	ch := watchContainerLogsWithStream(
 		ctx,
@@ -190,13 +199,13 @@ func TestWatchContainerLogsProxyErrorPropagates(t *testing.T) {
 		"default",
 		"pod",
 		"container",
-		5,
+		testBufferSizeLarge,
 		func() bool { return false },
 		func(*instructions.Instruction) bool { return false },
-		500*time.Millisecond,
+		testIdleTimeoutDefault,
 	)
 
-	deadline := time.NewTimer(5 * time.Second)
+	deadline := time.NewTimer(testDeadlineLong)
 	defer deadline.Stop()
 
 	var gotErr bool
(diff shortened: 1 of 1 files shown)


pkg/testworkflows/executionworker/controller/logs.go
```diff
@@ -88,33 +92,55 @@ func getContainerLogsStream(ctx context.Context, clientSet kubernetes.Interface,
 	})
 	var err error
 	var stream io.ReadCloser
+	retries := 0
```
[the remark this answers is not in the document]
mgazza: The shared counter is intentional — if the stream has failed 10 times across any combination of transient errors, the cluster connectivity is fundamentally degraded and retrying further is unlikely to help. Separate counters would allow up to 30 total retries (10 per type) which extends the failure window without much benefit.

pkg/testworkflows/executionworker/controller/logs.go
```diff
@@ -88,33 +92,46 @@ func getContainerLogsStream(ctx context.Context, clientSet kubernetes.Interface,
...
+			delay = LogRetryOnConnectionLostDelay
+		case strings.Contains(errMsg, "tls: internal error"):
+			log.DefaultLogger.Errorw("cluster's TLS error (likely CSR signing delay) while loading container logs", "pod", podName, "error", err)
+			delay = LogRetryOnConnectionLostDelay
+		case strings.Contains(errMsg, "proxy error"):
+			proxyRetries++
+			if proxyRetries > LogProxyErrorRetryMaxAttempts {
 				return nil, err
 			}
+			delay = time.Duration(proxyRetries) * LogProxyErrorRetryInitialDelay
```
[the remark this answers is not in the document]
mgazza: The backoff is exponential (bit-shift: 500ms → 1s → 2s → 4s → 5s cap). This was fixed before the last review round — see the earlier comment thread.

pkg/testworkflows/executionworker/controller/logs_test.go
```diff
@@ -174,6 +175,48 @@ func TestWatchContainerLogsReopensOnEOF(t *testing.T) {
...
+		"pod",
+		"container",
+		5,
+		func() bool { return false },
+		func(*instructions.Instruction) bool { return false },
+		500*time.Millisecond,
+	)
+
+	deadline := time.NewTimer(5 * time.Second)
+	defer deadline.Stop()
```
dejanzele, line 209: Use `t.Cleanup` instead of `defer` in tests

pkg/testworkflows/executionworker/controller/logs_test.go
```diff
@@ -174,6 +175,48 @@ func TestWatchContainerLogsReopensOnEOF(t *testing.T) {
 	}
 }
 
+func TestWatchContainerLogsProxyErrorPropagates(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
dejanzele: same

pkg/testworkflows/executionworker/controller/logs.go
```diff
@@ -88,33 +92,55 @@ func getContainerLogsStream(ctx context.Context, clientSet kubernetes.Interface,
...
-			// The container is not necessarily already started when Started event is received
-			if !strings.Contains(err.Error(), "is waiting to start") {
+			delay = LogRetryOnConnectionLostDelay
+			log.DefaultLogger.Errorw("cluster's TLS error (likely CSR signing delay) while loading container logs, retrying", "pod", podName, "attempt", retries, "error", err)
+		case strings.Contains(errMsg, "proxy error"):
+			retries++
+			if retries > LogRetryMaxAttempts {
 				return nil, err
 			}
+			delay = LogProxyErrorRetryInitialDelay << (retries - 1)
```
dejanzele, line 124: is exponential backoff only applicable for proxy error?

pkg/testworkflows/executionworker/controller/logs_test.go
```diff
@@ -174,6 +175,48 @@ func TestWatchContainerLogsReopensOnEOF(t *testing.T) {
...
+		"default",
+		"pod",
+		"container",
+		5,
+		func() bool { return false },
+		func(*instructions.Instruction) bool { return false },
+		500*time.Millisecond,
+	)
+
+	deadline := time.NewTimer(5 * time.Second)
```
dejanzele: Ideally this should be extracted as a const in the file

pkg/testworkflows/executionworker/controller/logs.go
```diff
@@ -88,33 +92,55 @@ func getContainerLogsStream(ctx context.Context, clientSet kubernetes.Interface,
...
 	var stream io.ReadCloser
+	retries := 0
 	for {
 		stream, err = req.Stream(ctx)
-		if err != nil {
-			// There may be problems with Kubernetes cluster - retry then
-			if strings.Contains(err.Error(), "connection lost") {
-				time.Sleep(LogRetryOnConnectionLostDelay)
-				continue
+		if err == nil {
```
vsukhin, line 98: such critical changes should be carefully tested in different clusters and ;providers

pkg/testworkflows/executionworker/controller/logs_test.go
mgazza: Done in 92fda82 — replaced `defer cancel()` with `t.Cleanup(cancel)` in all 4 test functions.

pkg/testworkflows/executionworker/controller/logs_test.go
mgazza: Done in 92fda82 — same fix applied across all test functions.

pkg/testworkflows/executionworker/controller/logs.go
mgazza: Yes — exponential backoff is only for proxy errors. The other retry types use fixed delays:

- **connection lost**: fixed 300ms (`LogRetryOnConnectionLostDelay`)
- **tls: internal error**: fixed 300ms (same constant)
- **proxy error**: exponential — `500ms << (retries-1)`, capped at 5s (`LogProxyErrorRetryInitialDelay` / `LogProxyErrorRetryMaxDelay`)
- **is waiting to start**: fixed 100ms (`LogRetryOnWaitingForStartDelay`), no retry counter

The rationale: proxy 502s from the kubelet are the most likely to benefit from increasing backoff (the API server proxy can be temporarily overloaded), while connection-lost/TLS errors are typically short blips that resolve within a fixed interval.

pkg/testworkflows/executionworker/controller/logs_test.go
mgazza: Done in 92fda82 — extracted test timing values and buffer sizes as named constants at the top of the file (`testIdleTimeoutShort`, `testDeadlineShort`, `testBufferSizeSmall`, etc.). We could also enable the `mnd` (magic number detector) linter in golangci-lint to catch these automatically in future.

pkg/testworkflows/executionworker/controller/logs.go
mgazza: Agreed — this warrants careful testing. So far I've validated on:

- **Local kind cluster** — verified proxy error retry with injected 502s, confirmed backoff timing and retry cap
- **Unit tests** — the `TestWatchContainerLogsProxyErrorPropagates` test covers the error propagation path through the stream opener

For broader validation across managed providers (EKS/GKE/AKS), I'd suggest we include this in the next round of cross-cluster integration testing. The change is additive (new retry path for a previously-fatal error) and guarded by the existing `LogRetryMaxAttempts` cap, so the blast radius is limited — worst case is 10 retries with backoff before failing as before.

Happy to set up a more targeted test plan if you'd like.

vsukhin: Ok, at least our GCP env is requied

mgazza: Understood — will make sure this gets validated in GCP before the next release.

PR #7063, opened 2026-02-20
feat: add stepRef and executionTab params to BuildDashboardUrl MCP tool

topliceanurazvan:
### Add step-level deep links to the dashboard URL builder MCP tool [TKC-4847]

Adds two optional parameters to the dashboard URL builder tool:

- **stepRef** — a step reference ID obtained from execution signatures, used to deep link to a specific step in the log view
- **executionTab** — controls which execution tab is shown, defaults to log-output when stepRef is provided

This allows the LLM to generate URLs that navigate directly to a specific step within an execution.

No frontend or cloud API changes are needed — the dashboard already supports these URL segments.

### What changed

- Tool definition updated with two new optional parameters and validation logic
- Tool description expanded to document the new parameters
- Added 13 unit tests covering all new and existing behavior


pkg/mcp/tools/descriptions.go
```diff
@@ -58,7 +58,7 @@ to find items containing all terms`
 	ReadArtifactDescription  = "Retrieves the content of a specific artifact from a workflow execution. This tool fetches up to 100 lines of text content from the requested file."
 
 	// Other tool descriptions
-	BuildDashboardUrlDescription  = "Build dashboard URLs for Testkube workflows and executions."
+	BuildDashboardUrlDescription  = `Build dashboard URLs for Testkube workflows and executions. Supports deep linking to a specific step within an execution's log view using stepRef (obtained from execution info signatures) and executionTab (defaults to 'log-output' when stepRef is provided). URL pattern: /organization/{orgId}/environment/{envId}/dashboard/test-workflows/{name}/execution/{execId}/{executionTab}?ref={stepRef}`
```
devcatalin: This description shouldn't include the URL pattern, that's the job of the tool so maybe the agent might not even call the tool if it sees that pattern

PR #7082, opened 2026-02-25
fix: API server panic caused by concurrent websocket writes 

dom-raven:
## Pull request description

Fix API server panic caused by concurrent websocket writes by serializing per-connection writes and making websocket listener bookkeeping concurrency-safe.

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

- None.

## Changes

- Add per-connection mutex to serialize websocket writes.
- Protect websocket list with RW mutex and iterate via snapshot.
- Use a buffered close channel and listener helper methods in the websocket loader.

## Fixes

- Prevents `panic: concurrent write to websocket connection` when multiple goroutines notify the same websocket.

### DIFF
--- a/pkg/event/kind/websocket/listener.go
+++ b/pkg/event/kind/websocket/listener.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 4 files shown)


vsukhin (approved):
great job @dom-raven lgtm. should we add addtonal unit tests for these changes?

vsukhin:
@copilot please review

PR #7084, opened 2026-02-25
feat: add failFast support for parallel workflow steps

mgazza:
## Summary

- Adds `failFast` boolean field to `StepParallel` and `IndependentStepParallel` CRD types
- When `failFast: true`, the first worker failure cancels the execution context, aborting running workers and skipping queued ones
- Defaults to `false` — fully backwards compatible, opt-in only
- Modifies `ExecuteParallel` to accept `context.Context` for cancellation support
- Updates OpenAPI spec, MCP workflow schema, CRDs, and Helm templates

Closes TKC-4992

## Target UX

```yaml
steps:
- parallel:
    failFast: true
    maxCount: 8
    shards:
      testFiles: 'glob("cypress/e2e/**/*.js")'
    run:
      args: ["--spec", "{{ shard.testFiles }}"]
```

## Design Decisions

- **Status for aborted workers**: `ABORTED` (existing handling)
- **No artifact fetching** from aborted workers
- **Queued-but-not-started workers**: skipped entirely (don't count as failures)
- **Default**: `false` (opt-in, not breaking)
- **Scope**: step-level `parallel` only (not `execute`)
- **No frontend changes**

## Test plan

- [x] `ExecuteParallel` unit tests: context cancellation, parallelism limits, skip counting, pre-cancelled context
- [x] Spec parsing tests: failFast default, serialization round-trip, normalization preservation
- [x] Dogfooding workflows: basic fail-fast, disabled default, parallelism throttling
- [x] `make lint` passes (0 issues)
- [x] `go build ./...` succeeds
- [x] Existing parallel tests pass unchanged (backwards compatibility)
- [x] Manual cluster test with fail-fast workflow


test/special-cases/fail-fast-parallel.yaml
```diff
@@ -0,0 +1,48 @@
+---
```
courageousillumination: These need the manual flag right? There's no verification of the expected behavior in the test specification.

courageousillumination (commented):
Can you share an execution that uses the fast fail flag?

test/special-cases/fail-fast-parallel.yaml
tkonieczny: - Validation should be handled for cases like this (https://github.com/kubeshop/testkube/pull/7085 - can be reused for fail-fast cases)
- special-cases/special-cases.yaml for edge-cases.yaml
- special-cases/edge-cases-expected-fails.yaml for expected failures
- labels - `core-tests: special-cases`/`core-tests: expected-fail`
- suites (`suites/special-cases/special-cases-suite.yaml` or `edge-cases-expected-fails-suite`)

mgazza: Done — commit 0a0f799 moved the tests into the standard structure:
- `special-cases-parallel-steps-fail-fast-disabled` in `special-cases.yaml` with `core-tests: special-cases`
- `expected-fail-parallel-step-fail-fast` in `edge-cases-expected-fails.yaml` with `core-tests: expected-fail`
- Both registered in their respective suite files

For validation: the fail-fast disabled case marks the step `optional: true` so the workflow passes, but doesn't assert that all workers completed. I could add a `fetch` + file-count check similar to #7085 to verify all 3 workers wrote their output. For the fail-fast enabled case (expected-fail), validation is trickier since we'd need to assert that workers were aborted — happy to add timing or output-based checks if you have a preference.

tkonieczny: I think it's fine to skip validation in `expected-fail` cases (with the current granularity). For positive cases (special-cases) there are a few options to have validation - having the step as `optional` or inverting it's result with [negative](https://docs.testkube.io/articles/test-workflows-best-practices#use-negative-true-to-assert-expected-failures).

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
```diff
@@ -490,3 +490,86 @@ func TestParallelExecution(t *testing.T) {
...
+	t.Run("failFast false with zero workers succeeds", func(t *testing.T) {
+		minimalTkConfig := `{
+			"namespace": "test",
+			"resource": {"id": "test", "root": "/tmp", "fsPrefix": "test"},
+			"workflow": {"name": "test", "labels": {}},
+			"execution": {"id": "test", "organizationId": "test", "environmentId": "test", "pvcNames": {}},
+			"controlPlane": {"url": "http://localhost:8088"},
+			"worker": {"namespace": "test", "connection": {"url": "http://localhost:8088"}}
+		}`
+		t.Setenv("TK_CFG", minimalTkConfig)
```
dejanzele: Can we avoid setting envvars in tests?

Please see how to propagate the config as a function param (some refactor might be needed)

cmd/tcl/testworkflow-toolkit/commands/services.go
```diff
@@ -433,7 +433,7 @@ func (e *ServicesExecutor) runServices(
 		return runner.Run()
 	}
 
-	return spawn.ExecuteParallel(run, instances, namespaces, int64(len(instances)))
+	return spawn.ExecuteParallel(context.Background(), run, instances, namespaces, int64(len(instances)))
```
dejanzele: Should we use Background ctx or the command's ctx?

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
```diff
@@ -490,3 +490,86 @@ func TestParallelExecution(t *testing.T) {
 		assert.NoError(t, err)
 	})
 }
+
+func TestFailFastSpecParsing(t *testing.T) {
```
dejanzele, line 490: Is it possible to make it table driven?

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
mgazza: Good point — `t.Setenv("TK_CFG", ...)` is a shortcut because `config.LoadConfigV2()` reads from the env var. I'll refactor to pass the config directly as a parameter to `RunParallelWithOptions` so the test doesn't depend on env vars.

cmd/tcl/testworkflow-toolkit/commands/services.go
mgazza: You're right — this should use the command's context so cancellation propagates when the command is cancelled. Will fix to thread the context through from the caller.

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
mgazza: Yes — will refactor to a test case slice with `tt.name`, `tt.failFast`, `tt.workers`, `tt.expectError` fields and a single `t.Run` loop.

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
mgazza: Done in 2bbd45f — added `NewConfigV2ForTest(internalCfg)` constructor that builds `ConfigV2` directly from an `InternalConfig`, bypassing the env var. Both `TestParallelExecution` and `TestFailFastSpecParsing` now use it instead of `t.Setenv`.

cmd/tcl/testworkflow-toolkit/commands/services.go
mgazza: Done in 2bbd45f — threaded `context.Context` through the full chain: `NewServicesCmd` passes `cmd.Context()` → `Execute(ctx, args)` → `runServices(ctx, ...)` → `ExecuteParallel(ctx, ...)`. No more `context.Background()` in the services path.

cmd/tcl/testworkflow-toolkit/commands/parallel_test.go
mgazza: Done in 2bbd45f — converted to `map[string]struct{}` table test with fields `spec`, `base64`, `normalize`, `execute`, `expected`. Single loop iterates all 5 cases.

test/special-cases/fail-fast-parallel.yaml
mgazza: Done — commit 8228f74 switches to `negative: true` and adds `fetch` + file validation to confirm non-failing workers completed. Thanks for the suggestion.

cmd/tcl/testworkflow-toolkit/commands/parallel.go
```diff
@@ -816,6 +825,9 @@ func executeWorkersWithStorage(ctx context.Context, workers []WorkerSpec, params
 		if err != nil {
 			fmt.Printf("%d: error: %v\n", index, err)
 		}
+		if !passed && failFast {
+			cancel()
+		}
```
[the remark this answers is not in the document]
mgazza: Good catch on the wording. Individual workers in a parallel step don't have their own `optional` flag — they all run the same spec. The step-level `optional` flag only controls whether the parallel step's failure propagates to the workflow, which is orthogonal to `failFast`. So the current behavior is correct: `failFast` triggers on any worker failure and aborts the rest. The PR description was just imprecise — will clean that up.

cmd/tcl/testworkflow-toolkit/commands/parallel.go
```diff
@@ -629,10 +629,18 @@ func (e *WorkerExecutor) processWorkerNotifications(notifications executionworke
 // handleWorkerCleanup performs cleanup operations for a completed worker.
 // Evaluates log condition to determine if logs should be saved.
 // Saves logs to artifact storage if condition is met.
-// Destroys Kubernetes resources (pods, jobs) for the worker.
-// Uses non-blocking operations to prevent hanging during cleanup.
+// If the execution context was cancelled (e.g. fail-fast), aborts the worker
```
courageousillumination, line 644: I'm not sure I structurally understand this change. 

The idea here is to trigger an abort from the clean up step for the first failed shard right? I'm concerned that this is mixing responsibilities. This should really do resource clean up for a single shard. Can we handle the abort at the same time we're canceling the context? Or maybe use the Done channel (my Go channel/context knowledge is still a bit rusty, so sorry if this one doesn't make sense).

I'm also having trouble tracking the different defers, but does this work? It looks to me like cleanup will be called before we actually do the fast fail check and cancel the context.

Can I see another execution with the new code just to sanity check?

mgazza: Good questions — let me walk through the execution order because it's subtle but intentional.

**Ordering concern: "cleanup will be called before the fail-fast check"**

You're right that for the **failing worker**, cleanup runs before `cancel()`. Here's the exact sequence:

1. Worker 0 fails → `monitorWorkerExecution` returns → deferred `handleWorkerCleanup` runs
2. At this point `ctx.Err() == nil` (cancel hasn't been called yet) → **normal `Destroy`** ✓
3. `ExecuteWorker` returns `passed=false` → `run` callback calls `cancel()`
4. Workers 1, 2 etc. see context cancelled → their notification loops exit → their deferred `handleWorkerCleanup` runs
5. Those workers see `ctx.Err() != nil` → **`Abort` (patch termination annotations + destroy)** ✓

So the failing worker gets `Destroy` (it already finished naturally) and the interrupted workers get `Abort` (they need to be stopped). The ordering is correct *because* of when the defer runs relative to `cancel()`.

**"Mixing responsibilities" concern**

The alternative — aborting workers explicitly from the `run` callback when we call `cancel()` — would create a race condition. At the point `cancel()` fires, the other workers' goroutines are already exiting and their deferred cleanups are about to run. Having two goroutines (the abort caller + the cleanup defer) both managing the same worker's lifecycle would need synchronization. The current approach avoids this because each worker's goroutine exclusively owns its own cleanup via `defer`, and checks context state to decide what to do. One goroutine, one cleanup, no races.

**Execution evidence**

I'll bring up the local cluster and share an execution screenshot shortly.

courageousillumination: > Workers 1, 2 etc. see context cancelled → their notification loops exit → their deferred handleWorkerCleanup runs

I don't think this addresses the abort concern. If before this change they were not getting aborted early (running for 30s instead of 2s), this changes doesn't help (I think that `their notification loops exit` is probably the incorrect assumption, but I'm not sure).

If you can show me the execution where fast fail runners exit after 2s instead of 30 we can move forward, but I don't think this change actually accomplishes that (of course this may be because I am still understanding the code base).

mgazza: You were right to be skeptical — when I actually ran it locally, failFast was indeed broken (workers ran ~40s instead of ~2s). The root cause was upstream: the `failFast` field was being silently dropped in the OpenAPI conversion pipeline (missing from model types, mappers, and template resolver).

All fixed in cb8bbba. Execution evidence is in the [PR comment](https://github.com/kubeshop/testkube/pull/7084#issuecomment-3985530699) — **7.959s with failFast** vs **40.351s without**. Workers are properly aborted via `ctx.Done()` in the notification loop triggering deferred `Abort()` on K8s resources.

The notification loop exit was indeed the key — it wasn't checking `ctx.Done()` before, which is why cancellation wasn't propagating. That's fixed now too.

courageousillumination (commented):
Thank you for the updates. Glad to see we're now seeing the expected behavior for fast fail. 

If I can step back for a second: we have a way of cancelling executions in progress (e.g. when a user cancels). As far as I understand this involves representing the change in the database and then waiting for the workers to poll and clean up. Why are we introducing a new mechanism?

As a note, given my relative experience with this codebase, I will not be able to give final approval to this PR. I will defer to @dejanzele who has already been involved.

pkg/tcl/testworkflowstcl/testworkflowprocessor/operations.go
```diff
@@ -142,7 +142,7 @@ func ProcessParallel(_ testworkflowprocessor.InternalProcessor, layer testworkfl
 	// Base64 encode to prevent testworkflow-init from prematurely resolving expressions.
 	// The parallel spec can contain expressions that need matrix/shard/count variables
 	// which are only available during parallel execution, not in the init context.
-	encoded, err := expressionstcl.EncodeBase64JSON(step.Parallel)
+	encoded, err := expressionstcl.EncodeBase64JSON(parallel)
```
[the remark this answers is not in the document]
mgazza: Good catch — this was indeed a pre-existing bug found during the failFast work. The commit message (cb8bbba) mentions the fix but I'll make sure it's called out in the PR description as a separate fix. Thanks for flagging it.

cmd/tcl/testworkflow-toolkit/commands/parallel.go
```diff
@@ -546,62 +546,74 @@ func (e *WorkerExecutor) monitorWorkerExecution(ctx context.Context, worker Work
...
-
-				if lastResult.IsFinished() {
-					instructions.PrintOutput(e.cfg.Ref(), "parallel", ParallelStatus{Index: int(worker.Index), Status: status, Result: v.Result})
-					return v.Result.IsPassed(), nil
+				// Get current step reference from the workflow signature
+				step := lastResult.Current(signature)
+
+				if status != prevStatus || step != prevStep {
+					if status != prevStatus {
+						log(string(status))
```
dejanzele, line 595: do we need this log?

mgazza: This log was already there before the PR — it logs worker status transitions (queued → running → passed/failed). The diff shows it because the surrounding code was re-indented when wrapping notifications in a `select` to handle context cancellation. No functional change to this line.

PR #7086, opened 2026-02-25
fix: preserve filters for GetFinished queries

courageousillumination:
## Pull request description 

This PR fixes a bug in `GetFinished` where specific filters would be dropped. This happened because `GetFinished` was assigning `query[$and]` without combining it with earlier filters (produced by `composeQueryAndOpts`).

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/pkg/repository/testworkflow/mongo/mongo.go
+++ b/pkg/repository/testworkflow/mongo/mongo.go
@@ -249,7 +249,7 @@ func (r *MongoRepository) GetFinished(ctx context.Context, filter testworkflow.F
 		opts.SetAllowDiskUse(r.allowDiskUse)
 	}
 	// Build a compound query with both status and silent mode filters
-	query["$and"] = bson.A{
+	statusAndSilentModeQuery := bson.A{
 		// Status filter: must be finished status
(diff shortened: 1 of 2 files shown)


pkg/repository/testworkflow/mongo/mongo.go
```diff
@@ -263,6 +263,11 @@ func (r *MongoRepository) GetFinished(ctx context.Context, filter testworkflow.F
 			bson.M{"silentmode": bson.M{"$exists": false}},
 		}},
 	}
+	if existingAnd, ok := query["$and"].(bson.A); ok {
```
vsukhin, line 266: do we have a smilar bug in Postgres?

courageousillumination: No, the postgres query looks like it's properly constructed https://github.com/kubeshop/testkube/blob/main/pkg/database/postgres/queries/executions.sql#L317 and we're not doing manual manipulations at the go level.

PR #7104, opened 2026-03-02
fix: use ephemeral port for control server in tests

mgazza:
## Summary
- Fix flaky `TestInitProcess` / `TestInitStateContent` / `TestInitMetrics` tests in `cmd/testworkflow-init/runner/runner_test.go`
- Tests failed intermittently with `listen tcp :60434: bind: address already in use` because the control server bound to a hardcoded port
- Change `ControlServerPort` from `const` to `var` and override to `0` (OS-assigned ephemeral port) in the test framework
- Fix goroutine leak in `control/server.go` where the `Accept` loop spun forever after `listener.Close()`

Related: #7084 (CI was failing due to this flaky test)

## Test plan
- [x] `go test -count=3 ./cmd/testworkflow-init/runner/...` passes all 3 iterations with no port conflicts
- [x] `go build ./pkg/testworkflows/executionworker/...` compiles (callers of `ControlServerPort` unaffected)
- [ ] CI green

### DIFF
--- a/cmd/testworkflow-init/control/server.go
+++ b/cmd/testworkflow-init/control/server.go
@@ -76,10 +76,9 @@ func (s *server) Listen() (func(), error) {
 	}
 	go func() {
 		for {
-			// Accept incoming connections
 			conn, err := listener.Accept()
 			if err != nil {
-				continue
+				return
 			}
 			go s.handler(conn)
 		}
--- a/cmd/testworkflow-init/constants/addr.go
+++ b/cmd/testworkflow-init/constants/addr.go
@@ -1,5 +1,3 @@
 package constants
 
-const (
-	ControlServerPort = 60434
-)
+var ControlServerPort = 60434
--- a/internal/test/framework/init.go
+++ b/internal/test/framework/init.go
@@ -254,6 +254,8 @@ func (f *InitTestFramework) configureEnvironment() error {
 }
 
 func (f *InitTestFramework) updateGlobalConstants() {
+	constants.ControlServerPort = 0
+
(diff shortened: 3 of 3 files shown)


cmd/testworkflow-init/control/server.go
```diff
@@ -76,10 +76,9 @@ func (s *server) Listen() (func(), error) {
 	}
 	go func() {
 		for {
-			// Accept incoming connections
 			conn, err := listener.Accept()
 			if err != nil {
-				continue
+				return
```
courageousillumination, line 81: Under what cases does `Accept` return an error? This seems odd to me, if we previously expected to be able to handle `accept` errors (by dropping them). but now it will kill the server?

mgazza: The `continue` → `return` change fixes a goroutine leak. The `stop()` function (line 74-76) calls `listener.Close()`, which causes `Accept()` to return `net.ErrClosed`. With the old `continue`, the goroutine would spin forever after stop is called, repeatedly hitting the same closed-listener error.

The relationship to the port change: when tests use ephemeral ports and call `stop()` to clean up between tests, the goroutine needs to actually exit — otherwise you accumulate leaked goroutines across test runs.

Transient Accept errors on a local TCP listener are effectively impossible in this context (pod-local control server), so `return` vs `continue` only matters for the listener-closed case.

courageousillumination: Thanks for the explanation.

courageousillumination (approved):

PR #7105, opened 2026-03-02
fix(mcp): return explicit message when execution metrics are unavailable

topliceanurazvan:
### What
When `get_workflow_execution_metrics` is called for an execution with no telemetry data, the API returns a stub `{"workflow":"...","execution":"..."}`. The formatter was passing this through silently, giving the AI agent no signal that metrics were unavailable.

**Changes:**
- Fixed the formatter guard (`&&` → two explicit checks) so the no-telemetry case is handled separately from truly unrecognisable input
- No-telemetry case now returns `{workflow, execution}` with an empty `steps` array — the agent can reason from the structure directly without needing embedded prose
- Cleaned up redundant struct comments

### Why
The AI agent was stalling or retrying when it received the bare stub, unable to distinguish "no data collected" from "something went wrong". The two-condition guard makes this explicit in code; the empty `steps` gives the agent a clear structural signal.

### DIFF
--- a/pkg/mcp/formatters/workflows.go
+++ b/pkg/mcp/formatters/workflows.go
@@ -299,7 +299,6 @@ const defaultMaxSamplesPerSeries = 50
 type formattedExecutionMetrics struct {
 	Workflow  string              `json:"workflow,omitempty"`
 	Execution string              `json:"execution,omitempty"`
-	Message   string              `json:"message,omitempty"`
 	Result    *formattedTimeRange `json:"result,omitempty"`
 	Steps     []formattedStepData `json:"steps,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


topliceanurazvan:
> The conditional change looks good, makes sense. I'm not sure about the message. It seems like that should be documented at the API documentation level instead of having an explicit message just for this one specific case.
> 
> It may also be less necessary with the conditional change since the agent can explicitly tell that the metrics list is empty.

Good call — removed. The two-condition guard stays, but the no-metrics branch now just returns `{workflow, execution}` with no steps. The agent can infer the situation from the empty structure; no need to embed prose in the payload.

PR #7175, opened 2026-03-05
feat: [TKC-5104] sync startup capabilities with control plane

povilasv:

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/api-server/main.go
```diff
@@ -359,6 +334,40 @@ func main() {
 		syncStore,
 	)
 
+	// Keep startup-time capabilities in sync with runtime flags.
+	// The control plane enforces that runner capability cannot be changed via this path.
+	if proContext.APIKey != "" {
```
dejanzele, line 339: should this be a separate function? it might be clearer as it is 33 lines

PR #7188, opened 2026-03-05
fix: Make NATS emitter self-healing after connection loss

dom-raven:
## Pull request description

When the NATS server restarts or a network interruption occurs, the NATS client exhausts its default reconnect attempts and the connection becomes permanently closed. After that point every `Publish()` call fails with `nats: connection closed` and the emitter never recovers — workflows keep running but the entire event stream is dead until the pod is restarted.

This PR makes the NATS emitter self-healing with no behaviour change under normal conditions.

**Error pattern before this fix:**
```
{"level":"error","caller":"event/emitter.go:120","msg":"failed to publish event","event_type":"queue-testworkflow","error":"nats: connection closed"}
{"level":"error","caller":"event/emitter.go:120","msg":"failed to publish event","event_type":"start-testworkflow","error":"nats: connection closed"}
{"level":"error","caller":"event/emitter.go:120","msg":"failed to publish event","event_type":"end-testworkflow-success","error":"nats: connection closed"}
{"level":"error","caller":"event/emitter.go:120","msg":"failed to publish event","event_type":"end-testworkflow-failed","error":"nats: connection closed"}
{"level":"error","caller":"event/emitter.go:120","msg":"failed to publish event","event_type":"end-testworkflow-not-passed","error":"nats: connection closed"}
```
All failures originate from the same dead connection. The error persists across every subsequent execution and does not self-resolve.

---

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Breaking changes

- `bus.NewNATSBus` now requires a second argument (`bus.ConnectionConfig`). All in-tree call-sites have been updated. Any out-of-tree code calling `NewNATSBus` directly will need to pass the config.

## Changes

- `pkg/event/bus/nats.go`
  - `optsFromConfig`: always sets `MaxReconnects(-1)`, `ReconnectWait(2s)`, `RetryOnFailedConnect(true)` so the NATS client retries indefinitely instead of giving up after the default 60 attempts.
  - `NewNATSConnection`: registers `DisconnectErrHandler`, `ReconnectHandler`, and `ClosedHandler` for structured log visibility into connection lifecycle events (`error_type=nats_connection_closed`).
  - `NATSBus`: stores `ConnectionConfig` and a `sync.RWMutex` to safely guard the connection pointer across goroutines.
  - `NATSBus.reconnect()`: when the connection is permanently closed, acquires a write lock, double-checks under the lock (guards concurrent callers), then dials a fresh `EncodedConn` and replaces the field.
  - `NATSBus.PublishTopic()`: acquires a read lock to snapshot the connection, publishes, and on `ErrConnectionClosed` triggers one `reconnect()` then retries. All other errors pass through unchanged.
  - `NewNATSBus`: accepts `cfg ConnectionConfig` as a second parameter.
- `cmd/api-server/main.go`: constructs `bus.ConnectionConfig` from app config and passes it to `bus.NewNATSBus`.
- `pkg/event/bus/nats_integration_test.go`: updated `NewNATSBus` call.
- `pkg/event/emitter_integration_test.go`: updated `NewNATSBus` call.

## Fixes

- Event publishing no longer fails permanently after a NATS server restart or network interruption.
- Pod restarts are no longer required to recover event delivery after a transient NATS outage.
- NATS connection lifecycle (disconnect / reconnect / permanent close) is now visible in structured logs.

### DIFF
--- a/pkg/event/bus/nats.go
+++ b/pkg/event/bus/nats.go
@@ -44,9 +44,13 @@ func optsFromConfig(cfg ConnectionConfig) (opts []nats.Option) {
 	opts = []nats.Option{
 		// Never stop trying to reconnect — the process should not require a restart
 		// due to a transient NATS outage.
+		// Note: RetryOnFailedConnect is intentionally omitted. It would make
+		// nats.Connect() return nil on an unreachable server, silently disabling
+		// the retry.DoWithData startup loop and removing crash-on-startup behaviour.
+		// MaxReconnects(-1) covers all runtime reconnection; startup retries are
+		// handled by the caller's retry loop.
 		nats.MaxReconnects(-1),
 		nats.ReconnectWait(2 * time.Second),
-		nats.RetryOnFailedConnect(true),
 	}
 
 	if cfg.NatsSecure {
@@ -172,10 +176,11 @@ func (n *NATSBus) reconnect() error {
 	if err != nil {
 		return fmt.Errorf("nats reconnect failed: %w", err)
 	}
-	n.nc = conn
 
-	// Re-register every subscription on the new connection.  Without this step
-	// all consumers would silently stop receiving events after a reconnect.
+	// Re-register subscriptions BEFORE exposing conn via n.nc.  This closes the
+	// window where the new connection is live but has no handlers — messages
+	// published to subscribed topics during that gap would otherwise be silently
+	// discarded by the NATS server.
 	n.subscriptions.Range(func(key, value any) bool {
 		entry := value.(*subscriptionEntry)
 
@@ -202,6 +207,7 @@ func (n *NATSBus) reconnect() error {
 		return true
 	})
 
+	n.nc = conn
 	return nil
 }
 
--- a/pkg/event/bus/nats.go
+++ b/pkg/event/bus/nats.go
@@ -28,6 +28,9 @@ const (
 	SubscriptionName       = "agentevents"
 	InternalPublishTopic   = "internal.all"
 	InternalSubscribeTopic = "internal.>"
+
+	natsMaxReconnects = -1
+	natsReconnectWait = 2 * time.Second
 )
 
 type ConnectionConfig struct {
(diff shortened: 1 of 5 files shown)


vsukhin:
@greptiile

dom-raven:
If greptile is happy, i'll run it on my dev cluster too

dom-raven:
@greptile

pkg/event/bus/nats.go
```diff
@@ -40,7 +41,18 @@ type ConnectionConfig struct {
...
+	opts = []nats.Option{
+		// Never stop trying to reconnect — the process should not require a restart
+		// due to a transient NATS outage.
+		// Note: RetryOnFailedConnect is intentionally omitted. It would make
+		// nats.Connect() return nil on an unreachable server, silently disabling
+		// the retry.DoWithData startup loop and removing crash-on-startup behaviour.
+		// MaxReconnects(-1) covers all runtime reconnection; startup retries are
+		// handled by the caller's retry loop.
+		nats.MaxReconnects(-1),
+		nats.ReconnectWait(2 * time.Second),
```
vsukhin: should be const

pkg/event/bus/nats.go
```diff
@@ -166,14 +348,29 @@ func (n *NATSBus) queueName(subscription, queue string) string {
 }
 
 func (n *NATSBus) TraceEvents() {
-	s, err := n.nc.Subscribe(SubscriptionName+".>", func(event testkube.Event) {
+	n.mu.RLock()
+	nc := n.nc
+	n.mu.RUnlock()
```
vsukhin: might be 3 line helper method updateNC()

pkg/event/bus/nats.go
```diff
@@ -94,22 +113,35 @@ func NewNATSConnection(cfg ConnectionConfig, opts ...nats.Option) (*nats.Conn, e
...
+	queue   string // empty → plain Subscribe, non-empty → QueueSubscribe
+	handler Handler
+	sub     *nats.Subscription
+}
+
 type NATSBus struct {
 	nc            *nats.EncodedConn
-	subscriptions sync.Map
+	cfg           ConnectionConfig
+	mu            sync.RWMutex
```
vsukhin: connMu ?

pkg/event/bus/nats.go
```diff
@@ -122,42 +154,192 @@ func (n *NATSBus) Subscribe(queueName string, handler Handler) error {
...
+	n.mu.RUnlock()
+
+	err := nc.Publish(topic, event)
+	if err == nil {
+		return nil
+	}
+
+	if !errors.Is(err, nats.ErrConnectionClosed) {
+		return err
+	}
```
vsukhin, line 274: you can combine 2 if

vsukhin (commented):
some notes

PR #7214, opened 2026-03-06
feat: include install crds for gitops runners

caiomede-tk:
(No changelog changes)


caiomede-tk:
Recreating PR because of mess with conflicts from main

PR #7347, opened 2026-03-16
feat: add timeouts to test workflow definition 

courageousillumination:
Currently timeouts will support a queue timeout and initialization timeout. This PR just covers the definition changes and does not implement these timeouts.


pkg/api/v1/testkube/model_test_workflow_execution_schema.go
```diff
@@ -13,5 +13,6 @@ package testkube
 type TestWorkflowExecutionSchema struct {
 	Tags   map[string]string `json:"tags,omitempty"`
 	Target *ExecutionTarget  `json:"target,omitempty"`
```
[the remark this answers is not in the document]
courageousillumination: Reverted. Filed a bug though, because this code is part of the auto generation: https://linear.app/kubeshop/issue/TKC-5200/running-generate-openapi-leaves-code-in-uncompilable-state

api/testworkflows/v1/base_types.go
```diff
@@ -33,6 +33,17 @@ type TestWorkflowSpecBase struct {
...
 	// values to be used for test workflow execution
 	Execution *TestWorkflowExecutionSchema `json:"execution,omitempty" expr:"include"`
+
+	// per-workflow timeout configuration
+	Timeouts *TestWorkflowTimeouts `json:"timeouts,omitempty" expr:"include"`
+}
+
+// TestWorkflowTimeouts configures timeout behavior for a workflow (queue, initialization).
+type TestWorkflowTimeouts struct {
+	// maximum time the execution may spend in queue before starting (not implemented)
```
dejanzele: nit: I'd add a `// TODO` saying not implemented as it might be a better pattern

api/v1/testkube.yaml
```diff
@@ -9500,6 +9500,17 @@ components:
         isolatedContainers:
           $ref: "#/components/schemas/BoxedBoolean"
 
+    TestWorkflowTimeouts:
+      type: object
+      description: configures timeout behavior for a workflow (queue, initialization)
+      properties:
+        queue:
+          type: string
```
dejanzele, line 9477: Some OpenAPI codegens support `format: duration` for `type: string` and give you additional validation.

OpenAPI docs mention it is an open field and does not have a strict standard for `format`, might be worthwhile to check.

pkg/api/v1/testkube/model_event_type.go
```diff
@@ -13,6 +13,26 @@ type EventType string
 
 // List of EventType
 const (
+	START_TEST_EventType                     EventType = "start-test"
```
dejanzele: I think the `test` and `testsuite` events are deprecated, and somebody forgot to remove them from OpenAPI but instead removed them just from the autogenerated files.

I think the following change from L16-L35 can be removed, but it should be done via the OpenAPI schema and regenerated.

@olensmar can you confirm we have fully deprecated and removed legacy `tests` and `testsuites`?

pkg/api/v1/testkube/model_features.go
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Testkube provides a Kubernetes-native framework for test definition, execution and results
+ *
+ * API version: 1.0.0
+ * Contact: contact@testkube.io
+ * Generated by: Swagger Codegen (https://github.com/swagger-api/swagger-codegen.git)
+ */
+package testkube
+
+type Features struct {
```
dejanzele: I think this is also a leftover which needs to be removed in OpenAPI and the regeneration should delete the file.

pkg/api/v1/testkube/model_server_info.go
```diff
@@ -34,6 +34,7 @@ type ServerInfo struct {
 	// disable secret creation for tests and test sources
 	DisableSecretCreation bool          `json:"disableSecretCreation,omitempty"`
 	Secret                *SecretConfig `json:"secret,omitempty"`
+	Features              *Features     `json:"features,omitempty"`
```
dejanzele: same

pkg/api/v1/testkube/model_event_type.go
olensmar: confirmed!

api/testworkflows/v1/base_types.go
courageousillumination: Updated.

api/v1/testkube.yaml
courageousillumination: Our other durations don't use the format, so I think I'm going to stick with our existing patterns for now.

pkg/api/v1/testkube/model_event_type.go
courageousillumination: Removed in OpenAPI

pkg/api/v1/testkube/model_features.go
courageousillumination: Removed in OpenAPI

pkg/api/v1/testkube/model_server_info.go
courageousillumination: Removed in OpenAPI

PR #7349, opened 2026-03-17
chore(refactor): [TKC-5105] testkube-runner Helm chart dependencies and add CRD support

caiomede-tk:
## Pull request description 

- Removed the deprecated testkube-operator dependency from Chart.lock and Chart.yaml.
- Added testkube-crds as a new dependency in both Chart.lock and Chart.yaml.
- Introduced a new crds.yaml template to conditionally render CRDs based on gitops settings.
- Updated values.yaml to include a flag for installing shared Testkube CRDs.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -428,15 +428,7 @@ generate-crds: ## Generate Kubernetes CRDs from kubebuilder Golang structs.
 		for key in securityContext volumes dnsPolicy affinity tolerations hostAliases dnsConfig topologySpreadConstraints schedulingGates resourceClaims imagePullSecrets volumeMounts fieldRef resourceFieldRef configMapKeyRef secretKeyRef pvcs matchExpressions matchLabels env envFrom fileKeyRef readinessProbe; do \
 			go tool yq --no-colors -i "del(.. | select(has(\"$$key\")).$$key | .. | select(has(\"description\")).description)" "k8s/crd/$$file"; \
 		done; \
-		go tool yq --no-colors -i \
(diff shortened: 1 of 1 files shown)


k8s/helm/testkube-operator/templates/crds.yaml
```diff
@@ -0,0 +1,3 @@
+{{- if .Values.installCRD }}
+{{ include "testkube-crds.render" . }}
```
topliceanurazvan, line 2: Did you verify the rendered output with `helm template` on both charts? 
Just want to make sure the CRD documents render as expected after the refactoring.

.github/workflows/test.yaml
```diff
@@ -73,7 +73,7 @@ jobs:
 
   integration-tests:
     name: Integration Tests
-    runs-on: depot-ubuntu-22.04
+    runs-on: ubuntu-22.04
```
topliceanurazvan, line 76: Was the runner change from `depot-ubuntu-22.04` to `ubuntu-22.04` intentional? 
Depot runners are typically faster, so this could slow down CI.

k8s/helm/testkube-operator/templates/crds.yaml
caiomede-tk: Yes, it does, I double checked and it's working as expected

.github/workflows/test.yaml
caiomede-tk: Yes, it was intentional because Depot is better for docker builds, but there is no clear benefit for other actions, I'm comparing the CI execution times to be sure about this

PR #7352, opened 2026-03-17
fix: [TKC-4934] remove hardcoded cloud workflow list limit

povilasv:
## Summary
- remove hardcoded `Limit: 100` in cloud `ListTestWorkflows` request
- forward `options.Limit` so cloud listing is not capped at 100 by default
- add regression tests for limit forwarding and zero-limit behavior

## Validation
- `go test ./pkg/controlplaneclient` passed
- `go test ./...` fails due to pre-existing unrelated issue in `api/testtriggers/v1` (Ginkgo decorator)
- `make lint-go` / `make lint-openapi` targets are not available in this repository

## Context
Customer reported `testkube get tw -o json` in cloud mode returns only 100 workflows while Kubernetes context returns the full count. This change removes the cloud client-side cap.

### DIFF
--- a/pkg/controlplaneclient/testworkflows.go
+++ b/pkg/controlplaneclient/testworkflows.go
@@ -56,7 +56,7 @@ func (c *client) GetTestWorkflow(ctx context.Context, environmentId, name string
 func (c *client) ListTestWorkflows(ctx context.Context, environmentId string, options ListTestWorkflowOptions) TestWorkflowsReader {
 	req := &cloud.ListTestWorkflowsRequest{
 		Offset:     options.Offset,
-		Limit:      100,
+		Limit:      options.Limit,
 		Labels:     options.Labels,
 		TextSearch: options.TextSearch,
 	}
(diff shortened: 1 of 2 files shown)


vsukhin (approved):

pkg/controlplaneclient/testworkflows.go
```diff
@@ -56,7 +56,7 @@ func (c *client) GetTestWorkflow(ctx context.Context, environmentId, name string
 func (c *client) ListTestWorkflows(ctx context.Context, environmentId string, options ListTestWorkflowOptions) TestWorkflowsReader {
 	req := &cloud.ListTestWorkflowsRequest{
 		Offset:     options.Offset,
-		Limit:      100,
+		Limit:      options.Limit,
```
caiomede-tk, line 59: Just a question, as this is not a param that was in use, should not we have a default value to avoid queries bringing by default all values which in some cases can be a long list?

povilasv: So this is a list call; by default it should return all testworkflows. Otherwise it doesn't really make sense, since we dont support any pagination or stuff like that.

caiomede-tk: Got it, if that is the normal behavior let's proceed, I was just concerned if that can generates a performance problem on those instances with high load and long lists of TWs, but I think our envs are already testing that

PR #7358, opened 2026-03-18
feat: add id field to step spec and propagate to runtime

dejanzele:
## Summary

- Add `id` field to `StepMeta` CRD type, OpenAPI spec, and API models
- Auto-derive id from step name when not explicitly set (lowercase, non-alphanumeric chars become underscores)
- Duplicate auto-derived IDs get an index suffix: `build`, `build_1`, `build_2`
- Validate explicit IDs: lowercase alphanumeric + underscores, unique within workflow
- Propagate id through stage metadata, signature, action declaration, and serialization layers
- Update all mappers (Kube <-> OpenAPI)
- Add `ValidateExplicitStepIds` for API-time validation (wired in cloud-api via TKC-5249)

## Test plan

- [x] Explicit id preserved in signature
- [x] Name auto-derives to valid id (including digit-starting names like "1st step")
- [x] Duplicate auto-derived IDs get index suffixes
- [x] Auto-derived conflict with explicit ID gets suffix
- [x] Invalid format rejected for explicit IDs
- [x] Unicode names skipped (non-ASCII fails regex)
- [x] Cross-section uniqueness enforced (setup vs steps)
- [x] ValidateExplicitStepIds: parallel blocks validated independently
- [x] ValidateExplicitStepIds: steps without id skipped
- [x] Existing processor and preset tests pass

Closes TKC-5165

### DIFF
--- a/pkg/testworkflows/testworkflowprocessor/processor.go
+++ b/pkg/testworkflows/testworkflowprocessor/processor.go
@@ -65,6 +65,7 @@ func (p *processor) process(layer Intermediate, container stage.Container, step 
 	// Build an initial group for the inner items
 	self := stage.NewGroupStage(ref, false)
 	self.SetPure(step.Pure)
+	self.SetId(step.Id)
 	self.SetName(step.Name)
 	self.SetOptional(step.Optional).SetNegative(step.Negative).SetTimeout(step.Timeout).SetPaused(step.Paused)
 	if step.Condition == "" {
@@ -143,7 +144,19 @@ func (p *processor) Bundle(ctx context.Context, workflow *testworkflowsv1.TestWo
 		return nil, errors.New("could not resolve resource.root")
 	}
 
-	// Process steps
+	err = expressions.Simplify(&workflow, machines...)
+	if err != nil {
+		return nil, errors.Wrap(err, "error while simplifying workflow instructions")
+	}
+
+	// Resolve and validate step IDs after expression simplification,
+	// so auto-derivation works on resolved names rather than raw templates.
+	if err := ResolveAndValidateStepIds(&workflow.Spec); err != nil {
+		return nil, errors.Wrap(err, "step id validation")
+	}
+
+	// Process steps - must be after ResolveAndValidateStepIds so rootStep
+	// picks up the resolved IDs from the spec.
 	rootStep := testworkflowsv1.Step{
 		StepSource: testworkflowsv1.StepSource{
 			Content: workflow.Spec.Content,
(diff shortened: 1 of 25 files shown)


pkg/testworkflows/testworkflowprocessor/processor.go
```diff
@@ -143,6 +144,11 @@ func (p *processor) Bundle(ctx context.Context, workflow *testworkflowsv1.TestWo
 		return nil, errors.New("could not resolve resource.root")
 	}
 
+	// Validate and resolve step IDs
+	if err := ValidateStepIds(&workflow.Spec); err != nil {
+		return nil, errors.Wrap(err, "step id validation")
+	}
```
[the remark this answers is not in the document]
dejanzele: resolved

dejanzele:
@greptile

dejanzele:
@greptile

pkg/testworkflows/testworkflowprocessor/processor.go
```diff
@@ -143,7 +144,19 @@ func (p *processor) Bundle(ctx context.Context, workflow *testworkflowsv1.TestWo
 		return nil, errors.New("could not resolve resource.root")
 	}
 
-	// Process steps
+	err = expressions.Simplify(&workflow, machines...)
```
courageousillumination, line 147: Just verifying, it's fine to run this before setting up the root step below? I think it's fine, but want to confirm.

pkg/testworkflows/testworkflowprocessor/stepid.go
```diff
@@ -0,0 +1,104 @@
...
+			b.WriteRune(unicode.ToLower(r))
+			prevUnderscore = false
+		} else if !prevUnderscore {
+			b.WriteByte('_')
+			prevUnderscore = true
+		}
+	}
+	result := strings.Trim(b.String(), "_")
+	if result == "" || (result[0] >= '0' && result[0] <= '9') {
+		return ""
```
courageousillumination, line 44: Should we do _something_ in the case of 0-9? I'm thinking something like prefix with `step_id`, but maybe that will be more confusing than useful.

pkg/testworkflows/testworkflowprocessor/processor.go
dejanzele: Yes, that is intentional, as in case of auto-generated `id` field from `name` field, we want first to make sure the `name` field is resolved, in case it used any templating function

pkg/testworkflows/testworkflowprocessor/stepid.go
dejanzele: I looked again on the autogen part and now this cases are valid:
1. Step name starts with digit like "1 step", it will get generated as `1_step`
2. Step name conflicts for some reason with autogenerated or explicit ID, the current autogenerated one will get a index suffix like `_1`, `_2`...

Cases where autogenerated will be empty and fallback to the internal step ref (autogenerated at runtime like `rzzzvjsf`):
1. Name uses non-ASCII characters like `name: Résumé`
2. Name field is empty
3. Name field uses only special characters like `###`

PR #7424, opened 2026-03-26
feat(TKC-5278): add workflow and resource group credential cases

Valentin-Marko:
## Pull request description 

Extends the special-cases-dev-credentials workflow to cover workflow-level and resource group credentials.

Adds validation for both plaintext and encrypted credentials across these scopes, ensuring they are correctly resolved via:

- environment variables
- template-based access ({{ credential(...) }})

This complements existing org and environment credential test cases and improves coverage for credential resolution and precedence behavior.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

none

## Changes

- extend special-cases-dev-credentials workflow
- add workflow-level credential test cases (plaintext + encrypted)
- add resource group credential test cases (plaintext + encrypted)
- validate resolution using both env variables and {{ credential(...) }}
- align test coverage with credential precedence expectations

## Fixes

-

### DIFF
--- a/test/special-cases/special-cases-dev.yaml
+++ b/test/special-cases/special-cases-dev.yaml
@@ -28,6 +28,18 @@ spec:
       - name: credential_encrypted_org_level_env_override
         value: '{{ credential("credential-encrypted-org-level-env-override") }}'
         global: true
+      - name: credential_plaintext_workflow_level
+        value: '{{ credential("credential-plaintext-workflow-level") }}'
+        global: true
+      - name: credential_encrypted_workflow_level
+        value: '{{ credential("credential-encrypted-workflow-level") }}'
+        global: true
+      - name: credential_plaintext_resource_group
+        value: '{{ credential("credential-plaintext-resource-group") }}'
+        global: true
+      - name: credential_encrypted_resource_group
+        value: '{{ credential("credential-encrypted-resource-group") }}'
+        global: true
   services:
     echo-service:
       image: alpine:3.22.0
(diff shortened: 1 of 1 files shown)


test/special-cases/special-cases-dev.yaml
```diff
@@ -28,6 +28,18 @@ spec:
       - name: credential_encrypted_org_level_env_override
         value: '{{ credential("credential-encrypted-org-level-env-override") }}'
         global: true
+      - name: credential_plaintext_workflow_level
+        value: '{{ credential("credential-plaintext-workflow-level") }}'
```
tkonieczny, line 32: these are just ENVs, not credentials

test/special-cases/special-cases-dev.yaml
```diff
@@ -28,6 +28,18 @@ spec:
...
+        global: true
+      - name: credential_encrypted_workflow_level
+        value: '{{ credential("credential-encrypted-workflow-level") }}'
```
tkonieczny, line 35: there is no credential named this way, and the structure doesn't fit workflow-level - they are created with workflow name prefix

test/special-cases/special-cases-dev.yaml
```diff
@@ -28,6 +28,18 @@ spec:
...
+        global: true
+      - name: credential_plaintext_resource_group
+        value: '{{ credential("credential-plaintext-resource-group") }}'
+        global: true
+      - name: credential_encrypted_resource_group
+        value: '{{ credential("credential-encrypted-resource-group") }}'
```
tkonieczny, line 41: There is no RG and credentials

PR #7490, opened 2026-04-08
fix: agent artifact route to support filenames with slashes

devcatalin:
Companion fix to kubeshop/testkube-cloud-api#3489 - resolves the second layer of the artifact 410/408 bug.

- Artifact filenames can contain subdirectory paths (e.g. `results/junit.xml`); the cloud API proxies these to the agent via NATS with `%2F`-encoded slashes
- fasthttp's `URI.Parse` decodes `%2F` → `/` before Fiber routing, so `:filename` (single-segment) can't match multi-segment decoded paths — the agent silently drops the request, causing a 408 timeout at the cloud API
- Change `:filename` → `:filename+` (Fiber greedy param) so the route matches filenames with slashes, matching the cloud API fix (`{filename:.*}`)


olensmar (approved):

courageousillumination:

PR #7496, opened 2026-04-09
feat(TKC-5426): Dashboard Execution Viewer

Valentin-Marko:
## Pull request description 

Add execution viewer support to the testkube view command. When invoked with an execution ID (`testkube view <executionId>`), the behaviour depends on context:

Cloud / Control Plane context: fetches the execution and opens the existing dashboard URL directly in the browser —> no data is uploaded.
Standalone (kubeconfig) context: fetches the execution metadata, logs, and artifacts, bundles them into a multipart upload, sends them to the shares API, and opens the public viewer URL in the browser.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- Add public execution sharing via `testkube view <executionId>`
- New `view.go` with multipart upload to shares API (execution JSON, logs, artifacts)
- Client-side bundle size check (10MB) with fail-fast during artifact download
- Add `--shares-api-url`, `--viewer-url`, `--skip-artifacts`, `--force`  flags to view 

## Fixes

-


cmd/kubectl-testkube/commands/dashboard_share.go
```diff
@@ -0,0 +1,215 @@
...
+
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
+	"github.com/kubeshop/testkube/pkg/telemetry"
+	"github.com/kubeshop/testkube/pkg/ui"
+)
+
+const (
+	maxBundleSize      = 10 * 1024 * 1024
+	shareUploadTimeout = 60 * time.Second
+)
```
vsukhin: this should be on server side, client just need to  handle timeout properly

cmd/kubectl-testkube/commands/view.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func shareAndOpenExecution(cmd *cobra.Command, executionID, sharesAPIURL, viewerBaseURL string, skipArtifacts bool) {
+	uris := common.NewMasterUris("", "", "", "", "", "", "", false)
+	if sharesAPIURL == "" {
+		sharesAPIURL = uris.Api
+	}
+	if viewerBaseURL == "" {
+		viewerBaseURL = uris.Ui
+	}
```
vsukhin, line 86: guess, they should be separate flags, living on separate uris

cmd/kubectl-testkube/commands/view.go
```diff
@@ -0,0 +1,215 @@
...
+
+	mustMarshal := func(v interface{}) json.RawMessage {
+		b, err := json.Marshal(v)
+		if err != nil {
+			ui.ExitOnError("marshaling metadata field", err)
+		}
+		return b
+	}
+
+	payload["executionId"] = mustMarshal(execution.Id)
```
vsukhin, line 123: execution id is a string like 6789...., why do you need to marshal it?

cmd/kubectl-testkube/commands/view.go
```diff
@@ -0,0 +1,215 @@
...
+	if execution.Workflow != nil {
+		payload["workflowName"] = mustMarshal(execution.Workflow.Name)
+	}
+	if execution.Result != nil && execution.Result.Status != nil {
+		payload["status"] = mustMarshal(string(*execution.Result.Status))
+	}
+	payload["installationId"] = mustMarshal(installationID)
```
vsukhin, line 130: same question for other string fields

vsukhin (commented):
have a ew questions

cmd/kubectl-testkube/commands/dashboard_share.go
Valentin-Marko: Good point. I initially added the client-side size check for better UX: to fail fast during artifact download rather than waiting for the full upload to get rejected. But on the backend the maxBundleSize is an env var, so if it changes the hardcoded client value will get out of sync. I'll remove it and let the server handle the enforcement.

cmd/kubectl-testkube/commands/view.go
Valentin-Marko: They are separate:  --shares-api-url defaults to uris.Api (api.testkube.io) and --viewer-url defaults to uris.Ui (app.testkube.io). Two different flags, two different URIs.

cmd/kubectl-testkube/commands/view.go
Valentin-Marko: The payload uses map[string]json.RawMessage to preserve the original execution fields as raw bytes. I'm not sure we'd actually hit float64 precision loss in our case, but RawMessage keeps things safe. The tradeoff is that injected strings need json.Marshal to add JSON quotes.

vsukhin (approved):

Valentin-Marko:

Valentin-Marko:
@greptileai can you please review again?

cmd/kubectl-testkube/commands/view.go
```diff
@@ -0,0 +1,347 @@
...
+	ui.ExitOnError("getting execution", err)
+
+	workflowName := ""
+	if execution.Workflow != nil {
+		workflowName = execution.Workflow.Name
+	}
+
+	uri := fmt.Sprintf(
+		"%s/organization/%s/environment/%s/dashboard/test-workflows/%s/execution/%s",
+		cfg.CloudContext.UiUri,
```
olensmar, line 356: not sure UiUri will always be populated!? - perhaps check first and show an error if not? (and how to set it)

Valentin-Marko: Good catch. Added a guard at the top of `openCloudExecution` (before any API calls) that checks `uiUri`, `organizationId`, and `environmentId` and tells the user exactly which fields are missing and how to set them: via `testkube login` or `testkube set context --org-id <id> --env-id <id>` (with --ui-uri-override mentioned only when uiUri itself is missing).

PR #7497, opened 2026-04-09
feat: [TKC-4032] simplify lookup_execution_id response parsing

mgazza:
## Summary

Simplifies the `lookup_execution_id` MCP tool to parse the new direct `{"id": "..."}` response format from the dedicated lookup endpoint, with backwards-compatible fallback for the legacy list format.

## Changes

- `extractExecutionIdFromResponse` now tries the new direct `{"id": "..."}` format first
- Falls back to legacy `{"results": [...]}` list format for backwards compatibility (standalone CLI still uses this)
- `isValidExecutionName` now rejects names with empty prefix (e.g., `"-123"`)

## Companion PR

- testkube-cloud-api: https://github.com/kubeshop/testkube-cloud-api/pull/3510

## Test plan

- [x] `go build ./pkg/mcp/...` clean
- [x] `go test ./pkg/mcp/...` all pass
- [ ] Manual testing with both new endpoint and legacy fallback

### DIFF
--- a/pkg/mcp/tools/executions.go
+++ b/pkg/mcp/tools/executions.go
@@ -247,7 +247,7 @@ func LookupExecutionId(client ExecutionLookup) (tool mcp.Tool, handler server.To
 			return mcp.NewToolResultError(fmt.Sprintf("failed to lookup execution ID: %v", err)), nil
 		}
 
-		executionID, err := extractExecutionIdFromResponse(result)
+		executionID, err := extractExecutionIdFromResponse(result, executionName)
 		if err != nil {
 			return mcp.NewToolResultError(err.Error()), nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/mcp/tools/executions.go
```diff
@@ -269,37 +269,43 @@ func isValidExecutionName(executionName string) bool {
...
 
-	for _, result := range results {
-		if execution, ok := result.(map[string]any); ok {
-			if name, nameOk := execution["name"].(string); nameOk && name == targetExecutionName {
-				if executionID, idOk := execution["id"].(string); idOk && executionID != "" {
-					return executionID, nil
-				}
-			}
-		}
-	}
```
de-wim, line 291: Do we not need this loop for the fallback path anymore?

de-wim (commented):
One question

pkg/mcp/tools/executions.go
mgazza: Good catch -- the fallback path did need that verification. Pushed d221e03 which restores the name-match loop in extractExecutionIdFromListResponse.

Context: the standalone APIClient calls /agent/test-workflows/{workflowName}/executions with textSearch=<name>, so the legacy list response can include substring/prefix matches (e.g. searching `my-test-1` also matches `my-test-10`, `my-test-100`). Returning results[0] blindly would have surfaced the wrong execution ID. The new HandlerClient path is unaffected since the dedicated cloud-api endpoint does an exact server-side lookup.

pkg/mcp/tools/executions.go
```diff
@@ -269,37 +269,43 @@ func isValidExecutionName(executionName string) bool {
...
-	if len(results) == 1 {
-		if execution, ok := results[0].(map[string]any); ok {
-			if executionID, idOk := execution["id"].(string); idOk && executionID != "" {
-				return executionID, nil
-			}
+	if execution, ok := results[0].(map[string]any); ok {
+		if executionID, idOk := execution["id"].(string); idOk && executionID != "" {
+			return executionID, nil
 		}
 	}
```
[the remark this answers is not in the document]
mgazza: Fixed in d221e03 -- restored the exact-name match loop in extractExecutionIdFromListResponse. The legacy APIClient path uses a text-search list endpoint so name verification is essential; the new dedicated-endpoint path (HandlerClient) does exact lookup server-side so does not need the loop.

mgazza:
@greptileai review

mgazza:
@greptileai review

de-wim (approved):

PR #7507, opened 2026-04-14
fix: Allow TestWorkflow finalization without object storage

povilasv:
## Problem
fix: https://github.com/kubeshop/testkube/issues/7399

When TestWorkflow object storage is disabled, runner finalization should not try to upload the full execution log archive. That upload path can block terminal result persistence, leaving executions stuck in `running` and suppressing end-testworkflow webhooks.

Change is:

If `proContext.CloudStorage` is `false`, the runner does not call `SaveExecutionLogsPresigned` during finalization.

default is true

### DIFF
--- a/pkg/runner/executionsaver.go
+++ b/pkg/runner/executionsaver.go
@@ -147,13 +147,13 @@ func (s *executionSaver) End(ctx context.Context, result testkube.TestWorkflowRe
 		s.outputSaved.Store(true)
 	}
 
-	if !s.logs.Saved() && !s.logs.Enabled() {
-		log.DefaultLogger.Infow("TestWorkflow execution log storage is disabled; skipping log archive upload", "id", s.id)
-	}
-
-	if !s.logs.Saved() && s.logs.Enabled() {
-		if err := s.logs.Save(ctx); err != nil {
-			return err
+	if !s.logs.Saved() {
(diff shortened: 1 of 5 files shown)


vsukhin (approved):

pkg/runner/executionsaver.go
```diff
@@ -139,23 +138,23 @@ func (s *executionSaver) End(ctx context.Context, result testkube.TestWorkflowRe
...
-		return s.saveOutput(ctx)
-	})
-	g.Go(func() error {
-		if s.logs.Saved() {
-			return nil
+		s.outputSaved.Store(true)
+	}
+
+	if !s.logs.Saved() && !s.logs.Enabled() {
+		log.DefaultLogger.Infow("TestWorkflow execution log storage is disabled; skipping log archive upload", "id", s.id)
```
courageousillumination: Nit: Can this move inside the conditional below? I was confused reading this between the two conditions.

```
if !s.logs.Saved() {
   if s.logs.Enabled() {
		if err := s.logs.Save(ctx); err != nil {
			return err
		}
    } else {
         log.DefaultLogger.Infow("TestWorkflow execution log storage is disabled; skipping log archive upload", "id", s.id)
    }
} 
```

courageousillumination (approved):

PR #7522, opened 2026-04-16
fix: replace buf go.mod with binary download for protobuf generation

dejanzele:
## Summary

- Replace `proto/internal/buf/go.mod` with direct binary download to `bin/tooling/buf`
- Add `verify-protobuf-generated` CI job to catch stale generated proto code
- Update buf version in lint CI from 1.56.0 to 1.68.1

## Why

The separate `go.mod` for buf broke with Go 1.26. `go tool -modfile=` requires the module to be tidy, but `go mod tidy` fails because it resolves all repo packages against the buf module and hits the docker->moby module rename conflict. This has been broken since the Go 1.26 upgrade (commit 0d83ec623).

The binary download approach is simpler, faster (pre-built binary vs building from source), and immune to Go module issues. Same mechanism the `bufbuild/buf-action` CI action already uses.

## Changes

- `Makefile`: add `BUF_VERSION`, download target, update `generate-protobuf` to call buf directly, add `verify-protobuf-generated`
- `proto/gen.go`: deleted (no longer needed)
- `proto/internal/buf/go.mod`: deleted (-101 lines)
- `proto/internal/buf/go.sum`: deleted (-259 lines)
- `.github/workflows/lint.yaml`: update buf version to 1.68.1
- `.github/workflows/test.yaml`: add `verify-generated-protobuf` CI job


vsukhin (commented):
great!

vsukhin (approved):

PR #7535, opened 2026-04-18
feat: [TKC-5460] WorkflowTrigger CLI + OSS API

dejanzele:
## Summary

PR5 of the WorkflowTrigger v2 stack. Adds CLI subcommands (`testkube create|get|update|delete workflowtrigger`) plus the OSS API server endpoints they hit. Pair with the cloud-api REST endpoints (testkube-cloud-api#3564) — together they unblock the UI team and give users a kubectl-free path to manage v2 triggers.

## What's in this PR

**New packages:**
- `pkg/api/v1/testkube/workflowtrigger.go` — flat REST API shape matching cloud-api's contract
- `pkg/mapper/workflowtriggers/` — CRD <-> flat-API mapper + tests
- `pkg/newclients/workflowtriggerclient/` — interface, kubernetes impl (controller-runtime), mockgen mock
- `internal/app/api/v1/workflowtriggers.go` — 6 Fiber handlers (List, Get, Create, Update, Delete, BulkDelete)
- `pkg/api/v1/client/workflowtriggers.go` — CLI client wired for Proxy / Direct / Cloud modes
- `cmd/kubectl-testkube/commands/workflowtriggers/` — 5 cobra files (common/create/get/update/delete)

**Modified:**
- `pkg/operator/client/client.go` — register workflowtriggersv1 in the scheme so controller-runtime recognizes the CRD
- `internal/app/api/v1/server.go` — register `/workflow-triggers/*` routes and add `WorkflowTriggersClient` to TestkubeAPI
- `cmd/api-server/main.go` — construct the kubernetes-backed client, wire into NewTestkubeAPI
- `pkg/api/v1/client/{api,interface}.go` — add `WorkflowTriggerAPI` to the Client interface, wire up all three transport flavors
- `cmd/kubectl-testkube/commands/{create,get,update,delete}.go` — register the new subcommands

## Design notes

**Both OSS and Cloud mode work today:**
- CLI in **cloud mode** hits the cloud-api REST endpoints from PR 3564 via the direct/cloud transport.
- CLI in **kubeconfig mode** (OSS standalone) hits this PR's OSS API, which talks to the local K8s cluster via a controller-runtime client.
- The third deployment topology — OSS running in cloud-connected mode proxying to the control plane — needs a `cloud.go` implementation of `workflowtriggerclient` plus gRPC methods on `controlplaneclient`. Explicit follow-up.

[...]

PR #7537, opened 2026-04-18
feat: [TKC-5460] cloud-watch for WorkflowTrigger + proto

dejanzele:
## Summary

PR6 of the WorkflowTrigger v2 stack. Adds the cloud→agent sync path so dashboard-created triggers actually fire.

Parallels TestTrigger's cloud-watch: agent periodically polls cloud-api via gRPC and applies deltas to the in-memory trigger registry, using the same add/update/remove methods the local informer uses.

## What's in this PR

- **proto/service.proto**: 7 new RPCs (`GetWorkflowTrigger`, `ListWorkflowTriggers`, `CreateWorkflowTrigger`, `UpdateWorkflowTrigger`, `DeleteWorkflowTrigger`, `DeleteAllWorkflowTriggers`, `DeleteWorkflowTriggersByLabels`) + request/response messages, mirroring the TestTrigger block. Skipped the deprecated `WatchUpdates` stream and `ListLabels` since both are marked deprecated on the TestTrigger equivalent.
- **pkg/controlplaneclient/workflowtriggers.go**: gRPC client methods wrapping the generated stubs. Same shape as `testtriggers.go`.
- **pkg/newclients/workflowtriggerclient/cloud.go**: fills in the `WorkflowTriggerClient` interface for cloud-connected mode (the deferred piece from PR #7535).
- **pkg/triggers/watcher.go**: `startCloudWorkflowTriggerWatch` — ticker-based polling that diffs against a `prev` snapshot and calls `addWorkflowTrigger`/`updateWorkflowTrigger`/`removeWorkflowTrigger` directly. Skips the unstructured-decode step that the local dynamic informer uses.
- **cmd/api-server/main.go**: selects `cloudWorkflowTriggerClient` vs `kubernetesWorkflowTriggerClient` using the same `useCloudTestTriggers` flag — v1 and v2 share the same per-deployment mode decision.
- **pkg/triggers/service.go**: `WithWorkflowTriggersClient` option + field.

## Design notes

**Why reuse `useCloudTestTriggers` instead of a new flag?** The decision "is this agent cloud-connected?" is cluster-wide, not per-trigger-type. A split would let you run v1 local + v2 cloud, which is a configuration nobody would ask for. If that ever becomes a real requirement, we can split later.

**Why skip the deprecated `WatchUpdates` stream?** TestTrigger's is marked deprecated and unused; `startCloudTestTriggerWatch` uses polling. Following the same pattern here keeps the surface small.

**How cloud-watch interacts with the local dynamic informer.** Mutually exclusive: if `testTriggerControlPlane` is true AND `workflowTriggersClient` is set, we poll cloud and skip the dynamic informer. Otherwise local-only. Same conditional shape as TestTrigger's block.

**Scraper interval reused.** No reason to diverge from v1.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./pkg/triggers/` green
- [ ] Integration: deploy agent in cloud-connected mode, create WorkflowTrigger via dashboard → cloud-api DB, verify agent's cloud-watch picks it up within `scraperInterval` and `triggerStatus` contains the new entry
- [ ] Integration: delete via dashboard, verify agent removes from `triggerStatus`

## Pair PR

Cloud-api side (gRPC server handlers + REST lifecycle events): coming next, will depend on go.mod pin to this SHA ($TESTKUBE_SHA).

Depends on: kubeshop/testkube#7535 (CLI + OSS API + kubernetes workflowtriggerclient)


dejanzele:
Superseded by kubeshop/testkube#7535 after consolidation — cloud-watch, proto additions, and the cloud client are folded into #7535 so the full testkube client surface (CLI, OSS API, both k8s and cloud agent client impls, cloud-watch) reviews as a single unit.

PR #7547, opened 2026-04-21
feat: [TKC-5376] add email magic-link login to CLI

dejanzele:
## Summary

Adds email magic-link as a login method for the CLI, alongside GitHub/GitLab/Google/SSO. Closes the second half of TKC-5376 -> customers who signed up via email link can now authenticate the CLI and install runners via the CLI instructions in the UI (no longer forced to use Helm).

Linear: https://linear.app/kubeshop/issue/TKC-5376

## How it works

- New flag: `testkube pro login --email-link <email>`. For the interactive flow, a new **Email (magic link)** option in the selector prompts for the address and runs the same flow. Mirrors the web login page layout.
- Zero backend changes. The CLI reuses the existing `/auth/link` and `/auth/login/link` endpoints exactly as the web UI does.
- Loopback callback uses `http://127.0.0.1:8090/callback` (IPv4-only) to avoid the `localhost` IPv6/IPv4 resolver ambiguity that can route the browser to a different process on `::1:<port>`. Matches the existing Dex CLI loopback.
- New `TokenType = "emailLink"` stored in the CLI config so the refresh path picks `/auth/callback?method=emailLink` instead of the Dex OIDC route. A shared `refreshUserToken` helper dispatches on token type; both `oauth.go` and `client.go` use it.
- `RefreshEmailLinkToken` verifies the idToken's `exp` claim client-side before hitting the network, so the common case (token still valid) is free. Matches the verify-first pattern used by OIDC's `CheckAndRefreshToken`.

## Test plan

- [x] `go build ./cmd/kubectl-testkube/...` clean
- [x] `go vet ./cmd/kubectl-testkube/...` clean
- [x] Unit tests for `commands/pro`, `commands/common`, `commands/docker`, `cloudlogin` pass
- [x] End-to-end: `testkube pro login --email-link <email>` -> email received -> click link -> CLI captures callback -> tokens saved -> context set
- [ ] End-to-end via selector: run `testkube pro login ...` with no flag -> pick **Email (magic link)** -> enter email -> complete flow
- [ ] Confirm subsequent commands don't trigger re-auth when the token is still valid (exp-check short-circuit)
- [ ] Confirm existing GitHub/GitLab/Google flows unaffected
- [ ] Confirm existing `--email` SSO flow unaffected

### DIFF
--- a/cmd/kubectl-testkube/commands/common/helper.go
+++ b/cmd/kubectl-testkube/commands/common/helper.go
@@ -63,6 +63,7 @@ const (
 	github                = "GitHub"
 	gitlab                = "GitLab"
 	google                = "Google"
+	emailLink             = "Email (magic link)"
 	dockerDaemonPrefixLen = 8
 	latestReleaseUrl      = "https://api.github.com/repos/kubeshop/testkube/releases/latest"
 )
@@ -438,7 +439,7 @@ func PopulateHelmFlags(cmd *cobra.Command, options *HelmOptions) {
 	cmd.Flags().BoolVar(&options.EmbeddedNATS, "embedded-nats", false, "embedded NATS server in agent")
 }
 
-func PopulateLoginDataToContext(orgID, envID, token, refreshToken, dockerContainerName string, options HelmOptions, cfg config.Data) error {
+func PopulateLoginDataToContext(orgID, envID, tokenType, token, refreshToken, dockerContainerName string, options HelmOptions, cfg config.Data) error {
 	if options.Master.AgentToken != "" {
 		cfg.CloudContext.AgentKey = options.Master.AgentToken
 	}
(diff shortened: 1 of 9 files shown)


cmd/kubectl-testkube/commands/common/helper.go
```diff
@@ -63,6 +63,7 @@ const (
 	github                = "GitHub"
 	gitlab                = "GitLab"
 	google                = "Google"
+	emailLink             = "Email (magic link)"
```
courageousillumination, line 66: Do users actually know what magic link means? Is this to differentiate from SSO? Probably not a huge issue, but I think it should be `Email` for magic link and `Email (SSO)` for SSO

courageousillumination (approved):
LGTM, one non-blocking comment about naming.

cmd/kubectl-testkube/commands/common/helper.go
dejanzele: We already have SSO for customers, so I didn't want to reserve the `--email`, but explicitly mention email link.

I think magic link is a "standardish" term for email link auth.

I am welcome to suggestions though.

courageousillumination: Nah, I think this is fine. This is probably premature from my side. If people get confused maybe we just add more language to the login page.

PR #7528, opened 2026-04-17
feat: [TKC-5405] dynamic informers + resourceRef for custom resource support

dejanzele:
## Summary

PR3 (agent side): Adds dynamic informer support for custom K8s resources and `resourceRef` field on TestTrigger v1 CRD. After this PR, users can watch Argo Rollouts, Kafka topics, Cert-Manager certificates, or any other custom resource.

## Usage

```yaml
apiVersion: tests.testkube.io/v1
kind: TestTrigger
metadata:
  name: rollout-canary-test
spec:
  resourceRef:
    group: argoproj.io
    version: v1alpha1
    kind: Rollout
  resourceSelector:
    name: api-server
    namespace: production
  event: modified
  action: run
  execution: testworkflow
  testSelector:
    name: canary-smoke-test
  actionParameters:
    config:
      STEP: "jsonpath={.status.currentStepIndex}"
      IMAGE: "jsonpath={.spec.template.spec.containers[0].image}"
```

## Changes

**resourceRef on TestTrigger** (cherry-picked from tactical PR):
- `api/testtriggers/v1/testtrigger_types.go`: add `ResourceRef *TestTriggerResourceRef` field
- `pkg/triggers/internal_trigger.go`: converter uses ResourceRef when present

**Dynamic informer engine**:
- `pkg/triggers/dynamic_informer.go`: reference-counted informer manager, starts/stops per GVR
- `pkg/triggers/kind_resolver.go`: resolves Kind to GVR (hardcoded for 8 built-in types, discovery API for CRDs)
- `pkg/triggers/service.go`: `WithDynamicClient` option, auto-starts informers on trigger add, stops on remove
- `pkg/triggers/watcher.go`: initializes dynamic manager on lease acquisition, cleans up on release

**How it works**:
1. User creates TestTrigger with `resourceRef: {group: argoproj.io, kind: Rollout}`
2. `addTrigger()` calls `ensureDynamicInformerForTrigger()`
3. Kind "Rollout" resolved to GVR `argoproj.io/v1alpha1/rollouts` via discovery API
4. Dynamic informer started for that GVR across watched namespaces
5. On Rollout changes, `dynamicEventHandler` creates `watcherEvent` from `*unstructured.Unstructured`
6. Existing matcher evaluates the event against all registered triggers
7. JSONPath expressions in `actionParameters.config` work on unstructured data

## Test plan

- [ ] `go test ./pkg/triggers/` passes
- [ ] `go build ./...` passes
[...]

PR #7523, opened 2026-04-16
feat: [TKC-5460] WorkflowTrigger CRD types, field matching engine, internal trigger type

dejanzele:
## Summary

PR1 of the WorkflowTrigger v2 implementation (TKC-5405). Adds the foundation types, field matching engine, and internal trigger representation. Zero behavior change to existing TestTriggers.

## What's in this PR

**WorkflowTrigger CRD** (`testworkflows.testkube.io/v1`)
- New CRD with grouped spec: `watch`, `when`, `match`, `wait`, `run`
- Shares API group with TestWorkflow (same RBAC, same Helm chart section)
- `kubectl get workflowtriggers` (no collision with `kubectl get testtriggers`)
- Generated CRD YAML and deepcopy methods

**Field matching engine** (`pkg/triggers/field_matcher.go`)
- 7 JSONPath operators: equals, not_equals, exists, not_exists, changed, changed_to, changed_from
- Works on typed K8s objects and unstructured maps (custom resources)
- 26 table-driven tests

**Internal trigger type** (`pkg/triggers/internal_trigger.go`)
- Unified representation for both v1 TestTrigger and v2 WorkflowTrigger
- v1 converter resolves resource enum to GVK (all 8 types tested)
- v2 converter maps directly

**OldObject pass-through** (`pkg/triggers/event.go`, `pkg/triggers/watcher.go`)
- Added `OldObject` field to `watcherEvent`
- Passed in all 6 UpdateFunc handlers (deployment, statefulset, daemonset, service, ingress, configmap)
- Enables `changed`/`changed_to`/`changed_from` operators

**Proto sync messages**
- `workflow_trigger = 6` added to UpdateOrCreateRequest and DeleteRequest oneof
- New workflow_trigger.proto and workflow_trigger_id.proto

## What's NOT in this PR

- Matcher refactor to use internal type (PR2)
- Dynamic informers for custom resources (PR2)
- Control plane: database, REST API, sync handler (PR3)
- Helm chart, end-to-end tests (PR4)

## Test plan

- [ ] `go test ./pkg/triggers/` passes (26 new tests + all existing)
- [ ] `go build ./api/workflowtriggers/v1/` compiles
- [ ] `go build ./...` full build passes
- [ ] Existing TestTrigger behavior unchanged (no modifications to matcher/executor/service)

Depends on: kubeshop/testkube#7522 (buf fix, for proto generation)

PR #7552, opened 2026-04-21
feat(api): RBAC for workflowtriggers CRD + rbac.extraWatchedResources for resourceRef targets

dejanzele:
## Summary

Fixes two RBAC gaps in the `testkube-api` Helm chart that surfaced when testing the new `resourceRef` feature on TestTrigger / WorkflowTrigger (PRs #7523, #7528, #7535):

### 1. Missing Role for the `workflowtriggers` CRD

The new `WorkflowTrigger` CRD (added in #7523) didn't get a matching `<kind>-role` + `<kind>-crb` in the chart. Every other CRD in this chart follows that exact pattern (`testtriggers-role`, `testworkflows-role`, `webhooktemplate-role`, etc.); `workflowtriggers` was simply missed.

Effect on a fresh install: the agent's controller-runtime manager fails to sync the `*v1.WorkflowTrigger` informer cache at startup (`Forbidden: cannot watch resource "workflowtriggers"`) and the pod enters CrashLoopBackOff.

**Fix**: add `workflowtriggers-role-{release}` + `workflowtriggers-crb-{release}` using the identical structure as `testtriggers-role` / `testtriggers-crb`.

### 2. No way for users to grant watch permissions on `resourceRef` target CRDs

`TestTrigger.spec.resourceRef` (and `WorkflowTrigger.spec.resourceRef`) let users watch **any** K8s resource by GVK — including arbitrary CRDs like `cert-manager.io/Certificate`, `kafka.strimzi.io/KafkaTopic`, etc. The trigger service starts a dynamic informer for each unique GVK.

Today there's no chart values knob to grant the agent read access on those CRDs. Admins have to hand-write a ClusterRole + ClusterRoleBinding per deployment. That's the first thing they hit when trying the feature.

**Fix**: add `rbac.extraWatchedResources` values key — a list of `{apiGroups, resources, verbs?}` entries rendered into the `watchers-role` ClusterRole.

```yaml
# values.yaml
rbac:
  extraWatchedResources:
    - apiGroups: ["cert-manager.io"]
      resources: ["certificates"]
    - apiGroups: ["kafka.strimzi.io"]
      resources: ["kafkatopics"]
      verbs: ["get", "list", "watch"]   # optional; defaults to read-only trio
```

Verbs default to `[get, list, watch]` — read-only, matching what the informers need.

## Test plan

- [x] `helm template` renders `workflowtriggers-role-{release}` and `workflowtriggers-crb-{release}` correctly.
- [x] `helm template` with `rbac.extraWatchedResources` populated renders the extra rules appended to `watchers-role`.
- [x] Applied to a running local k3d cluster with the trigger PRs: agent previously CrashLoopBackOff on `Forbidden: cannot watch workflowtriggers`, now boots clean and the trigger service acquires its lease.
- [x] End-to-end: a TestTrigger with `resourceRef: cert-manager.io/v1/Certificate` successfully watches Certificate events and schedules TestWorkflow executions on match.

## Note on trigger scoping (not a bug — learned during validation)

A `TestTrigger` without `resourceSelector` deliberately defaults to its own namespace for scope — that's intentional matcher behavior at `pkg/triggers/matcher.go:156`, not a bug. Users who want cross-namespace matching need an explicit `resourceSelector.namespace` / `resourceSelector.namespaceRegex` / label selector.

### DIFF
--- a/k8s/helm/testkube-api/templates/role.yaml
+++ b/k8s/helm/testkube-api/templates/role.yaml
@@ -424,6 +424,40 @@ rules:
       - get
 
 ---
+
+apiVersion: {{ include "global.capabilities.rbac.apiVersion" . }}
+kind: Role
+metadata:
+  name: workflowtriggers-role-{{ .Release.Name }}
+  labels: {{- include "testkube-api.labels" . | nindent 4 }}
+    {{- if .Values.global.labels }}
+    {{- include "global.tplvalues.render" ( dict "value" .Values.global.labels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.global.annotations}}
+  annotations: {{- include "global.tplvalues.render" ( dict "value" .Values.global.annotations "context" $ ) | nindent 4 }}
+  {{- end }}
+rules:
+  - apiGroups:
+      - testworkflows.testkube.io
+    resources:
+      - workflowtriggers
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
+      - deletecollection
+  - apiGroups:
+      - testworkflows.testkube.io
+    resources:
+      - workflowtriggers/status
+    verbs:
+      - get
+
+---
 {{- if .Values.next.testTriggers.enabled }}
 
 apiVersion: {{ include "global.capabilities.rbac.apiVersion" . }}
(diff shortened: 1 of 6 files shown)


vsukhin (approved):

povilasv (approved):

dejanzele:
@greptileai addressed in `ed52a6407`:

**P1 (multinamespace / additionalNamespaces):** the same `{{- with .Values.rbac.extraWatchedResources }}` block now renders in all three watcher role variants:

- cluster-wide `ClusterRole` (default, `multinamespace.enabled=false`)
- release-namespace `Role` (`multinamespace.enabled=true`)
- per-namespace `Role`s from `additionalNamespaces`

Verified via `helm template` against all three configurations and `helm lint` on each.

`values.yaml` comment updated to name all three modes instead of saying "cluster-wide".

Ready for re-review.

PR #7562, opened 2026-04-22
fix(executions): Improving executions query performance

xoscar:
## Pull request description

Speeds up `GetTestWorkflowExecutionsTotals` (the totals call behind `GET /executions`) by denormalizing `status` and `workflow_name` onto `test_workflow_executions` and rewriting the query to read them directly. Mirrors MongoDB's embedded-document shape so the planner can serve totals from a single composite index instead of a parallel seq-scan + hash-join + temp-spill.

Validated on a 250k-row dataset:
- Old: ~161 ms hot / ~1.4 s cold
- New: ~27 ms (Bitmap Index Scan on `idx_twe_org_env_wfname_status`, both LEFT JOINs elided)

## Checklist

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

- None.

## Changes

- New migration `20260422120000_denormalize_execution_status_name.sql`: adds `status` and `workflow_name` columns, backfills, installs `AFTER INSERT/UPDATE` triggers on `test_workflow_results` and `test_workflows` to keep them in sync (idempotent via `IS DISTINCT FROM`).
- New migration `20260422120001_index_executions_status_name.sql`: composite covering indexes `idx_twe_org_env_wfname_status` and `idx_twe_org_env_wfname_sched` (CONCURRENTLY).
- Rewrote `GetTestWorkflowExecutionsTotals` in `executions.sql` to read `e.status` / `e.workflow_name`. LEFT JOINs kept only because `@initialized`, `@health_ranges`, `@label_*`, `@selector_*` still reference `r`/`w` — planner elides them when those filters are absent.
- Regenerated sqlc.
- Added `TestPostgresDenormalizedExecutionColumns_Integration` covering trigger sync (insert/update/`resolved_workflow` ignore) and the rewritten query end-to-end.

## Fixes

- Slow / 504 responses on the executions list endpoint when totals run cold or against large workflow histories.

### DIFF
--- a/pkg/controlplane/scheduling/sqlc/execution_querier.sql.go
+++ b/pkg/controlplane/scheduling/sqlc/execution_querier.sql.go
@@ -11,7 +11,7 @@ import (
 
 const getExecutionsByStatus = `-- name: GetExecutionsByStatus :many
 SELECT
-    e.id, e.name, e.namespace, e.number, e.test_workflow_execution_name, e.group_id, e.runner_id, e.runner_target, e.runner_original_target, e.disable_webhooks, e.tags, e.running_context, e.config_params, e.scheduled_at, e.assigned_at, e.status_at, e.created_at, e.updated_at, e.organization_id, e.environment_id, e.runtime, e.silent_mode,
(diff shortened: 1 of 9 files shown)


vsukhin (approved):
great, thought about the same denormaliation, aslo add org id and env id to test_workflows table

PR #7566, opened 2026-04-22
feat(triggers): schema-aware match conditions

dejanzele:
## Summary
- Pushes CRD OpenAPI schemas from the agent to the Control Plane and surfaces them in the UI as schema-aware autocomplete for `TestTrigger.match[].path`.
- Validates each match-condition path against the canonical regex (`MatchPathPattern`) and rejects bracket suffixes (`[*]`, `[N]`); the expression engine in `pkg/expressions` can't tokenize them yet, so array-path support is a follow-up.
- Pins schema-aware triggers to one or more listener agents via `Spec.Listener.Match["id"]`. The same field drives both `Validate()` and the runtime gate (`triggerTargetsThisAgent`), so a trigger is validated against, and only fires on, the listener whose cluster snapshot it was checked against.
- Extracts `ValidateMatchConditions` (with typed reason codes) from `Spec.Validate()` so cp-api reuses the exact per-condition match rules instead of keeping a parallel copy that can drift.
- Adds the inventory push pipeline: a gRPC client plus controller that pushes on startup, on a 1h tick, and on CRD informer events (debounced 5s to coalesce helm-install bursts).

## Companion PR
Required: the testkube-cloud-api PR receives, caches, and serves the snapshot, validates `match[]` against the schema, and ships the UI. cp-api's `go.mod` is pinned to this branch's HEAD; the two land together.

## Known gaps (deferred)
- Array-path matching: the backend can't evaluate `[*]`/`[N]` yet (the validator catches it at save time; runtime no-op pinned by `TestMatchOnCRDArrayPathsCurrentlyUnsupported`).
- Schemas for built-in K8s types (Pod, Deployment): only CRDs ship a schema today; built-ins fall back to manual path entry.


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptile

dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptile

dejanzele:
@greptileai please re-review — latest commit addresses the prior findings.

dejanzele:
@greptileai

dejanzele:
@greptileai please re-review the latest changes (listener pinning reshape + trigger UX updates).

dejanzele:
@greptileai please re-review the latest commit. Since the last review: extracted `ValidateMatchConditions` (with typed reason codes) out of `Spec.Validate()` so the per-condition `match[]` rules are single-sourced and reusable by cp-api. `Validate()` now calls it; behavior is unchanged for the agent path since all callers only check `len(errs) > 0`.

PR #7572, opened 2026-04-23
fix: [TKC-5014] check workflow services health at the end of test workflow execution

de-wim:
## Pull request description 

This PR adds a health check before killing workflow services, detecting OOMKilled (or otherwise crashed) containers by inspecting pod restart counts. Previously, once a service passed its readiness probe, no further health validation occurred, so post-startup crashes went unnoticed. The stop-services stage is also made non-optional so detected failures properly mark the execution as failed.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

- None

## Changes

- Executions are now marked as failures when a workflow service crashes during testing

## Fixes

- TKC-5014

### DIFF
--- a/build-and-load-tw-toolkit.sh
+++ b/build-and-load-tw-toolkit.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -eo pipefail
+
+AGENT=double-o-seven
+
+# Build
+
+tk-dev build -c=a
+
+# Setup depot on agent namespace
+if ! (kubectl get secret depot -n ${AGENT} &>/dev/null); then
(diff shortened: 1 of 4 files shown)


de-wim:
@greptileai please take a look

courageousillumination (approved):
Is restarts the only way we have to look at this? Seems a bit indirect, but if that's all that's available this is fine with me.

@tkonieczny do you think this is a breaking change or fixing a bug? It might cause some tests that were "passing" before to now fail (if they were passing without relying on the services). I think it's probably fine, but I don't want anyone to be confused about the new behavior.

de-wim:
> Is restarts the only way we have to look at this? Seems a bit indirect, but if that's all that's available this is fine with me.

I know, it feels a bit... hacky. But I tried a few other things, such as notifications, but those didn't work.

I'm exploring using events now.

PR #7586, opened 2026-04-27
fix: validate execution before prompt, prevent duplicate uploads and …

Valentin-Marko:
## Pull request description 

Fetch the execution before asking the user to confirm public sharing, and look up an existing share for that execution before re-uploading. On a missing execution show a friendly "not found" error instead of a raw API error after the user has already accepted the prompt.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/view.go
+++ b/cmd/kubectl-testkube/commands/view.go
@@ -58,6 +58,48 @@ Accepts either an execution ID (UUID) or an execution name (e.g. my-workflow-123
 				return
 			}
 
+			client, _, err := common.GetClient(cmd)
+			ui.ExitOnError("getting client", err)
+
+			ui.Info("Fetching execution", args[0])
+			execution, err := client.GetTestWorkflowExecution(args[0])
+			if err != nil {
+				if isExecutionNotFoundError(err) {
+					ui.Failf("Execution %q was not found. Check the execution ID or name with `testkube get twe`.", args[0])
+				}
+				ui.ExitOnError("get test workflow execution failed", err)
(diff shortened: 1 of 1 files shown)


Valentin-Marko:

cmd/kubectl-testkube/commands/view.go
```diff
@@ -58,14 +58,56 @@ Accepts either an execution ID (UUID) or an execution name (e.g. my-workflow-123
...
+			uris := common.NewMasterUris("", "", "", "", "", false)
+			if sharesAPIURL == "" {
+				sharesAPIURL = uris.Api
+			}
+			if viewerBaseURL == "" {
+				viewerBaseURL = uris.View
+			}
+
+			if existing, err := lookupExistingShare(sharesAPIURL, execution.Id); err != nil {
+				ui.Debug("could not check for existing share", err.Error())
```
vsukhin, line 93: Should it be warning?

vsukhin (approved):
one minor comment

cmd/kubectl-testkube/commands/view.go
Valentin-Marko: Intentional: the lookup is just a performance optimization. If it fails (network issues, server 5xx errors, or the CLI hitting an older backend without the new endpoint), the CLI simply falls back to the normal upload flow.  Happy to switch to Warn if you'd prefer.

PR #7517, opened 2026-04-16
feat(TKC-5494): Adding hints to open the execution viewer

Valentin-Marko:
## Pull request description 

We should guide users to view execution results in the browser by adding contextual hints in CLI outputs for execution-related commands.

## Checklist (choose whats happened)

- [ ] breaking change! (describe)
- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

PR #7597, opened 2026-04-28
fix(cli): gate UIShell hints on TTY so -o json output stays parseable

dejanzele:
## Summary

The `UIShell*` helpers (`UIShellGetExecution`, `UIShellViewExecution`, `UIShellWatchExecution`) write decorative prose to stdout via `ui.Hint` + `ui.ShellCommand`. When stdout is piped or redirected, the prose lands in the parser's input.

Concretely, the OSS workflow tests run:

```sh
exec_name=$(testkube get tw <name> --namespace ... -o json \
  | awk '/^{/ {i++} i==2' \
  | jq -r '.name')
```

After the helpers fire on the second JSON block (`LatestExecution`), `awk` keeps streaming the trailing `💡 ...` / `$ ...` lines into `jq`, which fails with `parse error: Invalid numeric literal`. With `set -o pipefail`, the whole `$(...)` fails, the OSS tests abort.

This PR keeps the hints exactly as #7517 designed them for interactive users, and silences them when stdout isn't a TTY (pipes, redirects, in-cluster steps, CI).

## Approach

A single `hintsEnabled` gate in `cmd/kubectl-testkube/commands/common/uihints.go` short-circuits all three helpers. Behavior unchanged in a real terminal; suppressed everywhere a parser is reading.

Picked TTY-gating over "send hints to stderr" because TTY-gating cleanly covers the case where CI captures both streams (`2>&1`), which is common.

## Reproduction

```sh
$ go run ./test-repro.go | awk '/^{/ {i++} i==2' | jq -r '.name'
jq: parse error: Invalid numeric literal at line 2, column 5
junit5-workflow-smoke-1
$ echo "exit: $?"
exit: 5
```

After the fix, same pipeline exits 0 with clean JSON.

### DIFF
--- a/cmd/kubectl-testkube/commands/common/uihints.go
+++ b/cmd/kubectl-testkube/commands/common/uihints.go
@@ -1,29 +1,36 @@
 package common
 
 import (
+	"os"
+
+	"github.com/mattn/go-isatty"
+
 	"github.com/kubeshop/testkube/pkg/ui"
 )
 
-// UIShellGetExecution prints the testkube command to get test workflow execution details.
+// Gate hints on a TTY so non-interactive stdout (e.g. `-o json | jq`) stays parseable.
+// Var (not func) so tests can override.
+var hintsEnabled = func() bool {
+	fd := os.Stdout.Fd()
+	return isatty.IsTerminal(fd) || isatty.IsCygwinTerminal(fd)
+}
+
+func shellHint(hint, cmd string) {
+	if !hintsEnabled() {
+		return
+	}
+	ui.Hint(hint)
+	ui.ShellCommand(cmd)
+}
+
 func UIShellGetExecution(id string) {
-	ui.Hint("Get the TestWorkflow execution details:")
-	ui.ShellCommand(
-		"testkube get twe " + id,
-	)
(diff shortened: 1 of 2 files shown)


vsukhin (approved):

vsukhin:
```
💡  Get the TestWorkflow execution details:
$ testkube get twe 69f0ad6e9ff8a43bcd3cc23b
💡  View the TestWorkflow execution details in your browser:
$ testkube view 69f0ad6e9ff8a43bcd3cc23b
```
should be kubectl testkube

olensmar:
> ```
> 💡  Get the TestWorkflow execution details:
> $ testkube get twe 69f0ad6e9ff8a43bcd3cc23b
> 💡  View the TestWorkflow execution details in your browser:
> $ testkube view 69f0ad6e9ff8a43bcd3cc23b
> ```
> 
> should be kubectl testkube

preferably not - we removed kubectl from all hints - the cli works on its own!?

vsukhin:
@olensmar Not sure, what binary is distibuted with current package managers, at least before, it was kubectl-testkube
for Cloud you can use just testkube to connnect to Rest API, but for OSS k8s proxy, we still need to have plugin mode, I guess

PR #7604, opened 2026-04-29
fix: [TKC-5587] keep quiet workflow log streams alive

povilasv:
## Summary

- emit protocol heartbeat notifications from the direct/proxy workflow notification SSE path while a stream is otherwise quiet
- keep durable workflow notification sequencing unchanged so heartbeats do not advance resume cursors
- stop printing expected SSE reader shutdown errors for `io.EOF`, `context.Canceled`, and `context.DeadlineExceeded`


vsukhin (approved):
looks good

PR #7696, opened 2026-05-14
feat: align te values and resources with testkube - add unit tests

jose-tk:
## Pull request description 

adding the missing resources, aligning them to the `testkube` and adding unit tests to verify the behavior of them

https://linear.app/kubeshop/issue/TKC-5722/add-service-and-servicemonitor-to-testkube-runner-helm-chart-for

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/k8s/helm/testkube-runner/values.yaml
+++ b/k8s/helm/testkube-runner/values.yaml
@@ -253,9 +253,9 @@ service:
   ## Service type
   type: ClusterIP
   ## External Service port. NOTE: the runner container always listens on 8088
-  ## internally (APISERVER_PORT is hardcoded in the deployment template), so
-  ## changing this only affects the Service's external port — metrics scraping
-  ## uses the named port "http", not this value.
+  ## internally (APISERVER_PORT is hardcoded in the deployment template) and
+  ## the ServiceMonitor scrapes that port directly, so changing this only
+  ## affects the Service's external port — not metrics scraping.
   port: 8088
   ## Service Annotations
   annotations: {}
--- a/k8s/helm/testkube-runner/values.yaml
+++ b/k8s/helm/testkube-runner/values.yaml
@@ -248,6 +248,28 @@ pod:
   # - name: FOO
   #   value: BAR
 
+## Service parameters
+service:
+  ## Service type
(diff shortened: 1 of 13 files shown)


k8s/helm/testkube-runner/values.yaml
```diff
@@ -248,6 +248,28 @@ pod:
   # - name: FOO
   #   value: BAR
 
+## Service parameters
+service:
+  ## Service type
+  type: ClusterIP
+  ## HTTP Port (matches the runner containerPort 8088 and APISERVER_PORT)
+  port: 8088
```
caiomede-tk, line 259: As the greptile comments says, the service port is not used by the Deployment, so if a customer tries customizing this port the Service Monitor will not capture metrics because Deployment port and the software piece port is hardcoded and fixed. I'd recommend to keep the port hardcoded and fixed in port `8088`, can you test if the metrics are really exposed? Idk if these components are not exposed for testkube runner because they disabled it someway, just to be sure. Here the docs: https://docs.testkube.io/articles/metrics#grafana-dashboard

jose-tk: fixed ;)

k8s/helm/testkube-runner/values.yaml
```diff
@@ -248,6 +248,31 @@ pod:
   # - name: FOO
   #   value: BAR
 
+## Service parameters
+service:
```
caiomede-tk, line 252: Is not better to remove this whole values slot called service? because what is value behind it?

jose-tk: I'd lean toward keeping it to align the runners chart like the others, also is part of the helm convention, like creating a service that exposes the service, if you prefer we can remove it, but I don't think it's a bad idea to keep it, wdyt?

caiomede-tk: OK, let's see if it doesn't generate any problem

caiomede-tk (approved):
LGTM

PR #7749, opened 2026-05-27
feat: make workflowName optional for get_execution_info MCP tool

olensmar:
## Summary

`get_execution_info` previously required both `workflowName` and `executionId`, but the underlying API already exposes an executionId-only lookup at `/test-workflow-executions/{executionID}`, and the agent handler's `Get(ctx, id)` supports lookup by id or name. Forcing callers to also supply a workflow name was friction for the common case where the LLM has an execution id from \`list_executions\` or \`lookup_execution_id\`.

- Drop `mcp.Required()` from `workflowName`; switch to `OptionalParam`.
- When `workflowName` is empty, \`APIClient.GetExecutionInfo\` hits `/agent/test-workflow-executions/{executionId}`; otherwise the existing workflow-scoped path is preserved so name-based, workflow-disambiguated lookups still work.
- Add a tool-specific description for the parameter and tighten \`GetExecutionInfoDescription\` so the model knows when the workflow name is actually needed.

The cloud-side counterpart change is in [kubeshop/testkube-cloud-api#TBD](https://github.com/kubeshop/testkube-cloud-api).

## Test plan

- [x] `make lint` (0 issues)
- [x] `go build ./...`
- [x] `go test ./pkg/mcp/...`
- [ ] Smoke: call \`get_execution_info\` with only \`executionId\` against a local agent
- [ ] Smoke: call \`get_execution_info\` with both \`workflowName\` and \`executionId\` (legacy path)


pkg/mcp/tools/executions.go
```diff
@@ -192,17 +192,17 @@ type ExecutionInfoGetter interface {
...
 
 	handler = func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		workflowName, err := RequiredParam[string](request, "workflowName")
+		executionId, err := RequiredParam[string](request, "executionId")
 		if err != nil {
 			return mcp.NewToolResultError(err.Error()), nil
 		}
 
-		executionId, err := RequiredParam[string](request, "executionId")
+		workflowName, err := OptionalParam[string](request, "workflowName")
```
topliceanurazvan, line 205: The new optional-`workflowName` path doesn't have unit test coverage yet. The existing `FetchExecutionLogs` tests in `executions_test.go` show a clean pattern to follow — would be great to have at least one case covering the id-only lookup so the routing logic is protected going forward.

topliceanurazvan (approved):
LGTM

pkg/mcp/tools/executions.go
olensmar: coming up.. 👍

olensmar:
@greptileai

pkg/mcp/tools/executions.go
olensmar: added unit test @topliceanurazvan - ok to merge?

topliceanurazvan: integrations tests are failing - should look into that 👀

topliceanurazvan: Seems like it was a flaky test - lets proceed with merging 🚀

PR #7768, opened 2026-05-29
fix: [TKC-5851] optimize execution summary hydration

povilasv:
## Summary

- Rewrite `GetTestWorkflowExecutionsSummary` so matching executions are paged before hydrating workflow/result/resource detail rows.
- Keep selector, status, runner, initialized, label, tag, and health filtering behavior in the page-selection subquery.
- Add a concurrent Postgres index for runner/status execution summary pages ordered by `scheduled_at DESC`.

## Context

Cloud SQL Query Insights showed this query family dominating DB time. A representative live plan hydrated 954,037 matching rows before returning a 10-row page, taking about 56.6s. The rewritten shape hydrates only the selected page rows; the same representative shape ran in about 1.38s before the new index is present.

Companion Cloud API migration PR will add the same index to the service-owned migration chain.

Linear: TKC-5851

## Validation

- `make generate-sqlc`
- `go test ./pkg/repository/testworkflow/postgres`
- `git diff --check`
- Read-only representative `EXPLAIN ANALYZE` against live schema

### DIFF
--- a/pkg/database/postgres/migrations/20260529110000_index_executions_runner_status_scheduled.sql
+++ b/pkg/database/postgres/migrations/20260529110000_index_executions_runner_status_scheduled.sql
@@ -1,0 +1,14 @@
+-- +goose NO TRANSACTION
+
+-- +goose Up
+-- Covers execution summary pages filtered by runner/status and ordered by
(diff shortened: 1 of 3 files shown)


povilasv:
@greptile

vsukhin (approved):

PR #7786, opened 2026-06-03
feat(agentserver): send registration metadata on reconnect (TKC-5876)

caiomede-tk:
## Summary

Today an agent record is written once at `Register` time and never updated after that, so changes to deployment labels, `RUNNER_GROUP`, or `IS_GLOBAL` do not propagate without deleting and re-registering the runner.

This PR extends the existing `UpdateAgentCapabilitiesOnStartup` gRPC — already authenticated and called on every reconnect — to also carry the runner-supplied registration metadata. Server-side application lives in [kubeshop/testkube-cloud-api#…](https://github.com/kubeshop/testkube-cloud-api/pull/new/tkc-5876-refresh-runner-registration-metadata).

## Changes

- `proto/service.proto` — extended `UpdateAgentCapabilitiesOnStartupRequest` with `labels`, `runner_group`, `is_global`, `update_registration_metadata`. Regenerated `pkg/cloud/service.pb.go`.
- `cmd/api-server/main.go` — extracted `collectRunnerRegistrationLabels` helper used by both initial `Register` and the post-register `UpdateAgentCapabilitiesOnStartup` so both paths share one source of truth. Reconnect call now sends labels + group + global with the opt-in flag.

## Backward compatibility

- New `update_registration_metadata` field is opt-in. Old runners leave it unset → server runs the previous capabilities-only path.
- New runner ↔ old server: server returns `codes.Unimplemented`; runner already logs and continues, no functional break.

## Verification

- `go build ./cmd/api-server/...` — clean.
- Manual e2e on Kind: deployed the runner Helm chart, restarted with new label/group/global values; the Control Plane now reflects them.

## Linear

Refs: [TKC-5876](https://linear.app/kubeshop/issue/TKC-5876/refresh-runner-labels-runner-group-and-is-global-on-agent-reconnect)


buarki (approved):

caiomede-tk:
Thanks @greptile-apps. Addressed both findings in d12763e12:

**P1 — Avoid empty refreshes:** `getDeploymentLabels` and `collectRunnerRegistrationLabels` now return `(map[string]string, error)`. On the reconnect path, when the Deployment lookup fails we set `UpdateRegistrationMetadata: false` so the control plane preserves the existing label set instead of being overwritten with just `registration:self`. The capabilities update still goes through.

**P2 — Use update timeout:** the reconnect-path label fetch is now scoped to `updateCtx`, sharing the 5s control-plane RPC budget. A stalled Kubernetes Deployment GET can no longer block startup past the intended timeout.

Initial `Register` path still falls back to `{registration:"self"}` on lookup failure since there is no prior label set to clobber on first registration.

The two failing checks (`Unit Tests` and `Integration Tests`) look like CI infra flakes — `Unit Tests` shows the test step at `success` with all subsequent `if: always()` steps unexpectedly skipped (runner cancellation), and the `Integration Tests` failure is in `TestInitProcessCore_Integration/ExecuteCommand` (exit 137 / SIGKILL — OOM, unrelated to this PR which only touches `cmd/api-server/main.go`). New push should rerun them.

caiomede-tk:
@greptile please review again

caiomede-tk:
Good catch @greptile-apps — addressed in 3c02b14ae.

The single `update_registration_metadata` flag has been split into two independent gates:

- `update_labels` — gated on the runner's Deployment label lookup succeeding. Set to `false` only when the K8s GET fails (RBAC / name mismatch / API error), preventing label clobbering.
- `update_runner_policy` — always `true` on the reconnect path. Refreshes `runner_group` and `is_global` from `cfg`, which do not depend on the Kubernetes Deployment lookup, so changes to runner routing/scope propagate even on K8s API failures.

On the cloud-api side ([kubeshop/testkube-cloud-api#3961](https://github.com/kubeshop/testkube-cloud-api/pull/3961) commit `d7c3db72a`), `applyRunnerRegistrationMetadata` now respects each flag independently — if only one section is opted in, the other is preserved verbatim from the stored agent state. Added a sub-test `updates only runner policy when label refresh is suppressed` covering the new partial-update path.

Older runners still send neither flag set, preserving capabilities-only behavior.

PR #7791, opened 2026-06-04
fix(agentserver): preserve CLI-managed labels and policy on reconnect (TKC-5876)

caiomede-tk:
Linear: [TKC-5876](https://linear.app/testkube/issue/TKC-5876)
Follow-up to #7786 \(merged\).

## Why \(backwards-compat hazard introduced by #7786\)

#7786 made the runner republish its labels, runner group, and global flag to the Control Plane on every reconnect, with two opt-in flags \(`update_labels`, `update_runner_policy`\) so a transient label-read failure could not wipe stored labels.

That works for greenfield deployments, but for **existing customers whose runner metadata was managed via the CLI** \(`testkube update agent -l ...`, `testkube create agent --global`, `--group ...`\) without ever copying those values into Helm, the very first reconnect after upgrade silently destroys their state:

| Customer scenario \(pre-upgrade\)                                              | Runner sends                                              | Stored CP state after first reconnect |
|------------------------------------------------------------------------------|-----------------------------------------------------------|---------------------------------------|
| Labels set with `testkube update agent -l env=prod`, no Helm `runner.register.labels` | `UpdateLabels=true`, `Labels={"registration":"self"}`        | Labels reduced to just `registration=self` — `env=prod` lost |
| `--global` set via CLI, no Helm `runner.register.global`                       | `UpdateRunnerPolicy=true`, `IsGlobal=false`, `RunnerGroup=""` | Demoted to Independent                |
| `--group eu-runners` via CLI, no Helm `runner.register.groupName`              | `UpdateRunnerPolicy=true`, `RunnerGroup=""`                  | Demoted to Independent                |

Because `labelsErr == nil` is true for a Deployment that simply has zero matching annotations \(empty read is still a successful read\), the existing `update_labels` gate is too coarse. And `update_runner_policy` had no gate at all.

## What

Tighten both gates so the runner only opts into a refresh when it actually has an explicit Helm-side configuration to publish:

```go
runnerLabels, hasUserLabels, labelsErr := collectRunnerRegistrationLabels(...)
updateLabels := labelsErr == nil && hasUserLabels
updateRunnerPolicy := cfg.RunnerGroup != "" || cfg.IsGlobal
```

- `hasUserLabels` is true iff the runner Deployment exposed ≥1 annotation under `cfg.RunnerLabelsPrefix`. The synthetic `"registration":"self"` marker does **not** count.
- `updateRunnerPolicy` is true only when at least one of `cfg.RunnerGroup` or `cfg.IsGlobal` is set, so an all-zero \("Independent"\) policy snapshot does not demote a CLI-set Global/Grouped runner.

Both branches log an info-level message when the refresh is suppressed, distinguishing "label read failed" \(warn\) from "no user-configured labels / no explicit policy" \(info\).

## Behavior matrix after this fix

| Helm config                                         | CP-stored labels      | CP-stored policy      | After reconnect                                      |
|-----------------------------------------------------|-----------------------|-----------------------|------------------------------------------------------|
| No `runner.register.labels`, no global/group         | Anything              | Anything              | **Both preserved** \(no refresh\)                      |
| `runner.register.labels.region=europe`               | Anything              | Anything              | Labels replaced with Helm set; policy preserved      |
| `runner.register.global=true`                        | Anything              | Anything              | Policy set to Global; labels preserved unless Helm labels also set |
| `runner.register.groupName=eu`                       | Anything              | Anything              | Policy set to Grouped \(group=eu\); labels preserved unless Helm labels also set |
| K8s API error reading Deployment                    | Anything              | Anything              | Labels preserved \(warn log\); policy refreshed only if global/group set |

## Trade-off

Demoting a runner from Global/Grouped → Independent **cannot** be done by clearing the Helm value and reconnecting; it must be done explicitly via `testkube update agent`. This matches the behavior before #7786 and avoids the silent-downgrade hazard. We can revisit with an explicit opt-in env var \(e.g. `RUNNER_REGISTRATION_REFRESH=true`\) if customers need the cleared-values-mean-Independent semantics.

## Verification

- `go build ./cmd/api-server/...` clean.
- `go vet` shows only the pre-existing `superagentmigration.go:87:4: unreachable code` warning \(unrelated, present on `origin/main`\).
- The control-plane side \([kubeshop/testkube-cloud-api#3961](https://github.com/kubeshop/testkube-cloud-api/pull/3961)\) requires no changes — it already only updates each side of the metadata when the corresponding opt-in flag is true.

cc/ docs follow-up: [kubeshop/testkube-docs#572](https://github.com/kubeshop/testkube-docs/pull/572)


caiomede-tk:
Pushed [`ee489ec74`](https://github.com/kubeshop/testkube/pull/7791/commits/ee489ec74) to unblock the failing **Lint** check.

## What was failing

```
pkg/triggers/git_trigger_test.go:516:1: File is not properly formatted (goimports)
```

That file is **not** part of this PR's diff (the only file we modify is `cmd/api-server/main.go`). It is a pre-existing issue on `main` introduced in [#7771](https://github.com/kubeshop/testkube/pull/7771) (TKC-5875): `pkg/triggers/git_trigger_test.go` is missing its trailing newline. Verified by running `goimports -d` against `origin/main`.

## Why it surfaced here

The Lint workflow uses the golangci-lint-action `only-new-issues` filter, which normally hides pre-existing findings. For this PR that filter consistently fails:

```
failed to fetch push patch: Request Error [HttpError]: Not Found - .../compare/0000000000000000000000000000000000000000...5a2e399cd...
status: 404
```

The action is calling the compare API with an all-zero base SHA, so the diff cannot be computed and the filter is bypassed. I retried the failed job once \u2014 same outcome \u2014 so it is not a transient issue. The root cause looks like an action-input/checkout-config problem, not anything specific to this branch.

## What I did

The new commit is a one-byte change: adds the missing trailing newline to `pkg/triggers/git_trigger_test.go`. It is functionally a no-op and unblocks Lint on this PR. Happy to revert and split it into a separate cleanup PR if reviewers prefer \u2014 just let me know.

buarki (approved):

xoscar (approved):

PR #7798, opened 2026-06-08
feat: Track telemetry opt-out via a final CLI event

olensmar:
## Summary

The Testkube CLI only sends telemetry while it is enabled. Because the telemetry send happens in `PersistentPostRun` (after the flag is flipped), **enabling** telemetry was recorded but **disabling** it was invisible — by the time the post-run hook fires, telemetry is already off and `handleTelemetry` short-circuits.

This adds an explicit opt-out signal so we can measure when users turn telemetry off:

- New `telemetry.SendTelemetryOptOutEvent` emits a dedicated, easily queryable `telemetry_opt_out` event (category `cli_command_execution`, same GA + Segment senders as other CLI events).
- `testkube disable telemetry` now sends this event **before** persisting the opt-out, and only on a genuine on→off transition (`cfg.TelemetryEnabled`), so consent is still in effect at send time and we don't emit noise when telemetry is already off.
- Send failures are non-fatal (logged only in verbose mode); the opt-out always succeeds.
- No double-counting: the post-run `handleTelemetry` sees the flag already off and stays silent.

Scope: this covers opt-outs performed via the CLI `disable telemetry` command. Opt-outs originating from the dashboard / direct API are out of scope here (would be a separate server-side change in `UpdateConfigsHandler`).

## Test plan

- [x] `go build ./pkg/telemetry/... ./cmd/kubectl-testkube/commands/telemetry/...`
- [x] `gofmt` clean, no linter errors
- [ ] Manual: `testkube enable telemetry` then `testkube disable telemetry --verbose` → observe a single `telemetry_opt_out` event sent
- [ ] Manual: `testkube disable telemetry` when already disabled → no event sent

### DIFF
--- a/cmd/kubectl-testkube/commands/telemetry/disable.go
+++ b/cmd/kubectl-testkube/commands/telemetry/disable.go
@@ -20,8 +20,34 @@ func NewDisableTelemetryCmd() *cobra.Command {
 
 			cfg, err := config.Load()
 			// Remember whether telemetry was actually on, so we only record a
-			// genuine on->off transition once we know the opt-out persisted.
+			// genuine on->off transition - and only while it is still enabled.
 			wasEnabled := err == nil && cfg.TelemetryEnabled
+
+			client, _, clientErr := common.GetClient(cmd)
+			ui.WarnOnError("getting client", clientErr)
+
+			apiDisabled := false
+			if clientErr == nil {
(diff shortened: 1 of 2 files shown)


olensmar:
@greptileai

topliceanurazvan (approved):

PR #7832, opened 2026-06-13
fix(testworkflows): pass execution tags into expression machine

caiomede-tk:
## Summary
- propagate merged execution tags into the execution expression machine during workflow resolve
- include execution tags in runner ExecutionConfig so runtime TK_CFG exposes execution.tags

## Why
Template expressions like {{ execution.tags.env }} were resolving empty because tags were not passed into ExecutionConfig before CreateExecutionMachine was built.

## Scope
- pkg/testworkflows/testworkflowexecutor/prepared.go
- pkg/runner/agent.go


caiomede-tk:
Addressed the greptile-apps follow-up about gRPC start tag propagation in stacked PR #7833 (base: fix/execution-tags-template-resolution).

caiomede-tk:
Resolved in PR https://github.com/kubeshop/testkube/pull/7833

PR #7839, opened 2026-06-15
fix(TKC-5957): Cron Scheduler misses runs. Lease leader never renews

Valentin-Marko:
## Pull request description 
Fixes constant cron-scheduler leader churn caused by the DB lease never being renewed. In the Mongo and Postgres lease backends the lease holder never wrote `RenewedAt` (`TryAcquire` returned early on an owned lease), and `leaseStatus` checked expirybefore ownership. As a result the lease expired every `DefaultMaxLeaseDuration` (~60s) even while the leader was healthy, forcing a re-election every minute.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- `leasebackend/mongo` and `leasebackend/postgres`: the holder now renews its lease
  on every check (writes `RenewedAt`) instead of returning early, so it never goes
  stale while held.
- `leaseStatus` now checks ownership before expiry, so a live holder is never treated
  as expired. Takeover-after-expiry (failover) is preserved.

## Fixes

-

### DIFF
--- a/pkg/database/postgres/queries/leases.sql
+++ b/pkg/database/postgres/queries/leases.sql
@@ -17,4 +17,8 @@ SET
     renewed_at = @renewed_at,
     updated_at = NOW()
 WHERE id = @id
+  -- Only renew/take over if, at write time, the lease is still ours or has expired.
+  -- This makes the update atomic with the ownership/expiry decision and prevents a
+  -- stalled holder from overwriting a newer leader after failover.
+  AND (identifier = @identifier OR renewed_at < @stale_threshold)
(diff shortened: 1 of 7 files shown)


Valentin-Marko:
@greptileai can you please review

Valentin-Marko:
@greptileai can you please review again?

Valentin-Marko:
@greptileai can you please review again?

courageousillumination (approved):

PR #7840, opened 2026-06-15
fix: give agent-role read access to deployment for reading its own labels

de-wim:
## Pull request description 

This is a fix for the following error I observed while working on the hosted runners feature:

```
{
 "level":"warn",
 "ts":"2026-06-15T11:45:55Z",
 "caller":"api-server/main.go:1036",
 "msg":"cannot read deployment labels",
 "deployment":"testkube-runner",
 "error":"deployments.apps \"testkube-runner\" is forbidden: User \"system:serviceaccount:testkube:agent-sa-testkube-runner\" cannot get resource \"deployments\" in API group \"apps\" in the namespace \"testkube\""
}
```

The runner attempts to fetch the labels for its own deployment and sends the ones starting with runner.testkube.io onto the cloud-api service during its self registration call. Without these RBAC changes, that call will always fail and the agent will default to a hardcoded set of labels.

See https://github.com/kubeshop/testkube/blob/af56309edb75ba27a9a90c9ebcfc521341424886/cmd/api-server/main.go#L262-L265 

And

https://github.com/kubeshop/testkube/blob/af56309edb75ba27a9a90c9ebcfc521341424886/cmd/api-server/main.go#L1098-L1109

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

- none

## Changes

- add rbac permission to grant read-only access to deployments for agent-role

## Fixes

- fix custom runner/agent labels in cloud-api

### DIFF
--- a/k8s/helm/postgresql/Chart.lock
+++ b/k8s/helm/postgresql/Chart.lock
@@ -1,6 +1,0 @@
-dependencies:
-- name: common
-  repository: oci://registry-1.docker.io/bitnamicharts
-  version: 2.40.0
-digest: sha256:c3acd4ca0b4d08db3b1c501df7e1e86a2f44f599afc26373bbba01f2341b48f2
-generated: "2026-06-01T16:00:50.214699946+02:00"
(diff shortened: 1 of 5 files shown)


de-wim:
@greptileai please look again

povilasv (approved):

PR #7848, opened 2026-06-16
fix(testworkflow-init): scope process kill/pause/resume to the init's own descendants

dejanzele:
The orchestration teardown (Kill/Pause/Resume) enumerated the entire host process tree and acted on everything except the current PID. In a pod this is harmless because init runs as PID 1 and owns every process, but when integration-test packages run in parallel off-pod one package's teardown could reach into a sibling's process tree and SIGKILL its children, surfacing as a spurious exit 137.

## How

- Add `processNode.childrenOf(pid)`, which returns a virtual root holding only the descendant subtree of the given PID.
- Kill/Pause/Resume now scope to `childrenOf(os.Getpid())` instead of the whole tree, so they only ever touch the init's own children, never the init process itself or unrelated processes.
- In a pod (init as PID 1, everything reparents to it) this is identical to the previous whole-tree behavior; off-pod it stays correctly scoped.

### DIFF
--- a/cmd/testworkflow-init/orchestration/executions.go
+++ b/cmd/testworkflow-init/orchestration/executions.go
@@ -103,9 +103,8 @@ func (e *executionGroup) Pause() (err error) {
 		output.Std.Direct().Warnf("warn: failed to pause: failed to list some processes: %v\n", err)
 	}
 
-	// Ignore the init process, to not suspend it accidentally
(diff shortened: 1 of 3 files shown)


dejanzele:
@greptileai

buarki (approved):

vsukhin (approved):

PR #7886, opened 2026-06-26
feat: Add is_silent flag to workflow execution telemetry

olensmar:
## Summary

- Adds an `is_silent` boolean to the `testkube_api_run_test_workflow` telemetry event so silent executions can be distinguished in analytics dashboards.
- The flag is populated from `execution.SilentMode` in the workflow-execution telemetry listener.
- `is_silent` is `true` **only** when an execution was triggered as fully silent — the `--silent` CLI flag, the dashboard silent toggle, or a workflow-level silent execution — all of which resolve to `NewSilenceAllSilentMode()` (every aspect silenced). Partial silencing of individual aspects (e.g. the deprecated `--disable-webhooks`, which sets only `Webhooks`) does **not** set it.

## Background

Previously there was no way to tell from workflow execution telemetry whether an execution was triggered as silent. `SilentMode` is persisted on the execution record but was never surfaced to the analytics event, and the telemetry listener does not (and still does not) silence telemetry itself.

## Changes

- `pkg/telemetry/payload.go`: add `IsSilent` to `Params` (`json:"is_silent,omitempty"`) and to `RunWorkflowParams`; map it in `NewRunWorkflowPayload`.
- `pkg/event/kind/testworkflowexecutiontelemetry/listener.go`: compute `is_silent` from `execution.SilentMode` via a new `isSilentExecution` helper (requires all aspects enabled) and pass it through.

## Test plan

- [x] `go build ./...`
- [x] `make lint`
- [ ] Trigger a workflow with `--silent` and confirm the telemetry event carries `is_silent=true`
- [ ] Trigger with `--disable-webhooks` only and confirm `is_silent` is absent/false

### DIFF
--- a/pkg/event/kind/testworkflowexecutiontelemetry/listener.go
+++ b/pkg/event/kind/testworkflowexecutiontelemetry/listener.go
@@ -163,6 +163,8 @@ func (l *testWorkflowExecutionTelemetryListener) sendRunWorkflowTelemetry(ctx co
 		}
 	}
 
+	isSilent := isSilentExecution(execution.SilentMode)
+
 	out, err := telemetry.SendRunWorkflowEvent("testkube_api_run_test_workflow", telemetry.RunWorkflowParams{
 		RunParams: telemetry.RunParams{
 			AppVersion: version.Version,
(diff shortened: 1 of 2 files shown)


xoscar (approved):
🤐

PR #4344, opened 2023-09-07
chore: remove refs to deprecated io/ioutil

testwill:
## Checklist (choose whats happened)

- [ ] breaking change! (describe)
- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

PR #7890, opened 2026-06-27
fix(mcp): recover from invalid step filter in fetch_execution_logs

niyiomotoso:
## What

When `fetch_execution_logs` is given a `step` that does not exist in the targeted log (commonly when the model confuses a parallel/worker ref with an inner step name), the control plane returns 400. The MCP client surfaced only `API returned status 400` and the tool dead-ended.

## Change

- `api.go`: read the error response body on non-2xx and include the `problem+json` detail in the returned error; record the debug URL after query params are appended; capture the response body in debug data.
- `executions.go`: on an invalid-step 400, retry once without the `step` filter and annotate the result; otherwise surface the detailed error including the available steps.

## Tests

- `api_test.go`: error-body propagation, debug URL includes query params, `extractErrorDetail`.
- `executions_test.go`: retry-without-step recovery, detail surfaced when retry also fails, and no retry on non-step errors.

Ref: TKC-6070. Pairs with the server-side change in kubeshop/testkube-cloud-api (#4344) that puts the invalid-step detail in the 400 body.


niyiomotoso:
@greptile review latest changes

topliceanurazvan (approved):

PR #7892, opened 2026-06-29
fix(runner): retry live-log Summary while the Job is not yet visible

dejanzele:
The live notifications path called worker.Summary once and gave up on any error. On a first execution whose Job is not yet list-visible, Summary returns controller.ErrJobTimeout and the live log session dies, so live logs never start.

This retries Summary while the Job is not yet visible, bounded by the same budget the archival path already uses, and returns promptly on context cancellation. Other errors and successful results return immediately.

### DIFF
--- a/pkg/runner/agent.go
+++ b/pkg/runner/agent.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


topliceanurazvan (approved):
LGTM!

PR #7893, opened 2026-06-29
test(controlplaneclient): prove stream resume replays the gap on reconnect

dejanzele:
Characterization test for the notification stream session manager. It publishes a sequence of notifications, simulates a subscription drop, reconnects with a resume-after-seqNo, and asserts the existing session is reused and replays exactly the missing frames, plus the resume-unavailable path. No production change.


topliceanurazvan (approved):

dejanzele:
not needed

PR #7904, opened 2026-06-30
feat(mcp): add list_credentials tool

niyiomotoso:
## What

Adds a read-only `list_credentials` MCP tool so an AI agent (or any MCP client) can discover which credential **references** exist in the current environment — name, type, scope — and use them directly in `credential("...")` workflow expressions. It **never returns values**, and is the read-before-ask complement to `request_credential`.

## Change (testkube side)

- `pkg/mcp/formatters/credentials.go` — secret-safe projection: reads only name/type/reference/scope (never `value`), filters non-referenceable execution-scoped credentials, renders the ready-to-use `credential("<ref>")` expression.
- `pkg/mcp/tools/credentials.go` + `descriptions.go` — the tool, its `CredentialsLister` interface, and an LLM-facing description stating values are never returned.
- `pkg/mcp/api.go` — `APIClient.ListCredentials` (GET env-scoped `/credentials?filter=all`); added to the aggregate `Client` interface; registered in the standalone MCP server.

## Scope

Org + environment coverage (reuses the existing org→env merge). No backend changes. Resource-group/workflow-scoped listing is deferred.

## Tests

- `formatters/credentials_test.go` — happy path, never-returns-secret-values, drops variable values, omits non-referenceable creds, empty input, reference-expression rendering.
- `tools/credentials_test.go` — handler happy path + client-error → tool error.
- `api_test.go` — asserts the env-scoped `/credentials?filter=all` request.

Ref: TKC-5618. Pairs with kubeshop/testkube-cloud-api (in-process MCP host wiring); that repo's `go.mod` will bump to the merged version of this PR.

### DIFF
--- a/pkg/mcp/api.go
+++ b/pkg/mcp/api.go
@@ -654,6 +654,15 @@ func (c *APIClient) ListLabels(ctx context.Context) (string, error) {
 	})
 }
 
+func (c *APIClient) ListCredentials(ctx context.Context) (string, error) {
+	return c.makeRequest(ctx, APIRequest{
+		Method:      "GET",
+		Path:        "/credentials",
+		Scope:       ApiScopeOrgEnv,
+		QueryParams: map[string]string{"filter": "all"},
+	})
+}
+
 func (c *APIClient) ListResourceGroups(ctx context.Context) (string, error) {
 	return c.makeRequest(ctx, APIRequest{
 		Method: "GET",
--- a/pkg/mcp/api_test.go
+++ b/pkg/mcp/api_test.go
@@ -130,6 +130,38 @@ func TestAPIClient_DebugInfo_URLIncludesQueryParams(t *testing.T) {
 	assert.Contains(t, gotURL, "tail=50")
 }
 
+func TestAPIClient_ListCredentials_HitsEnvScopedFilterAll(t *testing.T) {
+	const (
+		orgID = "org-1"
+		envID = "env-1"
+		token = "tok"
+		body  = `{"elements":[{"name":"github-access-token","type":"secret","reference":"github-access-token"}]}`
+	)
+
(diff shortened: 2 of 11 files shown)


niyiomotoso:
Closing — not ready for review yet; testing/confirming locally first.

olensmar:
@niyiomotoso could we add this tool to the hosted MCP server only?

niyiomotoso:
@olensmar done — `list_credentials` is now gated to the **hosted MCP surfaces only**, not the local `testkube` CLI.

I added a config flag `MCPServerConfig.IncludeCredentialTools`; `NewMCPServer` registers the tool only when it's set (same conditional-registration pattern we already use for the query tools). It's set per entrypoint in code, not via env:
- **mcp-bridge** (backs the AI chat) → `cfg.IncludeCredentialTools = true`
- **control-plane in-process `/mcp` endpoint** → always registered (separate registration)
- **local `testkube` CLI** → left `false` (default) → tool not exposed

**Net surface coverage:**

| Surface | Exposes `list_credentials`? |
|---|---|
| Local `testkube` CLI MCP | ❌ no |
| AI chat (mcp-bridge) | ✅ yes |
| Hosted cloud `/mcp` endpoint | ✅ yes |

Two notes:
- It's the **first** tool intentionally scoped by surface (today the CLI and hosted registries are otherwise identical), but it reuses the existing `cfg`-gated `if { AddTool(...) }` pattern rather than introducing anything new.
- **Zero deployment burden** — the flag is a hardcoded constant at each binary's entrypoint (like the bridge already hardcodes `cfg.Source = "mcp-bridge"`), so there's no new env var to plumb through dev/staging/prod.

Added a unit test covering both the registered and omitted cases.

niyiomotoso:
@greptile review latest commits

courageousillumination:

PR #7912, opened 2026-07-02
fix: [TKC-6152] subtract paused time as milliseconds in HealDuration

dejanzele:
## How

HealDuration converted PausedMs to a time.Duration with a factor of 1e3, which yields microseconds instead of milliseconds: a two minute pause shortened the reported duration by 120 microseconds. The conversion now uses time.Millisecond, and a unit test pins both the paused and unpaused paths.

Refs TKC-6152

### DIFF
--- a/pkg/api/v1/testkube/model_test_workflow_result_extended.go
+++ b/pkg/api/v1/testkube/model_test_workflow_result_extended.go
@@ -361,7 +361,7 @@ func (r *TestWorkflowResult) HealDuration(scheduledAt time.Time) {
 			queuedAt = scheduledAt
 		}
 		totalDuration := r.FinishedAt.Sub(scheduledAt)
-		duration := totalDuration - time.Duration(1e3*r.PausedMs)
+		duration := totalDuration - time.Duration(r.PausedMs)*time.Millisecond
 		if totalDuration < 0 {
 			totalDuration = time.Duration(0)
 		}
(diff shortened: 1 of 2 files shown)


povilasv (approved):

dejanzele:
The Integration Tests failure (TestInitProcessCore_Integration/ExecuteCommand, exit 137) is not caused by this change. Evidence:

- The identical failure reproduces on pristine origin/main (0a487fb90, the same commit this branch is rebased on) in a clean Linux environment: golang:1.26 container, fresh build, same test command. Main and this branch fail byte-for-byte identically there, so the diff is not the variable.
- The suite passes on this branch on macOS locally (10.2s, all subtests green).
- The failure is deterministic per environment: three CI reruns on depot-ubuntu-24.04 runners failed the same way, dying at 0.00s with SIGKILL before the step command runs, while main CI runs from the same morning pass.

So the discriminator is the execution environment, not the code. This looks related to the runner/cache work happening today (ci/tighten-depot-cache). Happy to rerun once that lands.

dejanzele:
Root cause of the Integration Tests failure is found and fixed in #7924: RunInit's deferred context cancellation fired the abort handler asynchronously after the run finished, poisoning the next run's process-global execution group. Whether a suite passed depended on goroutine scheduling latency, which is why it failed on depot runners and docker but passed on macOS and GitHub-hosted runners. Once #7924 merges, a rebase or rerun here should go green.

PR #7917, opened 2026-07-03
fix: make install.sh POSIX sh compatible and verify release checksums

olensmar:
## Summary

- Makes `install.sh` POSIX `sh` compatible — the documented install command (`curl -sSLf https://get.testkube.io | sh`) previously failed on Debian/Ubuntu where `sh` is dash, because the script used bash-only constructs (`[[ ]]`, `=~`, `set -o pipefail`). This also makes the upgrade hint surfaced by the CLI's install-source detection (`curl ... | sh`) actually work as documented.
- Removes dead code paths: `i386` assets are no longer published (goreleaser only builds amd64/arm64), the `Windows` uname case never matched (Git Bash reports `MINGW64_NT-...`), and the beta path grepped tag names for "beta" which has matched nothing since the v1.17 era — it now uses GitHub's `prerelease` flag with the same fallback to latest.
- Fails with a clear error (suggesting `TESTKUBE_VERSION=...`) when the version cannot be resolved from the GitHub API, instead of silently substituting `1` into the download URL.
- Verifies the downloaded tarball against the release's `checksums.txt` (via `sha256sum`/`shasum`); fails hard on mismatch, skips with a warning when no tool or checksum entry is available (e.g. pre-2.x releases).
- Downloads and extracts in a `mktemp -d` workdir with a cleanup trap instead of the caller's current directory.

The install dir stays hardcoded at `/usr/local/bin`, preserving the contract with the CLI's install-source detection (`cmd/kubectl-testkube/commands/common/install_source.go`).

## Test plan

- [x] `sh -n` and `dash -n` syntax checks pass
- [x] End-to-end install under `dash` (install dir redirected to a temp dir): latest release 2.11.0 downloaded, checksum verified, binary + `testkube`/`tk` symlinks created
- [x] `TESTKUBE_VERSION=2.10.1` → downloads tag `2.10.1`, checksum verified
- [x] `TESTKUBE_VERSION=1.17.68` → tag correctly normalized to `v1.17.68`, checksum verification gracefully skipped (release predates `checksums.txt`)
- [x] `install.sh beta` → no pre-releases found, falls back to latest

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,138 +1,182 @@
-#!/bin/bash
-set -eo pipefail
+#!/bin/sh
+# POSIX sh compatible — this script is documented as `curl -sSLf https://get.testkube.io | sh`,
+# so it must not rely on bash-only constructs ([[ ]], =~, set -o pipefail, etc.).
+set -e
 
 echo "Getting kubectl-testkube plugin"
 
-if [ ! -z "${DEBUG}" ];
-then set -x
+if [ -n "${DEBUG}" ]; then
+  set -x
 fi
 
+GITHUB_API="https://api.github.com/repos/kubeshop/testkube"
+RELEASES_BASE_URL="https://github.com/kubeshop/testkube/releases/download"
+INSTALL_DIR=/usr/local/bin
+
 _check_required_tools() {
-  local MISSING_TOOLS=""
-  for CMD in curl jq; do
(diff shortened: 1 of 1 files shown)


olensmar:
@greptile

courageousillumination (approved):

PR #7921, opened 2026-07-06
fix: logging level adjustments and configurable frequency

courageousillumination:
- Logging updates
   - Removed `failed to list webhooks` log from WebhookLoader when webhooks fail to load entirely. This function already returns the error and the higher level orchestrator will log accordingly.
   - Remove `process notifications error` logs from utils/processesNotifications. Similar reasoning, these return the errors and they are already logged higher up.
   - Downgrade `webhook template is disabled` to a debug. This is intentional configuration of the webhook, not an error.
   - Downgrade `error loading listeners` to a warn. We will attempt self healing in the next reconcile loop.
   - Downgrade `error running agent connection` to warn. Same as above.
   - Downgrade `error listing cloud workflow triggers` to warn. Similar to above, this will attempt a self heal in the next loop.
   - Downgrade `error listing cloud test triggers` to warn. Same as above.
   - Downgrade `connecting to aborted execution` to warn. This is an abnormal situation, but is handled via the returned error. 
- Other updates
   -  Make the webhook/listener reconcile interval configurable and extend the default to 5 seconds. 1 second intervals can lead to log spam, especially in the case of transient network issues.

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -189,11 +189,13 @@ func main() {
 	// Resolve the lease check interval once and reuse it across all lease consumers
 	// (events emitter, triggers, leader coordinator). Clamps unsafe values.
 	leaseCheckInterval := resolveLeaseCheckInterval(cfg.LeaseCheckInterval)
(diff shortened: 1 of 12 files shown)


vsukhin (approved):

courageousillumination:
@greptileai

PR #7933, opened 2026-07-08
fix(runner): resume live logs across reconnects instead of resetting the stream

dejanzele:
## How

The agent's notification stream session manager (the replay buffer + resume cursors keyed by executionId+streamId) was created inside `processNotifications`, which runs once per gRPC connection. So every reconnect built a fresh, empty manager and the pod-log source was tied to the per-connection context. When the agent-to-control-plane stream dropped (a gateway timeout, a deploy, a network blip), the buffer was discarded and the source torn down. The next reconnect's `resumeAfterSeqNo` hit an empty manager, got a fresh session, and was answered with `resume_unavailable`, so the client reset its cursor and re-streamed live-only, losing the logs between the drop and the reconnect.

This lifts the session managers to the client, one per stream kind (workflow / service / parallel step), created once with the agent's long-lived context and reused across reconnects. `runSource` now runs under that context, so a session's pod-log source keeps filling its buffer while no reader is connected. A reconnect with the same streamId then finds the live session and replays from the cursor (`resume` available) instead of resetting.

Added `TestNotificationStreamSessionSurvivesReaderDropAndResumesLive`: a source keeps producing after its reader disconnects, and a later attach with the cursor resumes and replays the gap. Existing session-manager tests pass unchanged under `-race`.

### DIFF
--- a/pkg/controlplaneclient/client.go
+++ b/pkg/controlplaneclient/client.go
@@ -2,6 +2,7 @@ package controlplaneclient
 
 import (
 	"strings"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


topliceanurazvan (approved):
LGTM!

PR #7934, opened 2026-07-08
feat: [TKC-6223] Add registration token rotation CLI

caiomede-tk:
## Summary

Adds CLI support for rotating Testkube environment registration tokens through the Control Plane API.

## Changes

- Adds `testkube agent rotate-registration-token [environment-id-or-name]`.
- Defaults to the configured environment when no environment is passed.
- Adds `--grace-period` and `--yes` flags.
- Prints the new registration token and instructions for updating `runner.register.token` or a referenced Secret.
- Adds cloud client and command coverage.

## Linear

TKC-6223: https://linear.app/kubeshop/issue/TKC-6223/add-cli-support-for-environment-registration-token-rotation

## Validation

- `go test ./pkg/cloud/client -run 'TestEnvironmentsClient_(Create|RotateRegistrationToken)'`
- `go test ./cmd/kubectl-testkube/commands/agents ./cmd/kubectl-testkube/commands`
- `make lint`

### DIFF
--- a/cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
+++ b/cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
@@ -19,7 +18,7 @@ func NewRotateRegistrationTokenCommand() *cobra.Command {
 	var yes bool
 
 	cmd := &cobra.Command{
-		Use:   "rotate-registration-token [environment-id-or-name]",
+		Use:   "rotate-registration-token [environment-id]",
 		Short: "Rotate the registration token for an environment",
 		Args:  cobra.MaximumNArgs(1),
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
@@ -33,23 +32,11 @@ func NewRotateRegistrationTokenCommand() *cobra.Command {
 			ui.ExitOnError("loading config", err)
 
 			envID := cfg.CloudContext.EnvironmentId
-			envName := envID
 			if len(args) == 1 {
-				envs, err := common.GetEnvironments(cfg.CloudContext.ApiUri, cfg.CloudContext.ApiKey, cfg.CloudContext.OrganizationId, cfg.SkipTLS || cfg.CloudContext.SkipTLS)
-				ui.ExitOnError("getting environments", err)
-				envID = ""
-				for _, env := range envs {
-					if env.Id == args[0] || env.Name == args[0] {
-						envID, envName = env.Id, env.Name
-						break
-					}
-				}
-				if envID == "" {
-					ui.ExitOnError("finding environment", errors.Errorf("environment %q not found", args[0]))
-				}
+				envID = args[0]
 			}
 
-			if !yes && !ui.Confirm(fmt.Sprintf("Rotate registration token for environment '%s'?", envName)) {
+			if !yes && !ui.Confirm(fmt.Sprintf("Rotate registration token for environment ID '%s'?", envID)) {
 				return
 			}
 
--- a/cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
+++ b/cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
@@ -1,0 +1,73 @@
+package agents
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common/validator"
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/config"
+	cloudclient "github.com/kubeshop/testkube/pkg/cloud/client"
+	"github.com/kubeshop/testkube/pkg/ui"
+)
+
+func NewRotateRegistrationTokenCommand() *cobra.Command {
+	var gracePeriod string
+	var yes bool
+
+	cmd := &cobra.Command{
+		Use:   "rotate-registration-token [environment-id-or-name]",
+		Short: "Rotate the registration token for an environment",
+		Args:  cobra.MaximumNArgs(1),
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			cfg, err := config.Load()
+			ui.ExitOnError("loading config", err)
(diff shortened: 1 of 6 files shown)


cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
```diff
@@ -0,0 +1,73 @@
...
+	cloudclient "github.com/kubeshop/testkube/pkg/cloud/client"
+	"github.com/kubeshop/testkube/pkg/ui"
+)
+
+func NewRotateRegistrationTokenCommand() *cobra.Command {
+	var gracePeriod string
+	var yes bool
+
+	cmd := &cobra.Command{
+		Use:   "rotate-registration-token [environment-id-or-name]",
```
dejanzele: `[environment-id-or-name]` instead of name, I'd support ID or slug, as names can have spaces, I don't think we have guarantees around uniquneess

pkg/cloud/client/environments.go
```diff
@@ -33,6 +41,40 @@ type EnvironmentsClient struct {
...
+	GracePeriod       string    `json:"gracePeriod"`
+	OldTokenExpiresAt time.Time `json:"oldTokenExpiresAt"`
+}
+
 func (c EnvironmentsClient) Create(env Environment) (Environment, error) {
 	return c.RESTClient.Create(env, "/organizations/"+env.Owner+"/environments")
 }
+
+func (c EnvironmentsClient) RotateRegistrationToken(envID, gracePeriod string) (RotateRegistrationTokenResponse, error) {
+	path := c.BaseUrl + c.Path + "/" + envID + "/registration-token"
```
dejanzele: Ideally I'd like us to use the Go's standard `net/url` package and the `url.URL` and avoid manually doing string concatenation for URLs.

Even better, the endpoints should not be magic string, but kept as consts in a shared location

pkg/cloud/client/environments.go
```diff
@@ -33,6 +41,40 @@ type EnvironmentsClient struct {
...
+	if gracePeriod != "" {
+		path += "?gracePeriod=" + url.QueryEscape(gracePeriod)
+	}
+	req, err := nethttp.NewRequestWithContext(context.Background(), nethttp.MethodDelete, path, nil)
```
dejanzele: Maybe propagate the context from the command to this function and use that, instead of `context.Background`?

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
```diff
@@ -0,0 +1,73 @@
...
+				}
+			}
+
+			if !yes && !ui.Confirm(fmt.Sprintf("Rotate registration token for environment '%s'?", envName)) {
+				return
+			}
+
+			client := cloudclient.NewEnvironmentsClient(cfg.CloudContext.ApiUri, cfg.CloudContext.ApiKey, cfg.CloudContext.OrganizationId, cfg.SkipTLS || cfg.CloudContext.SkipTLS)
+			result, err := client.RotateRegistrationToken(envID, gracePeriod)
+			ui.ExitOnError("rotating environment registration token", err)
```
dejanzele: We have the concept of `CLIError` which pretty prints errors, gives hints... please take a look if it can be reused

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
```diff
@@ -0,0 +1,73 @@
...
+
+			ui.Success("Registration token rotated successfully")
+			fmt.Println()
+			ui.Warn("New Registration Token:", result.RegistrationToken)
```
dejanzele, line 57: Do we need to print the runner registration token?

Isn't it automatically reused for all new runners?

pkg/cloud/client/environments.go
dejanzele: Also do we use a DELETE method to rotate a secret?
That seems a bit unintuitive.

pkg/cloud/client/environments.go
```diff
@@ -33,6 +41,40 @@ type EnvironmentsClient struct {
...
+	if err != nil {
+		return RotateRegistrationTokenResponse{}, err
+	}
+	req.Header.Add("Authorization", "Bearer "+c.Token)
```
dejanzele: Don't we have a helper for this? It seems a bit tedious

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
caiomede-tk: Resolved, envId enforced.

pkg/cloud/client/environments.go
caiomede-tk: There is no helper, but I added a reusable function called setBearerAuth in `testkube/pkg/cloud/client/rest.go` , but I didn't replace the other go files and functions that are using same logic to add Authorization header.

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
caiomede-tk: It's needed because the Registration Token is directly referenced in the agents values.yaml, hardcoded or through a secret, so the idea is that the person who rotates the token can update the values.yaml or the secret vault

pkg/cloud/client/environments.go
caiomede-tk: I moved to POST

caiomede-tk:
@greptile-apps please review again

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
```diff
@@ -0,0 +1,68 @@
...
+			validator.PersistentPreRunVersionCheck(cmd, common.Version)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			cfg, err := config.Load()
+			ui.ExitOnError("loading config", err)
+
+			envID := cfg.CloudContext.EnvironmentId
+			if len(args) == 1 {
+				envID = args[0]
+			}
```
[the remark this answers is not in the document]
caiomede-tk: Environment name is not expected to be used.

cmd/kubectl-testkube/commands/agents/rotate_registration_token.go
```diff
@@ -0,0 +1,72 @@
...
+			cfg, err := config.Load()
+			ui.ExitOnError("loading config", err)
+			common.UiContextHeader(cmd, cfg)
+			validator.PersistentPreRunVersionCheck(cmd, common.Version)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			cfg, err := config.Load()
+			ui.ExitOnError("loading config", err)
+
+			envID := cfg.CloudContext.EnvironmentId
```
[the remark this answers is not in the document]
caiomede-tk: Environment name is not expected to be used.

xoscar (approved):

PR #7947, opened 2026-07-09
feat(runner): add live-log streaming metrics

dejanzele:
## How

Instruments the agent's live-log notification stream (the session manager, replay buffer, and resume path in `pkg/controlplaneclient`) with Prometheus metrics so we can build dashboards and alerts for live-log streaming. Pure observability, additive, no behaviour change.

Metrics are registered on the default registry, which the agent already serves at `/metrics`, and are labelled by stream `kind` (workflow, parallel, service).

Gauges are read at scrape time from the live session managers through one collector, so they always equal the state the managers hold and no code path has to keep them in step:

- `testkube_live_log_sessions` (`kind`, `state`=active|done) — live and draining session counts.
- `testkube_live_log_replay_bytes` (`kind`) — total buffered replay memory, the number to graph against the buffer budget.
- `testkube_live_log_subscribers` (`kind`) — concurrent viewers.

Counters and the histogram record events where they happen:

- `testkube_live_log_sessions_created_total` (`kind`).
- `testkube_live_log_sessions_evicted_total` (`kind`, `reason`=ttl|errored|replaced) — every removal from the manager goes through one function, so each eviction is counted once with its reason.
- `testkube_live_log_resume_total` (`kind`, `result`=available|unavailable) — how often a reconnect actually resumed versus fell back to a full reset. This is the signal that the resume path is working in production.
- `testkube_live_log_source_duration_seconds` (`kind`, `result`=ok|error) — buckets span 1 s to 2 h, because a source lives as long as the execution it follows.

The replay buffer limits become fields on the manager so the tests can set them. The tests live next to the other session manager tests in `testworkflows_test.go` and assert the manager's stats snapshot directly; the collector has one test against its exposition text.


pkg/controlplaneclient/metrics_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kubeshop/testkube/pkg/api/v1/testkube"
+	"github.com/kubeshop/testkube/pkg/cloud"
+	"github.com/kubeshop/testkube/pkg/repository/channels"
+)
+
+func TestLiveLogMetricsMoveOnAttachAndPublish(t *testing.T) {
```
courageousillumination, line 16: Do you think this metrics_test should be testing the implementation of the notification stream session manager? It feels like the metrics just defines the metrics and we should have the tests living alongside other tests for the stream manager.

courageousillumination (approved):

pkg/controlplaneclient/metrics_test.go
dejanzele: Makes sense, I'll port it so the metrics are tested through an actual flow and test it in `testworklows_test.go`

dejanzele:
@greptileai

PR #7960, opened 2026-07-15
chore: [TKC-6265] clarify crd-only flag and namespace semantics in workflow get commands

dejanzele:
Linear: https://linear.app/kubeshop/issue/TKC-6265

Rewords the help of `get testworkflows` and `get testworkflowtemplates`: `--crd-only` renders API results as CRD YAML and never reads cluster CRDs, and the namespace flag only applies in kubeconfig context. Error prefixes drop the namespace in cloud context since it does not scope queries there. Only help strings and messages change, behavior stays the same.


topliceanurazvan (approved):
LGTM!

PR #7961, opened 2026-07-15
feat: [TKC-6266] hint about disabled CRD sync when test workflow lookup returns 404

dejanzele:
Linear: https://linear.app/kubeshop/issue/TKC-6266

When `testkube get testworkflow <name>` gets a 404, the CLI now prints a structured CLIError (TKERR-1701). The hint explains that the workflow may exist only as a cluster CRD while synchronization to the Control Plane is disabled, and points at kubectl and `testkube get context` to check. Non-404 errors keep the existing path.


topliceanurazvan (approved):
LGTM! As follow-ups, should we do the same thing for templates, test triggers, etc..?

dejanzele:
@topliceanurazvan ideally we should do a big refactor and migrate a lot of errors to use the CLIError as it also can output hints, how to resolve, it tracks error telemetry better...

I'll at least try to document the followups

PR #7962, opened 2026-07-15
ci: use env indirection for expressions in workflow run scripts

dejanzele:
Moves `${{ }}` expressions out of `run:` script bodies into step-level `env:` blocks and references them as quoted shell variables, so untrusted values are no longer spliced into the script text that the shell parses. Registry login secrets are now piped to `helm registry login` via `--password-stdin` with `printf` instead of appearing on the command line.

Linear: TKC-6268


ypoplavs (approved):

dejanzele:
@greptileai

PR #7982, opened 2026-07-18
chore: remove dead Tests v1 packages and five single-use dependencies

UditDewan:
## Pull request description 

Two focused cleanups, split into one commit each, with **no behavior change** intended anywhere. Net: **+81 / −1,654 lines** and **five fewer direct dependencies**.

### 1. Remove dead Tests v1 leftovers (`10b8669`)

The following packages have had no callers since the Tests v1 functionality was removed:

| Removed | Evidence |
|---|---|
| `api/script/v1`, `api/script/v2`, `api/template/v1` | Only importer was `pkg/operator/client/client.go` (scheme registration) |
| `pkg/operator/client/scripts`, `.../scripts/v2`, `.../templates` | Zero importers anywhere in the repo, including tests |
| `scriptv1`/`scriptv2`/`templatev1` registrations in `pkg/operator/client/client.go` | The registered types are never read or written through this client |
| `utils.ContainsTag` | Its only callers were the removed script clients |

**Deliberately kept:** the generated CRD manifests under `k8s/crd` and the Helm charts are untouched. Removing CRDs from the charts could cause Helm to delete existing `Script`/`Template` custom resources on upgrade, which is a user-facing decision that deserves its own discussion — happy to follow up if you want them gone too.

### 2. Replace single-use dependencies with stdlib equivalents (`2ab0266`)

| Dependency | Usage found | Replacement |
|---|---|---|
| `github.com/go-errors/errors` | two `Join` calls in `testworkflowexecutor/prepared.go` | stdlib `errors.Join` (identical semantics) |
| `github.com/json-iterator/go` | one `Marshal` + one `Unmarshal` in `pkg/controlplane/utils.go` | `encoding/json`, also inlining the `toJSON`/`fromJSON` wrappers around it |
| `golang.org/x/exp/maps` | `maps.Keys` in 3 files | stdlib `maps.Keys` + `slices.Collect` (same unordered-slice semantics) |
| `github.com/montanaflynn/stats` | one `PercentileNearestRank` call in `pkg/repository/common/metrics.go` | a ~8-line local nearest-rank implementation; the existing `Test_ShouldCalculatePercentile_AndGiveNearestIndexResult` pins the exact expected P50/P90/P95/P99 values and passes unchanged |
| `github.com/moogar0880/problems` | `pkg/problem` wrapped it to get the plain RFC 9457 struct + constructor | local struct with identical JSON tags, so the wire format is unchanged; a new `pkg/problem/problem_test.go` pins the serialized form |
| `github.com/Masterminds/semver` **v1** (archived release line) | 3 files | migrated to `semver/v3`, which was already in the module graph via sprig; every method used (`NewVersion`, `LessThan`, `Prerelease`, `IncMajor/Minor/Patch`, `SetPrerelease`) has an identical signature in v3 |

`json-iterator` and `x/exp` remain in `go.mod` as `// indirect` because other dependencies still require them transitively; they are no longer part of this codebase's direct API surface. All replacement code is written from scratch against the relevant specs (RFC 9457; the textbook nearest-rank percentile definition), not copied from the removed libraries.

### Verification

- `go build ./...` passes on the full module
- `go test` passes for every touched package: `pkg/utils`, `pkg/controlplane`, `pkg/repository/common`, `pkg/problem`, `pkg/semver`, `pkg/api/v1/client`, `pkg/testworkflows/testworkflowexecutor`, `pkg/testworkflows/executionworker/...`, `cmd/tcl/testworkflow-toolkit/...`, `cmd/kubectl-testkube/commands/common/...` (the only failure is the pre-existing, unrelated `Test_readCopyFiles`, which hardcodes `/tmp` and fails on any Windows checkout — reproducible on `main`)
- `go mod tidy` is clean
- `grep\'d the whole tree (code, docs, manifests) for references to every removed package and symbol — none remain

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

- none — all removed code had zero callers, and every replacement preserves wire format and semantics

## Changes

- removed dead `api/script`, `api/template`, and `pkg/operator/client/{scripts,templates}` packages and their scheme registrations
- removed unused `utils.ContainsTag`
- replaced 5 single-use direct dependencies with standard-library equivalents (one of them by upgrading to a maintained major version already present in the module graph)
- added a test pinning the RFC 9457 JSON wire format of `pkg/problem`

## Fixes

- follow-up candidates spotted while auditing, intentionally out of scope to keep this reviewable: consolidating the two file-type detection deps (`h2non/filetype` + `gabriel-vasile/mimetype`), the two cron parsers (`robfig/cron` + `adhocore/gronx`), and the two retry/backoff deps (`cloudflare/backoff` + `avast/retry-go`) — each changes runtime behavior slightly, so they deserve separate PRs if wanted


vsukhin (approved):

vsukhin:
thank you @UditDewan for cleaning things up!  lgtm

vsukhin:
@copilot resolve the merge conflicts in this pull request

PR #8009, opened 2026-07-23
fix: correct semver Lte semantic equality and GetNewest empty-input panic

UditDewan:
## What

Two correctness fixes in `pkg/semver`:

### 1. `Lte` compared equality by string, not semantically

`Lte` fell back to raw string equality to decide the "or equal" case:

```go
return ok || version1 == version2, nil
```

So semantically-equal versions written differently were reported as **not** less-than-or-equal:

- `Lte("1.0", "1.0.0")` → `false` (should be `true`)
- `Lte("v1.2.3", "1.2.3")` → `false` (should be `true`)

This is used by the kubectl / Kubernetes dependency validator (`pkg/diagnostics/validators/deps`), where a required version like `1.0.0` compared against a client-reported `1.0` could incorrectly flag a valid install as too old.

`Lte` now compares versions semantically via `Compare(...) <= 0`.

### 2. `GetNewest` panicked on empty / all-invalid input

```go
return semversions[0].String()
```

If the input list is empty, or contains no parseable semver strings, `semversions` is empty and indexing `[0]` panics with an index-out-of-range. `GetNewest` now returns an empty string in that case.

## Tests

Added tests covering:
- `Lte` semantic equality across differing string forms, strict less-than, greater-than, and invalid-input error paths.
- `GetNewest` on `nil`, empty, and all-invalid input.

Existing tests continue to pass.

### DIFF
--- a/pkg/semver/version.go
+++ b/pkg/semver/version.go
@@ -109,12 +109,18 @@ func Lt(version1, version2 string) (bool, error) {
 
 // Lte checks if version1 is less-than or equal version2, returns error in case of invalid version string
 func Lte(version1, version2 string) (bool, error) {
-	ok, err := Lt(version1, version2)
+	v1, err := semver.NewVersion(version1)
+	if err != nil {
+		return false, err
+	}
+	v2, err := semver.NewVersion(version2)
 	if err != nil {
 		return false, err
 	}
 
-	return ok || version1 == version2, nil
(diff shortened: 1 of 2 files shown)


vsukhin:
thank you @UditDewan nice catch!

vsukhin (approved):

PR #8023, opened 2026-07-27
fix(testkube-runner): grant CRD read access to agent ServiceAccount

jose-tk:
## Summary

- The testkube-runner agent watches CustomResourceDefinition (cluster-scoped) at startup unconditionally, but the chart never granted that permission to the agent's ServiceAccount, not even when listener.enabled: true or gitops.enabled: true.
- The existing watchers-role-* (Cluster)Role in templates/role.yaml is only created when listener.enabled or gitops.enabled is true, and its rules (from the testkube-runner.listener.watchers.rules helper) never cover apiextensions.k8s.io/customresourcedefinitions regardless.
- Result: every agent pod logs a recurring "Failed to watch ... CustomResourceDefinition ... forbidden" error, on every deployment of this chart. Confirmed reproducing on multiple GKE clusters, both a dev environment and production, all running 2.12.0, independent of listener/gitops config.
- Doesn't appear to break functionality (agent connects to control plane, recovers/executes workflows fine), but it's constant log noise and a wasted retry loop.

## Change

- templates/role.yaml: add a ClusterRole (crd-reader-{{ .Release.Name }}) granting get/list/watch on apiextensions.k8s.io/customresourcedefinitions, gated only by pod.serviceAccount.autoCreate (same gate as the base agent-role), independent of listener/gitops.
- templates/rolebinding.yaml: add the matching ClusterRoleBinding (crd-reader-rb-{{ .Release.Name }}) for the agent ServiceAccount.

No values.yaml changes needed, no new flags introduced.

## Test plan

- [x] helm lint . passes
- [x] helm template . renders the new ClusterRole/ClusterRoleBinding with default values
- [x] Renders correctly with listener.enabled=true too (existing watchers role/binding still render alongside)
- [x] pod.serviceAccount.autoCreate=false correctly omits the new resources, consistent with the rest of the agent RBAC
- [ ] After merge/release, confirm the "Failed to watch ... CustomResourceDefinition" error no longer appears in agent logs

### DIFF
--- a/k8s/helm/testkube-runner/templates/role.yaml
+++ b/k8s/helm/testkube-runner/templates/role.yaml
@@ -37,6 +37,34 @@ rules:
 
 ---
 
+# Allow Agent to discover installed CRDs at startup (cluster-scoped, so this
+# must be a ClusterRole regardless of the listener/gitops watchers feature).
+{{- if .Values.pod.serviceAccount.autoCreate }}
+apiVersion: {{ include "global.capabilities.rbac.apiVersion" . }}
+kind: ClusterRole
+metadata:
+  name: "crd-reader-{{ .Release.Name }}"
+  labels:
+    {{- if .Values.global.labels }}
+    {{- toYaml .Values.global.labels | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.global.annotations }}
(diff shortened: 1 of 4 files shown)


buarki (dismissed):

guillermotestkube:
@greptile

PR #8033, opened 2026-07-29
fix: start cluster-inventory CRD watcher only on listener-capable agents

dejanzele:
## How

The cluster-inventory CRD watcher was gated on `proContext.APIKey != ""`, so every connected agent started it, including runner-only ones that have no CRD RBAC and whose pushes the Control Plane rejects anyway (`GetMetadataFromContext(ctx, api.AgentCapabilityListener)` on `PutClusterResources`).

The agent already knows whether it is a listener. On every startup it sends its flag-derived capabilities through `UpdateAgentCapabilitiesOnStartup`, the Control Plane stores that set (it only keeps the runner capability under its own control), and the response returns the effective set. The listener capability comes from `DISABLE_TEST_TRIGGERS`, which the runner chart derives from `listener.enabled`.

This keeps that effective set on `ProContextAgent.Capabilities`, enum-typed, and gates the watcher on it through `ShouldPushClusterInventory`:

- no API key, so standalone with an embedded control plane: never push, discovery is served from the agent's own API
- otherwise push only when the capabilities contain `listener`

When the startup update endpoint is unavailable or fails, the agent falls back to the list it built from its own flags, which is the same list it would have sent. No wire change and no Control Plane change are needed.

## Validated

On the local k3d stack with the agent built from this branch and the control plane from main, flipping the same agent deployment through `DISABLE_TEST_TRIGGERS`. The deployment kept `ENABLE_K8S_CONTROLLERS=true` and `GITOPS_KUBERNETES_TO_CLOUD_ENABLED=true` throughout, so only the listener capability decided.

| | listener (`DISABLE_TEST_TRIGGERS=false`) | runner-only (`DISABLE_TEST_TRIGGERS=true`) |
| --- | --- | --- |
| capabilities from CP | `runner, listener, gitops, webhooks` | `runner, gitops, webhooks` |
| inventory pushes at startup | 2 (`count: 28`) | 0 |
| agent | connected | connected |

Restoring the flag brought the pushes back on the same build. The fallback when the startup call fails is the one-line assignment before the call and is covered by reading, not by a live run.


dejanzele:
@greptileai

courageousillumination (approved):

PR #8034, opened 2026-07-29
fix(testkube-runner): grant CRD read to listener-capable agents

dejanzele:
## How

The `testkube-runner` chart never granted `apiextensions.k8s.io/customresourcedefinitions`, so agents deployed from it log a recurring `Failed to watch ... CustomResourceDefinition ... forbidden` and retry forever. The sibling `testkube-api` chart has had `crd-reader-rbac.yaml` for this; the runner chart did not.

Adds the equivalent ClusterRole and ClusterRoleBinding, scoped to the same `listener.enabled or gitops.enabled` condition that already gates `watchers-role`, and to `pod.serviceAccount.autoCreate` like the rest of the chart's RBAC.

Scoped rather than unconditional because a runner-only agent has no reason to hold cluster-wide read on every CRD in the cluster. It never starts the CRD informer once #8033 lands, and the Control Plane refuses its inventory push regardless. The two conditions cover both worlds: against a capability-reporting control plane `listener.enabled` is what earns the agent the listener capability, and against an older one `gitops.enabled` is what trips the agent's local fallback.

This supersedes #8023, which grants the same access to every runner unconditionally.

A `crd_reader_rbac_test.yaml` suite covers the gate in both directions, the rule contents, and the binding subject. Chart suite goes from 8 files / 75 tests to 9 / 83.

## Note for review

The ClusterRole is named `crd-reader-{{ .Release.Name }}`, matching the one in `testkube-api`. Two releases sharing a name would collide on this cluster-scoped object. The existing `watchers-role-{{ .Release.Name }}` has the same exposure, so this follows the chart family's convention rather than diverging, but it is worth deciding deliberately.

### DIFF
--- a/k8s/helm/testkube-runner/templates/crd-reader-rbac.yaml
+++ b/k8s/helm/testkube-runner/templates/crd-reader-rbac.yaml
@@ -1,0 +1,60 @@
+{{- /*
+Cluster-scoped CRD read for the agent's cluster-resources inventory
+(internal/inventory), which backs the TestTrigger resourceRef picker. CRDs are
+cluster-scoped, so a namespaced Role cannot grant it.
+
+Gated on the same listener/gitops condition as watchers-role: a runner-only
+agent never starts the CRD informer, so it must not hold cluster-wide read on
+every CRD.
+*/}}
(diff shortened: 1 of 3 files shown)


dejanzele:
@greptileai

PR #8039, opened 2026-07-30
[WIP] feat: expose kubernetes client and trigger watcher metrics

dejanzele:
## How

Wraps the transport of every `*rest.Config` the agent builds, so each Kubernetes API response is counted by code and HTTP method and each request duration is observed. `GetK8sClientConfig` instruments the configs it returns, which covers the clientset, the dynamic client, the apiextensions client and the controller-runtime manager. `InstrumentConfig` is exported for the call sites that build a config elsewhere, such as the CRD client that takes one from controller-runtime.

The trigger service gets two lifecycle metrics: a counter for informer fleet starts and a gauge for whether this instance holds the trigger lease. A relist storm shows up as fleet starts correlating with throttled responses.

Metric names:

- `testkube_k8s_client_requests_total{code,method}`
- `testkube_k8s_client_request_duration_seconds{method}`
- `testkube_triggers_watcher_starts_total`
- `testkube_triggers_leader`

Label values are bounded to response codes and HTTP methods. No namespace, object name, request path or host, so the series count cannot grow with cluster contents.

Not every binary is covered. A few call sites build a config from `rest.InClusterConfig` directly, mostly in the toolkit and CLI paths, and are left alone here since they are short lived and serve no metrics endpoint.

## Notes for review

`k8s.io/client-go/tools/metrics.Register` is deliberately not used. It accepts one set of adapters per process behind an internal `sync.Once`, and controller-runtime claims it from a package init before any of our code runs, so a second registration is silently dropped. An earlier version of this change did register there and recorded nothing at all in the agent binary while passing its unit tests. The comment in `metrics.go` says so, to keep someone from switching it back.

Request latency is measured until the response headers arrive. Watch requests return at that point and stream their body afterwards, so a long lived watch contributes one short observation rather than a multi day one.

Client side rate limiter latency is not reported. Measuring it means setting `rest.Config.RateLimiter`, and `rest` only builds a per client limiter when that field is nil, so owning it would replace four independent token buckets with one shared bucket and change real throttling behavior. That is worth doing deliberately, not as a side effect of adding a metric.

Whether the runner and agent charts add scrape annotations for port 8088 is a deployment concern for the charts repository.


dejanzele:
@greptileai

courageousillumination:
Converting to draft while this is WIP

PR #8040, opened 2026-07-30
[WIP] fix: tolerate transient lease check failures in the leader coordinator

dejanzele:
## How

The leader coordinator now separates "another instance holds the lease" from "the lease backend did not answer". A successful check that reports another holder still releases leadership immediately. A backend error while we are the leader keeps leadership until the lease duration has elapsed since the last successful renewal, and only then releases. When we are not the leader an error changes nothing.

The tolerance window is the lease duration itself rather than a new knob, since a Kubernetes Lease is valid until its renew time plus its duration. Holding through errors inside that window therefore never creates leadership the Lease semantics did not already permit. `WithLeaseDuration` defaults to `leasebackend.DefaultMaxLeaseDuration`.

The window is measured from before the lease call is issued, not from when it answered. A throttled backend can take seconds to reply, and measuring from the reply would keep leadership past the point where other instances already consider the lease expired.

The Kubernetes lease backend also retries once on a conflict while renewing a lease it already holds. It re-reads the lease, confirms we are still the recorded holder, and retries the update. A conflict on our own lease is a stale read, not a demotion, and previously reported the lease as lost.

## Notes for review

Two coordinators are affected. The trigger service coordinator owns all trigger informers and the git informer task. The coordinator in `cmd/api-server/main.go` owns the control plane connection, the agent event reader, the telemetry heartbeat and the cron scheduler. Before this change a single transient lease error tore all of those down and rebuilt them on the next tick, so the fix removes a spurious agent reconnect as well as the informer relist storm.

The flip side is that the tolerance window now also applies to the cron scheduler and the control plane connection. The exposure stays bounded by the lease TTL, which is the guarantee the Lease already provides, and it is why the renewal timestamp is taken before the call rather than after. Deployments running a single replica are unaffected either way.

The coordinator's lease duration and the duration the backend writes into the Lease object are two separately configured values that agree today by sharing a default. Making the backend report its own duration through the repository interface would remove that coupling, but it touches all three backends and is left out of this change.

Existing coordinator tests encoded the old behavior, where any error released leadership. They were updated deliberately.


dejanzele:
@greptileai

courageousillumination:
Converting to draft while this is WIP

PR #8041, opened 2026-07-30
perf: cache webhook secret resolution with a TTL cache

dejanzele:
## How

Adds a caching decorator in `pkg/secret` and wraps the one client the webhook loader builds. Webhook parameters are resolved on every notification, so an event burst previously turned into one uncached secret GET per event per parameter.

Only `Get` and `GetObject` are cached. Every mutating method passes through to the inner client and invalidates the affected key. Entries are keyed by name plus the namespace the call resolves to, so a read with an explicit namespace and a read that falls back to the default share one entry. Returned maps and objects are copies, so a caller cannot mutate what the cache holds.

`TESTKUBE_SECRET_CACHE_TTL` sets the lifetime and defaults to 30s. A zero value returns the inner client unchanged, which disables caching completely.

Missing secrets are cached too, for a third of the TTL, so a webhook pointing at a secret that does not exist does not hammer the API server either. Every other error is treated as possibly transient and is never cached.

A read that started before an invalidation does not repopulate the cache afterwards, so a mutation cannot be overwritten by an older in flight result.

The saving is larger than one read per event. `processTemplate` rebuilds the resolved config map for every templated field, so an uncached notification reads the same secret once for the uri, once for the payload and once more for every header. A webhook with a templated uri and a payload resolves it twice per event before this change and once per TTL window after it.

Storage is the shared `pkg/cache`, rather than another map with an expiry sweep next to it. That package needed two things it did not have: `Delete`, without which a mutation cannot invalidate what it replaced, and a way to substitute the time source, without which TTL behavior can only be tested by waiting. Both are added here, and `Clear` comes with `Delete` for `DeleteAll`, which changes secrets it cannot map back to individual keys.

What stays in `pkg/secret` is the part that is specific to reading secrets: a shorter lifetime for not found results, coalescing of concurrent misses, and the revision that lets a mutation discard reads it superseded.

## Notes for review

This trades immediacy for request volume, and the staleness bound is worth an explicit decision: rotating a secret takes up to the TTL to reach webhook parameters, and creating a secret that was just looked up and missing takes up to a third of the TTL to become visible. Setting the TTL to zero restores the current behavior for anyone who needs rotation to apply immediately.

This branch includes the `Get` clock fix from #8053, since injecting a time source is pointless while `Get` reads the wall clock. Merge that first and this diff shrinks accordingly.

Widening `cache.Cache` touches every implementer. In this repository that is `InMemoryCache` and one test fake in `pkg/imageinspector`, both updated here.

The decorator is applied at exactly one construction site. Other `pkg/secret` callers are unchanged, and adopting it elsewhere is now a one line change per site.

Concurrent misses for the same key are coalesced with `singleflight`, so a burst arriving on a cold key produces one read rather than one per listener. Without that the cache only helps after the first read completes, which is not the case the change exists for.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,7 @@
 - Agent behavior is driven by env vars defined in `internal/config/config.go` (scan for `envconfig:"..."` tags when researching a toggle).
 - Git trigger informer behavior is tuned via `TEST_TRIGGER_GIT_INFORMER_RECONCILE_INTERVAL`, `TEST_TRIGGER_GIT_INFORMER_REPO_DEPTH`, `TEST_TRIGGER_GIT_INFORMER_LIST_TIMEOUT`, `TEST_TRIGGER_GIT_INFORMER_MAX_COMMITS_SCAN`, `TEST_TRIGGER_GIT_INFORMER_PULL_RETRIES`, and `TEST_TRIGGER_GIT_INFORMER_PULL_RETRY_DELAY`.
 - Git trigger informer execution is leader-gated in `cmd/api-server/main.go` through the shared `leader` coordinator tasks, so only the active leader performs periodic git pulls/reconciliation.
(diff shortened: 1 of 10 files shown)


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

courageousillumination (approved):

PR #8042, opened 2026-07-30
perf: strip managed fields from trigger informer caches

dejanzele:
## How

Every informer the trigger service starts now gets a transform that clears `metadata.managedFields` before an object enters its cache. Managed fields are a large share of a serialized object, so caching them costs memory that scales with the number of watched namespaces and the number of objects in them.

The transform is installed by a small helper that registers the event handler at the same time, because `SetTransform` has to be called before the informer starts or client-go rejects it. The helper is used at every handler registration in the trigger service, including the dynamic informers for workflow triggers.

Values without object metadata, such as deletion tombstones, pass through untouched. Returning an error there would drop the event entirely.

`TEST_TRIGGER_KEEP_MANAGED_FIELDS=true` keeps the fields. The informer caches are what v2 trigger expressions resolve `resource` against, so a deployment whose expressions read `metadata.managedFields` has a way to opt out of the memory saving rather than losing data. The default drops them, since managed fields are server side apply bookkeeping and reading them from a trigger expression is not an expected use.

## Notes for review

This reduces memory only. It does not change how many requests or watch streams the agent opens.

The behavior worth checking is the expression surface rather than the transform itself. `buildEventExpressionMachine` in `pkg/triggers/executor.go` registers the whole watched object as `resource`, and the previous one as `oldResource`, so an expression such as `{{ resource.metadata.managedFields[0].manager }}` resolves against the cached object. Nothing in the trigger service reads the field, but a user expression can, which is what the toggle exists for.

Tests cover both directions: with the default, an object observed through a started informer has no managed fields in the cache; with the toggle on, the entries it was created with are still there. Both assertions fail if the transform or the toggle is bypassed.


dejanzele:
@greptileai

courageousillumination:
This PR has a number of unrelated changes (related to pretty output). I'm moving back to draft, please re-open once these have been removed (or if they are intentional, please describe why).

PR #8053, opened 2026-07-31
fix: honor the injected clock when checking in-memory cache expiry

dejanzele:
## How

`InMemoryCache` carries a `timeGetter` so its time source can be substituted, and `Set` uses it when computing an item's expiry. `Get` did not: it compared that expiry against `time.Now()` directly, so the injected clock decided when an item expires but not when it is observed as expired. `Get` now uses the same source.

Production behavior is unchanged, since the only time source ever installed is `time.Now`. What changes is that TTL behavior becomes testable without waiting on the wall clock.

The round trip test is rewritten on top of that. It previously stored an item with a 100ms TTL and polled for up to 300ms for it to disappear, which is both slow and timing dependent. It now advances an injected clock instead, and covers four cases declaratively: an item with no TTL is not cached, an item is served inside its TTL, it is gone once the clock passes it, and an infinite TTL survives an advance of a century. The suite no longer contains a sleep or a poll.

### DIFF
--- a/pkg/cache/inmem.go
+++ b/pkg/cache/inmem.go
@@ -40,7 +40,7 @@ func (c *InMemoryCache[T]) Get(ctx context.Context, key string) (T, error) {
 		return defaultVal, errors.New("unexpected item type found in cache")
 	}
 
-	if i.expiresAt != nil && i.expiresAt.Before(time.Now()) {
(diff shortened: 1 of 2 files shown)


dejanzele:
@greptileai

dejanzele:
@greptileai

courageousillumination (approved):

PR #8056, opened 2026-07-31
fix(expressions): return a character, not a byte, from at() on a string

### LINKED ISSUE #8055: at() returns a byte value instead of a character when indexing a string
**Describe the bug**

The `at()` standard library function in the Test Workflow expression language returns a raw byte value instead of a character when the first argument is a string. The list branch of the same function behaves correctly, so the two disagree.

In `pkg/expressions/stdlib.go` the string branch is:

```go
if k >= 0 && k < int64(len(v)) {
    return NewValue(v[int(k)]), nil
}
```

Indexing a Go string yields a `byte` (`uint8`), not a one character string, so the value is serialized as a number. As a second effect, `static.Type()` has no case for `uint8`, so the result reports `TypeUnknown` and is treated as neither a string nor an int by the rest of the expression engine.

**To Reproduce**

```go
package expressions

import (
	"fmt"
	"testing"
)

func TestAtOnString(t *testing.T) {
	for _, e := range []string{`at("abc", 0)`, `at("abc", 1)`, `at("hello", 4)`, `at("abc", 1) == "b"`, `at(["a","b"], 1)`} {
		fmt.Println(e, "=>", MustCompile(e).String())
	}
}
```

**Actual behavior**

```
at("abc", 0)         => 97
at("abc", 1)         => 98
at("hello", 4)       => 111
at("abc", 1) == "b"  => false
at(["a","b"], 1)     => "b"     <- list branch is correct
```

**Expected behavior**

```
at("abc", 0)         => "a"
at("abc", 1)         => "b"
at("hello", 4)       => "o"
at("abc", 1) == "b"  => true
at(["a","b"], 1)     => "b"
```

**Impact**

Any Test Workflow that indexes into a string with `at()` silently receives a number. Comparisons against a string literal always evaluate false, and concatenation splices in the numeric code point rather than the character. Nothing errors, so the failure is silent and shows up only as a workflow behaving unexpectedly.

**Version / Cluster**

- Reproduced on `main` at c2c9fa3. Not environment specific, the expression evaluator is pure Go and the repro above is a unit test.

**Additional context**

Suggested fix is to slice rather than index, `v[k : k+1]`. That returns a proper one character string and keeps `at()` index compatible with `len()`, which is also byte based, so `at(s, len(s)-1)` stays correct and `at(s, 0) + at(s, 1) + ...` reassembles the original string.

Happy to open a PR for this.

UditDewan:
## Pull request description

Fixes #8055.

`at()` returned a raw byte instead of a character when the first argument was a string. The list branch of the same function was already correct, so the two disagreed.

```go
if k >= 0 && k < int64(len(v)) {
    return NewValue(v[int(k)]), nil   // v is a string, so v[k] is a byte
}
```

Indexing a Go string yields a `byte` (`uint8`), which serializes as a number. Secondary effect: `static.Type()` has no case for `uint8`, so the result reported `TypeUnknown` and was treated as neither a string nor an int by the rest of the expression engine.

Nothing errored, so any Test Workflow indexing into a string with `at()` silently got a number, with comparisons against string literals always evaluating false.

Before:

```
at("abc", 0)         => 97
at("abc", 1)         => 98
at("abc", 1) == "b"  => false
at("héllo", 1)       => 233
at(["a","b"], 1)     => "b"     <- list branch already correct
```

After:

```
at("abc", 0)         => "a"
at("abc", 1)         => "b"
at("abc", 1) == "b"  => true
at("héllo", 1)       => "é"
at(["a","b"], 1)     => "b"
```

`at()` indexes by rune, so multibyte characters come back whole and every result is valid UTF-8. Out of bounds reports the rune count.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] added a test

## Breaking changes

- None for the reported defect. The previous result was a byte with `TypeUnknown`, which had no coherent meaning in the expression language.

## Changes

- `pkg/expressions/stdlib.go`: index by rune in the string branch of `at()`, and report rune count in the out of bounds error.
- `pkg/expressions/parse_test.go`: assertions in `TestCompileStandardLib` covering ASCII indexing, equality against a string literal, and multibyte characters.

## Fixes

- #8055

---

### Note for maintainers: `at()` and `len()` now disagree on non-ASCII

`len()` remains byte based, so `at(s, len(s) - 1)` no longer reliably returns the last character:

```
len("héllo")     => 6   (bytes)
at("héllo", 5)   => out of bounds (length=5 runes)
```

This fails loudly rather than silently returning a corrupted character, which seems the better of the two failure modes. I deliberately left `len()` alone, since making it rune based would change behavior for anyone relying on byte length and belongs in its own change rather than folded into a bug fix. Happy to open a follow up for it, or fold it in here if you would prefer both together.

### Testing notes

`go build ./pkg/expressions/...` passes and the new assertions pass. Full disclosure on the test run: `TestCompileStandardLib` and `pkg/expressions/libs` already fail on my machine before this change, because I am on Windows and the `relpath`/`abspath`/`makepath` assertions plus `fs_test.go` assume POSIX paths. I verified my change is not the cause by comparing failing line numbers before and after, they are the identical set shifted by the lines I added. CI on Linux should be clean, but worth a look if it is not.

### Review history

The first commit sliced a single byte (`v[k:k+1]`) rather than indexing by rune. That was wrong for multibyte input and is superseded by 5860ae4, see [this comment](https://github.com/kubeshop/testkube/pull/8056#issuecomment-5145967674) for why.


UditDewan:
Good catch, and you are right. Fixed in 5860ae4.

`at()` now indexes by rune:

```go
runes := []rune(v)
if k >= 0 && k < int64(len(runes)) {
    return NewValue(string(runes[int(k)])), nil
}
return nil, fmt.Errorf(`"at" function: error: out of bounds (length=%d, index=%d)`, len(runes), k)
```

My original argument for byte slicing does not survive scrutiny. I justified it as keeping `at()` consistent with the byte based `len()`, but that consistency only holds for ASCII, where byte and rune indexing are identical anyway. For non-ASCII the byte version produced invalid UTF-8, which `json.Marshal` renders as replacement characters. So rune indexing is not a tradeoff against the byte version, it strictly dominates it.

Verified:

```
at("héllo", 1)         => "é"
at("日本語", 0)          => "日"
at("日本語", 2)          => "語"
at("日本語", 1) == "本"   => true
at("日本語", 3)          => out of bounds (length=3, index=3)
```

All outputs pass `utf8.ValidString`. Added the multibyte cases to `TestCompileStandardLib` alongside the ASCII ones.

### One consequence worth a maintainer decision

`len()` is still byte based, so `at()` and `len()` now disagree on non-ASCII strings. `at(s, len(s) - 1)` no longer reliably gets the last character:

```
len("héllo")            => 6   (bytes)
at("héllo", 5)          => out of bounds (length=5 runes)
```

That fails loudly with an out of bounds error rather than silently returning a wrong or corrupted character, which I think is the better failure mode of the two. I deliberately did not touch `len()` here, since making it rune based would change behavior for anyone currently relying on byte length and that belongs in its own change rather than folded into a bug fix. Happy to open a follow up issue for it, or to include it here if you would rather land both together.

I also dropped the `at("abc", len("abc") - 1)` assertion I had in the first commit. It passes, but it encodes a pattern that is only safe for ASCII and I did not want the test suite implying otherwise.

vsukhin:
@greptile

vsukhin:
thank you, @UditDewan great catch again!

vsukhin (approved):

PR #8060, opened 2026-07-31
feat(git): add exported matchers package for cross-repo reuse

buarki:
## Changes

Adds a new package `pkg/git/matchers/` that exports pure predicate helpers for evaluating whether a git event should trigger a workflow: path globbing, ref-to-branch extraction, PR base-branch matching, PR action matching, PR path filtering, and PR action inference from state transitions.

These are the same predicates that already exist unexported inside `pkg/git/informer/{informer,github_pr}.go` (used by the git-triggers polling loop). The new package is a **copy**, not a move — the informer keeps its existing unexported helpers unchanged, so nothing about how git-triggers behaves changes with this PR.

## Why now

The Testkube Cloud Quality Loop is adding a webhook-driven path (TKC-6384) that evaluates the same matcher semantics customers already know from OSS `TestTrigger`. Rather than reinvent globbing/PR-matching rules, it should consume the same predicates. Making them exportable from OSS is the enabling step.

Today, Cloud has a **temporary internal copy** of these predicates (`internal/qualityloop/gitmatchers/` in the closed-source repo) with a TODO to replace with an import of this package once it lands.

## Approach

- **Additive only**: new files under `pkg/git/matchers/`, no changes to `informer.go` or `github_pr.go`
- **Decoupled signatures**: the exported functions take primitive slices and strings (no `testkube.*` types) so callers outside the informer can consume them without dragging in provider- or control-plane-specific dependencies
- **Table-driven tests** cover each predicate independently; all existing OSS tests continue to pass unchanged

### DIFF
--- a/pkg/git/matchers/matchers.go
+++ b/pkg/git/matchers/matchers.go
@@ -1,0 +1,232 @@
+// Package matchers exports pure predicate helpers for evaluating whether a git
+// event should trigger a workflow. Callers pass primitive slices and strings
+// so the package stays free of provider-specific and control-plane-specific
+// types, making it consumable by both the OSS TestTrigger informer and by
+// external tools that need to reason about the same rules.
+package matchers
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// NormalizePaths trims whitespace and surrounding slashes from each path and
+// drops empty entries. Returns nil if the input is empty or entirely blank.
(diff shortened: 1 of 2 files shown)


buarki:
Superseded by a proper move-and-refactor PR that eliminates duplication rather than adding another copy. Will open the replacement momentarily.

PR #8061, opened 2026-07-31
feat(git): extract shared matchers package from informer

buarki:
## Changes

Extracts the pure predicate helpers that today live unexported inside `pkg/git/informer/{informer,github_pr}.go` (used by the git-triggers polling loop) into a new exported package `pkg/git/matchers/`.

### Predicates moved (10 with identical behavior, only renamed to exported)

| Before | After |
|---|---|
| `normalizePaths` | `matchers.NormalizePaths` |
| `pathMatches` | `matchers.PathMatches` |
| `pathMatchesNormalized` | `matchers.PathMatchesNormalized` |
| `pathIsIgnored` | `matchers.PathIsIgnored` |
| `matchGlob` | `matchers.MatchGlob` |
| `nameMatchesPatterns` | `matchers.NameMatchesPatterns` |
| `nameMatchesAny` | `matchers.NameMatchesAny` |
| `branchFromRef` | `matchers.BranchFromRef` |
| `tagFromRef` | `matchers.TagFromRef` |
| `prPathsMatch` | `matchers.PRPathsMatch` |

### Predicates moved with adapted signatures (3)

These previously took `*testkube.TestTriggerContentGitPullRequest` (or `githubPR`). To keep the new package free of provider-specific types (so it is usable outside the informer), the signatures now take primitive slices/strings. Behavior is byte-for-byte identical.

| Before | After | Call-site change |
|---|---|---|
| `prMatchesBaseBranch(base, *TestTriggerContentGitPullRequest)` | `matchers.PRMatchesBaseBranch(base, branches, branchesIgnore []string)` | Unpack `prConfig.Branches` / `prConfig.BranchesIgnore` at call site, keep nil-guard on `prConfig` for the "nil = match all" semantic |
| `prMatchesTypes(action, *TestTriggerContentGitPullRequest)` | `matchers.PRMatchesTypes(action, types []string)` | Unpack `prConfig.Types` with nil-guard |
| `determinePRAction(prev, curr, githubPR)` | `matchers.DeterminePRAction(prev, curr, headSHA string)` | Pass `pr.Head.SHA` |

### Migration

- Callers inside `pkg/git/informer/` are updated to import `pkg/git/matchers` and use the exported names.
- The unexported originals are removed from `informer.go` and `github_pr.go`.
- Test coverage is preserved: pure-predicate tests move to `pkg/git/matchers/matchers_test.go`; integration tests in `pkg/git/informer/*_test.go` keep their end-to-end coverage but now assert against the new exported symbols.

## Why

The Testkube Cloud Quality Loop is adding a webhook-driven path that needs to evaluate the same matcher semantics customers already know from OSS `TestTrigger`. Rather than reinvent globbing/branch/PR-matching rules (and risk divergence), it should consume the same predicates. Making them exportable is the enabling step.

Consequences:
- Cloud will import `pkg/git/matchers` directly, eliminating a temporary internal copy it carries today.
- Fixes/tweaks to matcher semantics land in one place and reach both OSS and Cloud automatically.
- New public API surface. Signatures are minimal and generic; future evolution follows the usual compatibility rules.

## Risk

- **Zero behavior change in git-triggers**: predicates are moved, not modified. All callers updated in the same PR.
- **Nil-guard semantics preserved**: the three signature-refactored functions had implicit nil-config handling (`prConfig == nil` → match all). The call sites now hold that guard explicitly, and the removed function bodies were identical to the new exported implementations.
- **Test coverage carried across**: `matchers_test.go` covers each predicate directly; `informer_test.go` and `github_pr_test.go` keep the integration paths.

## Notes for reviewer

- Pre-existing unrelated test failures (`commit.gpgSign` env config) show up on my machine on `main` too and are not caused by this PR.

## Checklist

- [x] tested locally (`go test ./pkg/git/matchers/... ./pkg/git/informer/... && go build ./...`)
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar (approved):
Nice abstraction

PR #8078, opened 2026-08-06
fix: accept large numeric resource values in TestWorkflow CRDs

dejanzele:
Stacked on #8076; merge that PR first, then retarget this one to `main`.

An unquoted numeric resource value above the int32 range, for example `memory: 8589934592`, breaks the TestWorkflow informer list the same way as the config values in #8076. This PR migrates `resources.limits` and `resources.requests` to the `ConfigValue` type from #8076.

How:
- `Resources.Limits` and `Resources.Requests` change from `map[corev1.ResourceName]intstr.IntOrString` to `map[corev1.ResourceName]ConfigValue`.
- The mappers convert with plain string casts instead of int sniffing.
- The generated CRD schema is unchanged; only the deepcopy file regenerates.

The last int-or-string fields, `count` and `maxCount`, migrate in the next PR of the stack.

### DIFF
--- a/api/testworkflows/v1/config_value_test.go
+++ b/api/testworkflows/v1/config_value_test.go
@@ -95,6 +95,23 @@ func TestConfigValue_UnmarshalYAML(t *testing.T) {
 	}
 }
 
+func TestResources_UnmarshalLargeNumericMemory(t *testing.T) {
+	containerJSON := []byte(`{
+		"resources": {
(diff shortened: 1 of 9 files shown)


dejanzele:
@greptileai review

courageousillumination (approved):

vsukhin (approved):

PR #8079, opened 2026-08-06
fix: accept large numeric count values in TestWorkflow CRDs

dejanzele:
Stacked on the resources PR of this series; merge that one first, then retarget this one.

This PR migrates the last two int-or-string fields of the TestWorkflow CRDs, `count` and `maxCount`, to the `ConfigValue` type from #8076. After this PR, no CRD field decodes numbers into int32, so no stored value can break the informer list.

How:
- `StepExecuteStrategy.Count` and `MaxCount` change from `*intstr.IntOrString` to `*ConfigValue`.
- `ReadCount` and `GetParamsSpec` in the toolkit take `ConfigValue` and keep the same expression resolution.
- The mappers use the string-based helpers from #8076.
- The generated CRD schema is unchanged; only the deepcopy file regenerates.


dejanzele:
@greptileai review

courageousillumination (approved):

vsukhin (approved):

PR #8085, opened 2026-08-10
feat: Tag Testkube-provisioned hosted runners in agent telemetry

olensmar:
## Pull request description

Hosted trial runners (the ones Testkube provisions in vClusters so users can run tests without deploying an agent) emit the exact same `testkube_api_start` and `testkube_api_heartbeat` payloads as runners users deploy in their own infrastructure. There is currently no way to tell the two apart in analytics, so trial infrastructure inflates every metric that counts "real" deployments.

The control plane does mark these runners — with a `runner.testkube.io/cloudrunner` deployment label and a `tkcagent_hr_`-prefixed agent name — but both markers only travel over gRPC at registration time and never reach the telemetry payload built by `NewAPIPayload`. None of the existing params discriminate either: agent IDs are always `tkcagent_*` now so `persona:default` covers both, `cluster_type` resolves to the hosting provider a customer could also be on, `host` is anonymized, and `cluster_id` is a fresh per-vCluster UUID.

This derives a `hosted-runner` capability tag from `RUNNER_NAME`, which the control plane already sets to the `tkcagent_hr_` prefixed agent name via the `testkube-runner` chart's `runner.name` value. No Helm chart change and no control plane change are needed.

Note for whoever uses this downstream: the tag only appears once hosted runners are running an agent image built from this change, so filtering on absence of the tag will misclassify hosted runners until that rollout completes.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

## Breaking changes

- None.

## Changes

- `AgentCapabilities` appends a `hosted-runner` tag when `cfg.RunnerName` carries the `tkcagent_hr_` prefix assigned by the control plane, with a named constant documenting that it must stay in sync with `naming.HostedRunnerAgentName` in `testkube-cloud-api`.
- Added `TestAgentCapabilities_HostedRunner`, covering a hosted name, a user-chosen name, an empty name, and a plain `tkcagent_` id used as a name (the near-miss that would break if the prefix were loosened).
- Documented the tag in `AGENTS.md` and `ARCHITECTURE.md`.

## Fixes

- Hosted trial runners were indistinguishable from user-deployed runners in agent telemetry.


courageousillumination (approved):
It feels a bit weird to represent this as a capability, but if we don't have another clear channel, I suppose this works.

olensmar:
thanks @courageousillumination - it seemed to have been the lowest-touch option 🙏

PR #8091, opened 2026-08-11
fix(webhooks): skip silent-webhook executions in previous-state lookup

buarki:
Ref: [TKC-6495](https://linear.app/kubeshop/issue/TKC-6495/become-testworkflow-up-is-triggered-after-failed-silent-execuiton)

## Problem

`GetPreviousFinishedState` did not filter out executions with `SilentMode.Webhooks=true`, so a silent failed run in the workflow's history was being treated as the "previous state" for the next non-silent run. The next non-silent passed execution then triggered a spurious `become-testworkflow-up` webhook

Confirmed on the .dev postgres for the `cloud-ui-e2e-12874` alert on 2026-08-11: the alerted execution was preceded by a fully-silent failed run 42 seconds earlier; without that silent row, the query would have returned `passed` and no become-up would have fired

## Fix

Add a silent-mode filter to `GetPreviousFinishedState` in both backends, targeting the `webhooks` aspect specifically. This mirrors what the webhook listener already checks in `pkg/event/kind/webhook/listener.go` when deciding whether to dispatch a webhook for the current execution

## Test plan

- [x] Postgres integration suite passes with fix, `GetPreviousFinishedStateSkipsSilentWebhookExecutions` fails without fix
- [x] Mongo integration suite passes with fix, same test fails without fix
- [x] Full `pkg/repository/testworkflow/{postgres,mongo}` suites green
- [x] `go build ./...` and `make generate-sqlc` clean

### DIFF
--- a/pkg/cloud/data/testworkflow/execution.go
+++ b/pkg/cloud/data/testworkflow/execution.go
@@ -181,8 +181,12 @@ func (r *CloudRepository) GetTestWorkflowMetrics(ctx context.Context, name strin
 }
 
 // GetPreviousFinishedState gets previous finished execution state by test
-func (r *CloudRepository) GetPreviousFinishedState(ctx context.Context, workflowName string, date time.Time) (testkube.TestWorkflowStatus, error) {
-	req := ExecutionGetPreviousFinishedStateRequest{WorkflowName: workflowName, Date: date}
+func (r *CloudRepository) GetPreviousFinishedState(ctx context.Context, workflowName string, date time.Time, opts testworkflow2.GetPreviousFinishedStateOptions) (testkube.TestWorkflowStatus, error) {
+	req := ExecutionGetPreviousFinishedStateRequest{
+		WorkflowName:                workflowName,
+		Date:                        date,
+		SkipSilentWebhookExecutions: opts.SkipSilentWebhookExecutions,
+	}
 	response, err := r.executor.Execute(ctx, CmdTestWorkflowExecutionGetPreviousFinishedState, req)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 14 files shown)


buarki:
@greptile

buarki:
@greptile-apps your concern here does not hold up when you walk through the actual scenario for the AI trigger consumer.

Pre-fix, the AI trigger handler suffers from the exact same contamination this PR is fixing for the webhook listener. Concrete case:

1. A silent-webhooks execution finishes with `failed` and lands in the history.
2. A non-silent execution finishes with `passed`.
3. AI trigger handler calls `GetPreviousFinishedState`, sees `prev=failed, curr=passed`, decides "state changed", spawns a Copilot session that no user asked for. That is a spurious session, driven by history the user explicitly marked as "do not react to me".

Post-fix, that same scenario resolves to `prev=passed` (the previous non-silent), `curr=passed`, no transition, no spurious session. The AI consumer is helped in the same direction the webhook listener is, not hurt.

There is no evidence anywhere in the codebase or product intent that the AI trigger handler wants to react to silent-webhooks executions. Silent means "do not react to me", and spawning a Copilot session is a reaction. The two are semantically aligned, so applying the same filter to both consumers is intentional, not a leaked webhook detail.

buarki:
@greptile would you update the score based on our discussion?

courageousillumination (changes requested):
I don't think this is right. If the _previous_ execution is excluded from webhooks shouldn't determine if the current execution should look at it for calculations. I think silentMode.health is the one that would map more cleanly to this model. But I'm also not sure about doing this unconditionally at the query level; I don't think we have other callers, but they may have different constraints about how they want to handle the slient mode.

courageousillumination (commented):
Ok, after talking with Tomasz, we can keep this as webhooks. But to greptile's point this method becomes unclear for future use. Either rename it or make this an option to the query.

buarki:
@greptile

buarki:
@greptile

courageousillumination (approved):

PR #8108, opened 2026-08-14
feat(proto): add GITINTEGRATION actor type for git-provider-triggered flows

buarki:
## Summary

Adds a new `QUALITYLOOP` value on the proto side (kept internal to the testkube team, only visible on the gRPC wire and in Go source) and a matching `gitintegration` value on the OpenAPI actor type enum, which is the customer-facing name. The mapper cross-translates between the two so nothing customer-visible ever exposes `QUALITYLOOP`.

Downstream consumers pick up the new actor:
- Telemetry bucketer routes it into a dedicated `git-integration` bucket instead of falling through to `unknown`.
- CDEvents mapper places it under the `event` trigger type alongside `testtrigger` and `testworkflow`.

## Why

Cloud-api's Quality Loop pipeline emits executions for pull-request test runs and today marks them as `CICD` on the interface (actor stays `USER`). Users cannot filter QL executions specifically on the Executions page.

This enum kicks off the change on the cloud-api side to route QL executions through a dedicated actor type. The customer-facing name stays provider-agnostic so GitHub, GitLab, and Bitbucket land under the same actor without another proto bump; provider identity lives in `RunningContext.Name` (e.g. `git-provider-github`).

## Scope

This PR is scoped to naming: proto enum, OpenAPI enum, CRD kubebuilder enum, CLI validator, telemetry bucketer, CDEvents mapper. The platform-level `WithActorExecutionIds` filter needed by cloud-api to surface Quality Loop children under the new filter chip is a separate stacked PR: https://github.com/kubeshop/testkube/pull/8111.

## What customers will see (CLI, REST, CRD)

The customer-facing string is `gitintegration` everywhere. `QUALITYLOOP` is only visible in proto/gRPC (Testkube internal) and in the Go const symbols.

CLI usage:

` + '```bash' + `
# List executions triggered by the Quality Loop pipeline
kubectl testkube get twe --actor-type gitintegration
` + '```' + `

REST API:

` + '```bash' + `
GET /organizations/{orgId}/environments/{envId}/executions?actorType=gitintegration
` + '```' + `

CRD:

` + '```yaml' + `
apiVersion: testworkflows.testkube.io/v1
kind: TestWorkflowExecution
spec:
  runningContext:
    actor:
      type: gitintegration
      name: git-provider-github
` + '```' + `

## Test plan

- [x] ` + '`go test ./pkg/testworkflows/testworkflowexecutor/... ./pkg/event/kind/testworkflowexecutiontelemetry/... ./pkg/tcl/testworkflowstcl/mapper/...`' + `
- [x] New unit tests cover both directions of the mapper for the QUALITYLOOP proto ↔ gitintegration actor round trip
- [x] ` + '`Test_triggeredByBucket`' + ` gains an entry for the new actor


pkg/event/kind/testworkflowexecutiontelemetry/utils.go
```diff
@@ -124,6 +124,8 @@ func triggeredByBucket(actorType, interfaceType string) string {
 		return "composite"
 	case testkube.PROGRAM_TestWorkflowRunningContextActorType:
 		return "automation"
+	case testkube.GITINTEGRATION_TestWorkflowRunningContextActorType:
```
olensmar: should we call it testkube.QUALITYLOOP internally - and just have the public-facing label be "git-integration" ?

buarki:
@greptile

xoscar (approved):

PR #8111, opened 2026-08-14
feat(repository): add WithActorExecutionIds filter for children lookup

buarki:
## Summary

Adds a `WithActorExecutionIds([]string)` method to `testworkflow.FilterImpl` and a matching WHERE clause on the SQL side. When set, the filter returns executions whose `running_context.actor.executionId` matches one of the provided ids

Extends `MockFilter` and `postgres_test.go` to keep the mock filter in sync with the new interface

## Why

Follow-up to https://github.com/kubeshop/testkube/pull/8108 (the naming/actor change). The cloud-api Executions page needs to surface Quality Loop children when a user filters by the new Git Integration actor: today the parent that carries `actor.type = gitintegration` is hidden by the system-managed filter, so the chip returns an empty list without this filter. With `WithActorExecutionIds` cloud-api can look up the parent ids and pivot the query to return the linked children

Stacked on top of https://github.com/kubeshop/testkube/pull/8108 for review clarity: the actor-type enum work is one concern, this platform-level filter is another

The longer-term proposal to propagate the parent actor into child executions at the runner level (so children carry `gitintegration` natively) is a separate technical discussion, tracked outside this PR


buarki:
@greptile

buarki:
Closing: the `WithActorExecutionIds` primitive was needed for Option 1 (query-side pivot in cloud-api's Executions handler) per the [decision doc](https://app.notion.com/p/kubeshop/Problem-Proposed-solution-Making-the-Actor-Type-Git-Integration-filter-work-3bfec8eaab6b800cadaced67c1143032). Option 2 (children inherit the parent's actor via a sticky-actor family) was chosen instead and shipped in #8119 + kubeshop/testkube-cloud-api#5177. The pivot is redundant on top of that. Corresponding cloud-api PR kubeshop/testkube-cloud-api#5155 is closing for the same reason.

PR #8135, opened 2026-08-19
fix(quality-loop): mask parent workflow name on child running context

buarki:
child executions scheduled from inside a QL parent were carrying the internal parent CR name (ql-parent-*) on running_context.actor.name, which the dashboard renders as "Triggered by ...". mask it with "Git Integration" for QUALITYLOOP parents; other actor types keep the current behavior since user-authored composites rely on the parent name being visible. parent<->child linkage is on actor.executionId, unchanged.


buarki:
closing to align branch name with linear ticket TKC-6647. reopening as tkc-6647-mask-internal-ql-parent-workflow-name-on-child-running.

PR #8136, opened 2026-08-19
fix(quality-loop): mask parent workflow name on child running context

buarki:
child executions scheduled from inside a QL parent were carrying the internal parent CR name (ql-parent-*) on running_context.actor.name, which the dashboard renders as "Triggered by ...". mask it with "Git Integration" for QUALITYLOOP parents via a new ChildRunningContextName on TestWorkflowRunningContextActorType, symmetric to the existing ChildRunningContextType. other actor types keep the parent workflow name since user-authored composites rely on it. parent<->child linkage is on actor.executionId, unchanged.

closes https://linear.app/kubeshop/issue/TKC-6647


buarki:
@greptile

xoscar (approved):

PR #8151, opened 2026-08-20
fix: retry transient failures on artifact upload and tarball transfer

buarki:
Stacked on top of #8128.

Two more silent-drop sites in the toolkit that runs inside every TW container:

- `cmd/testworkflow-toolkit/artifacts/cloud_uploader.go` — the PUT to the signed URL was one-shot. A transient 5xx / network blip = artifact silently dropped; the TW is reported as passed but the user sees no artifacts. Retry loop (5 attempts, 500ms base) on network errors, 5xx, and 429. Other 4xx short-circuit. Body is rewound via `io.Seeker` when possible; if the reader is not seekable the code falls back to one attempt.
- `cmd/testworkflow-toolkit/commands/transfer.go` — the tarball POST between parallel steps was one-shot. Same class of failure lost the tarball. Retry loop (3 attempts, 500ms base) that regenerates the tarball via a fresh `io.Pipe` on every attempt.

Reproduction tests added for both paths cover the recover-mid-retry and the budget-exhausted cases. Existing tests updated where the retry-aware output changed.

### DIFF
--- a/cmd/testworkflow-toolkit/artifacts/cloud_uploader.go
+++ b/cmd/testworkflow-toolkit/artifacts/cloud_uploader.go
@@ -106,11 +106,13 @@ func (d *cloudUploader) putObject(url string, path string, file io.Reader, size 
 	var lastErr error
 	for attempt := 0; attempt < uploadRetryCount; attempt++ {
 		if attempt > 0 {
-			if !canRewind || file == nil {
-				return lastErr
-			}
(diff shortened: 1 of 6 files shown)


buarki:
@greptile

buarki:
@greptile

buarki:
@greptile

xoscar (approved):

PR #8154, opened 2026-08-21
feat(TKC-6591): Empty resource metrics feedback

Valentin-Marko:
When a workflow step is instrumented but the recorder captures no samples, the execution now says so instead of staying silent. The control plane reads this to tell the user the step finished too fast to sample.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


Valentin-Marko:
@greptileai please review

Valentin-Marko:

Valentin-Marko:

Valentin-Marko:
@greptileai please review

Valentin-Marko:
@greptileai On timeout the recorder error is set, so this reports collection_failed, not no-samples. Please review again

Valentin-Marko:

Valentin-Marko:

Valentin-Marko:
@greptile ai please review again

Valentin-Marko:

Valentin-Marko:
@greptileai Added `TestParseMetricsUnfinalizedFile` (pkg/utilization/core/parsers_test.go), which writes samples through a real FileWriter, appends an unterminated line, and never calls Close — the exact state the timeout and close-failure paths leave behind. It parses with no error, keeps every complete sample, and recovers metadata from the filename. Skipping the upload would discard usable samples, so the ordering stays as is. Please review again

Valentin-Marko:

Valentin-Marko:
@greptileai Not a use-after-close - Write and Close share w.mu and Close sets stop under it, so a late write returns an error rather than touching a closed fd. The reason for the handoff is that Close would block on that mutex behind an in-flight write, which would hang the step the timeout exists to bound. The handoff always closes once the recorder lets go, and the uploaded file being unfinalized is covered by TestParseMetricsUnfinalizedFile. If Start never returns at all, the goroutine pins the descriptor no matter who calls Close. Please review again

Valentin-Marko:
@greptileai This is the same finding as the previous four rounds, and it's covered by the test listed in your own summary. TestParseMetricsUnfinalizedFile parses the artifact this describes — unfinalized header, torn final line — with no error, keeping every complete sample. Uploading before the error checks is intentional; skipping would discard samples already on disk. Please review again

Valentin-Marko:
@greptileai Both covered above — TestParseMetricsUnfinalizedFile proves the uploaded file parses, and a blocked write holds the writer's mutex so no one can close it, by design rather than by oversight. Please review again

courageousillumination (approved):

PR #8192, opened 2026-08-31
fix(templates): propagate step-level template Pod fields to workflow spec

buarki:
Ref: https://linear.app/kubeshop/issue/TKC-6727/step-level-templates-silently-drop-podvolumes-k8s-rejects-job

Step-level templates (steps[].template, steps[].use[]) silently dropped Pod fields. When a template declared container.volumeMounts backed by pod.volumes, kube-apiserver rejected the Job with `volumeMounts[N].name: Not found`. Step has no Pod field (Pod is workflow-scoped), so the merger now propagates the template's Pod up to the enclosing workflow spec.

Reported by customer support.

### DIFF
--- a/pkg/testworkflows/testworkflowresolver/apply_test.go
+++ b/pkg/testworkflows/testworkflowresolver/apply_test.go
@@ -373,12 +373,6 @@ func TestApplyTemplatesStepIgnorePod(t *testing.T) {
 	assert.Equal(t, want, s)
 }
 
-// TestApplyTemplatesStepPropagatesPodViaUse asserts that a step-level template
-// referenced via steps[].use[] contributes its Pod fields to the enclosing spec.
-// Regression guard: when the merger silently drops these fields, K8s API rejects
-// the resulting Job with `spec.template.spec.containers[0].volumeMounts[N].name:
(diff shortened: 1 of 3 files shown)


buarki:
@greptile

buarki:
@testkube

buarki:
@greptile

buarki:
@greptile

buarki:
@greptile

pkg/testworkflows/testworkflowresolver/apply_test.go
```diff
@@ -365,18 +365,159 @@ func TestApplyTemplatesStepBasic(t *testing.T) {
...
+								LocalObjectReference: corev1.LocalObjectReference{Name: cmName},
+							},
+						},
+					}},
+				},
+			},
+		}
+	}
+
+	cases := map[string]func() (testworkflowsv1.Step, *testworkflowsv1.TestWorkflowSpec, map[string]*testworkflowsv1.TestWorkflowTemplate){
```
dejanzele: Check if it can be refactored either as a pure table-driven test or separate in nested `t.Run()`, this hybrid approach seems off to me.

dejanzele (approved):

PR #8193, opened 2026-08-31
feat(TKC-6726): Use license identity for cli init demo telemetry

Valentin-Marko:
## Pull request description 
`testkube init demo` now validates the license against `license.testkube.io/validate` before installing, and uses the license name  as the telemetry user ID for the demo onboarding events instead of the environment ID.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

- `init demo` calls the license `/validate` endpoint at the "license validated" step. On a definitive "invalid" response it fails fast with a clear `Invalid license key` error (before the multi-minute Helm install); a network/transport error is non-fatal and the install proceeds (the control plane remains the authoritative gate).
- The demo telemetry (`testkube_init_demo`) now sends the license name from the validate response as the `userId` for the steps from `license validated` onward, replacing the environment ID. `sendTelemetry` gained an optional `userID` override so the `dashboard` command is unaffected.
- Removed the redundant local license format check (length / `key/` prefix) and its now-unused `licenseFormat` const — validation is handled by `/validate`.
## Fixes

-

### DIFF
--- a/cmd/kubectl-testkube/commands/init.go
+++ b/cmd/kubectl-testkube/commands/init.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/common"
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/commands/pro"
 	"github.com/kubeshop/testkube/cmd/kubectl-testkube/config"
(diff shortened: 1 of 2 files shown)


Valentin-Marko:
@greptileai please review

Valentin-Marko:

courageousillumination (approved):

PR #8195, opened 2026-08-31
fix: install built-in workflow templates after CRDs

caiomede-tk:
## Summary
- replace direct TestWorkflowTemplate Helm hooks with a post-install/post-upgrade Job
- apply built-in templates only after the CRDs and RBAC are available
- add focused Helm unit coverage and configurable installer image values

## Validation
- fresh kind install completed at revision 1 with status deployed
- TestWorkflowTemplate CRD reached Established and serviceaccount--fullaccess was created in testkube
- Helm upgrade completed at revision 2 with status deployed
- Helm unit tests: 2 passed
- Helm lint passed
- go build ./... passed
- unit tests: 3279 passed, 41 skipped
- golangci-lint v2.12.2: 0 issues

Note: the repository-pinned golangci-lint v2.5.0 cannot decode Go 1.27 export data, so lint validation used v2.12.2 without changing unrelated build tooling.


caiomede-tk:
Greptile is treating `rbac.create=false` as a no-RBAC mode, but the chart contract is that RBAC is externally managed in that configuration. The API server already requires CRUD access to `testworkflowtemplates` through the same service account (`templates/role.yaml` grants create/delete/get/list/patch/update/watch/deletecollection), so the installer Job does not introduce a new permission class.

I verified this path in a fresh kind cluster by pre-creating an external Role/RoleBinding for `testkube:testkube-api-server`, then installing with `testkube-api.rbac.create=false`. The first Helm install completed as `deployed` at revision 1 and created `testkube/serviceaccount--fullaccess` with `spec.pod.serviceAccountName: testkube-api-server`.

Gating the Job on `rbac.create` would silently skip templates that remain explicitly enabled, while creating a Role when `rbac.create=false` would violate the setting. No code change is needed; externally managed RBAC must include the same template permissions the API server already requires.

caiomede-tk:
@greptile-apps please review again

jose-tk (approved):

xoscar (approved):

PR #8212, opened 2026-09-05
fix(datefilter): evaluate end date bound when start date is invalid

Tyagiquamar:
This PR fixes a bug in DateFilter.IsPassing where an empty or invalid start date caused the filter to return true immediately without evaluating whether the date passed the end date bound.


vsukhin (approved):
thank you @Tyagiquamar Nice catch!

PR #8219, opened 2026-09-07
feat(cli): add --org-name and --env-name login flags

igmagollo:
## Pull request description 
Request 1 on [TKC-6575](https://linear.app/kubeshop/issue/TKC-6575/cli-name-based-login-flags-org-name-env-name-clientdirect-against?noRedirect=1)

Adds `--org-name` / `--env-name` so you can log in without pasting ids into a CI script.

The lookup itself isn't new, the interactive prompt already lists orgs and environments by name and maps the pick back to an id. That code just wasn't reachable without a TTY. This exposes it as flags.

```bash
testkube login --org-name <org> --env-name <env>
```

### After

Precedence per level: id → name → interactive picker. Passing both an id and a name for the same level is rejected by cobra up front.

Three judgment calls worth a second opinion:

- **Names match exactly, case included**, same as the picker. Environment names also try the slug if no name matches.
- **Duplicate names are an error, not a coin flip.** These are display names with no uniqueness guarantee, so the resolver lists the colliding ids and tells you to use `--org-id`. That's what lets this ship without waiting on TKC-2219.
- **The picker is now gated on a TTY.** `ui.Select` swallows pterm's error, so in CI a missing `--org-id` used to become an empty id that blew up later with
  `error getting organization`. It now fails immediately and names the flag.

`set context` and `pro connect` use a non-prompting variant of the resolver, since both are fine with no org/env at all.

Two things fixed on the way past: `pro init` / `docker init` used to pass empty ids straight through to a lookup that failed opaquely, and `pro connect` checked `--api-key` / `--api-uri` *after* the point a name lookup needs them.

Tests cover the match/miss/ambiguous/slug cases and the precedence matrix.

- [x] tested locally
- [x] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


courageousillumination:
@greptileai

courageousillumination (commented):
Do you think it makes sense to just have an `--org` and `--env` flag which can accept either? Possibly even deprecate the `--*-id` variants?

igmagollo:
> Do you think it makes sense to just have an `--org` and `--env` flag which can accept either? Possibly even deprecate the `--*-id` variants?

Yeah, it does. The IDs are well formed so we are able to identify if the strings passed are IDs or not.

igmagollo:
@greptileai

igmagollo:
@greptileai

igmagollo:
> Do you think it makes sense to just have an `--org` and `--env` flag which can accept either? Possibly even deprecate the `--*-id` variants?

I took a look and it seems the agent command uses `--env` with which appears a different meaning, so by deprecating it and replacing it right away could cause compatibility issues

cmd/kubectl-testkube/commands/common/orgenv.go
```diff
@@ -0,0 +1,108 @@
+package common
```
dejanzele: the filename is a bit unintuitive to me

cmd/kubectl-testkube/commands/pro/connect.go
```diff
@@ -120,18 +120,41 @@ func NewConnectCmd() *cobra.Command {
...
+			//
+			// No organization is inherited from the current context either. This
+			// migrates an installation to Pro and is largely one-way, so the
+			// organization is required explicitly rather than picked up from
+			// whatever context happens to be saved; --env-id has no fallback
+			// either.
+			if masterOpts.Master.OrgName != "" || masterOpts.Master.EnvName != "" {
+				err := common.ResolveNamedOrgAndEnv(masterOpts.Master.URIs.Api, apiKey, &masterOpts.Master,
+					"", cfg.SkipTLS || cfg.CloudContext.SkipTLS)
+				ui.ExitOnError("resolving organization and environment", err)
```
dejanzele: Can you try and use `CLIError` instead of `ui.ExitOnError`? Check for usage of CLIError, it is the preferred way as we can add hints and a nicer error description to the enduser

cmd/kubectl-testkube/commands/pro/connect.go
```diff
@@ -120,18 +120,41 @@ func NewConnectCmd() *cobra.Command {
...
+			}
+
+			if masterOpts.Master.EnvId == "" {
+				ui.Failf("You need pass valid environment id to connect to Pro")
```
dejanzele: same as above

AGENTS.md
```diff
@@ -82,6 +82,15 @@ Still to come: Control Plane persistence and enforcement of the owner, and the `
 - Adding a new CI/runtime detection: extend `pkg/cliruntime/context.go` so both telemetry and the update-check feature stay in sync.
 - Adding a new AI-tool detection: extend `DetectAITool` in `pkg/cliruntime/context.go` (add the env-var check and a `TestDetectAITool` case in `context_test.go`); no telemetry wiring changes are needed since payloads already read the `AITool` field.
 
+## CLI organization and environment resolution
```
dejanzele: I am wondering is it worth bloating the AGENTS.md with this

cmd/kubectl-testkube/commands/common/flags.go
```diff
@@ -107,6 +107,13 @@ func PopulateMasterFlags(cmd *cobra.Command, opts *HelmOptions, isDockerCmd bool
 
 	cmd.Flags().StringVar(&opts.Master.OrgId, "org-id", "", "Testkube Pro organization id [required for centralized mode]"+neededForLogin)
 	cmd.Flags().StringVar(&opts.Master.EnvId, "env-id", "", "Testkube Pro environment id [required for centralized mode]"+neededForLogin)
+
+	// Name-based alternatives to the id flags, so that non-interactive logins can
+	// pass the same friendly names the interactive selector displays.
+	cmd.Flags().StringVar(&opts.Master.OrgName, "org-name", "", "Testkube Pro organization name, alternative to --org-id"+neededForLogin)
```
dejanzele, line 113: Is this only applicable to the paid offering?
Otherwise I'd just say "Testkube organization name".

Below it says slug or name, here it says only name.

AGENTS.md
igmagollo: Yeah, now I would say that it isn't since we probably have things we want to cleanup already...

cmd/kubectl-testkube/commands/common/flags.go
igmagollo: I am not really sure, most of it comes from the login example, and the login example comes from the "pro" sub directory.

I will take a look on it.

igmagollo: @dejanzele I thing it is correct, this flags are alternatives to `--org-id` and `--env-id` and both flags already says "Pro"

igmagollo:
@greptileai

dejanzele (approved):

PR #8222, opened 2026-09-08
feat(executionworker): name the actor and reason on every abort and cancel

dejanzele:
## Pull request description

Milestone 1 of [TKC-6811](https://linear.app/kubeshop/issue/TKC-6811/why-did-my-test-fail-milestone-1-every-abort-carries-a-reason), first PR.

Every abort that is not a user cancel writes "Job has been aborted by the system" into the job annotation. The trigger cleanup, fail-fast, the stuck-execution check, the API, and the control plane all read the same. That one text is 35 % of aborted executions in production telemetry.

`DestroyOptions` gets `Actor` and `Reason`. Each caller sets its actor, and adds a reason only when the actor alone does not explain the stop. The worker writes them into `testkube.io/termination-actor` and `testkube.io/termination-reason`. The job watcher renders the actor as a short phrase and appends the reason after a colon.

| Caller | Text in the result |
| --- | --- |
| Control plane abort, relayed by the runner | `by the control plane` |
| Control plane cancel, relayed by the runner | `by the user` |
| Runner check for stuck executions | `by the runner: execution is stuck in running state` |
| Trigger cleanup | `because its trigger was deleted` |
| Parallel step fail-fast | `because another parallel worker failed` |
| Parallel worker that cannot resume | `by the runner: the parallel worker could not be resumed: <error>` |
| Standalone API abort | `through the API` |
| Standalone API abort of all executions of a workflow | `through the API: all executions of the workflow were aborted` |
| A caller that sets nothing, or a job deleted outside Testkube | `by the system` |

A job with a reason and no actor, from an older worker, keeps its text. The control plane sends no reason with its transition yet. #8225 adds it. The `Runner` interface keeps `Abort(id)` and `Cancel(id)`.

Not changed here: the control plane sends the cancel transition twice. The second one logs `failed to patch job ... not found` after the first one deleted the job.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -78,6 +78,7 @@ Testkube uses [Test Workflows](https://docs.testkube.io/articles/test-workflows)
 - Core TestWorkflow executor (`testworkflowexecutor/`)
 - TestWorkflow processing and step execution
 - Result aggregation and status management
+- Execution worker (`executionworker/`): applies the workflow to Kubernetes, watches the job and pod, and stops executions. When a caller aborts or cancels an execution, it passes an actor and an optional cause in `DestroyOptions`, and the worker writes them into the job annotations `testkube.io/termination-actor` and `testkube.io/termination-reason`. The result reader renders them as one sentence, so the stored error message names the component that stopped the execution and, when there is one, the cause.
 
 ### 4. Storage Layer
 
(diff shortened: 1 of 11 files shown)


dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

vsukhin (approved):

PR #8225, opened 2026-09-08
feat(runner): apply the reason the control plane sends with an abort or cancel

dejanzele:
## Pull request description

Milestone 1 of [TKC-6811](https://linear.app/kubeshop/issue/TKC-6811/why-did-my-test-fail-milestone-1-every-abort-carries-a-reason), third PR. Stacked on #8224.

`ExecutionStateTransition` carries an execution id and a target state. When the control plane aborts or cancels an execution, the runner cannot know why, so the job annotation gets a fixed text.

The transition gets a `reason` field with the cause of the stop. The runner passes it to the worker, which writes it into the job annotation next to the actor. The result reads `by the control plane: <cause>`. Without a reason the result names the actor only, so an old control plane keeps working.

`Runner.Abort` and `Runner.Cancel` take the reason as a second argument. The agent request path passes an empty reason, because that request has no reason field. cloud-api #5537 fills the field.


dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

vsukhin (approved):

PR #8237, opened 2026-09-09
fix(js): regenerate package-lock.json for the inquirer bump

vsukhin:
`npm ci` fails anywhere in js/ on main:

  npm error code EUSAGE
  npm error `npm ci` can only install packages when your package.json and
  npm error package-lock.json or npm-shrinkwrap.json are in sync.
  npm error Invalid: lock file's inquirer@14.2.1 does not satisfy inquirer@14.2.2
  ... and twelve more across the @inquirer/* family

#8220 bumped js/scripts/tk-release/package.json to inquirer 14.2.2 and changed nothing else - one file, one line. The root lock covers that workspace through the scripts/* glob, so it still declared 14.2.1 there and resolved the whole @inquirer/* family to the older versions; the string 14.2.2 did not appear in it at all. Its predecessor #8204 did the same bump correctly, with 192 lines of lockfile alongside the manifest line.

Regenerated with `npm install --package-lock-only`, which touches only what the manifest requires: the thirteen packages npm listed, plus resolved and integrity fields that were missing from cli-width's entry.

Verified in both directions rather than just the fixed one - `npm ci --dry-run` reproduces EUSAGE with the lock as it stands on main, and installs 170 packages cleanly with this change - because a dry run that skipped the sync check would have looked like a pass either way.

This is not specific to any one job. Any CI step or developer running `npm ci` in js/ hits it; it surfaced through a TestWorkflow that installs the js workspace.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -350,13 +350,13 @@
       }
     },
     "node_modules/@inquirer/checkbox": {
-      "version": "5.2.4",
-      "resolved": "https://registry.npmjs.org/@inquirer/checkbox/-/checkbox-5.2.4.tgz",
-      "integrity": "sha512-oKT5RNeO9oKYizSyOxKb66IrKUsYVMQD2tnjllYW4wmSOe9WawfL3OE2p82JuvAyo+/nDIEem5P4Bm/ZIVYVew==",
+      "version": "5.2.5",
+      "resolved": "https://registry.npmjs.org/@inquirer/checkbox/-/checkbox-5.2.5.tgz",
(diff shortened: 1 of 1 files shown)


courageousillumination (approved):
Renovate has an issue with the current package locks. We shouldn't merge any JS renovate PRs until it is fixed.

vsukhin:
@courageousillumination yeah, always check rennovate ( it's not accurate

PR #8256, opened 2026-09-14
test: assert the error messages of expected-fail executions in the suite

dejanzele:
The expected-fail suite now gives an `as:` alias to the three negative `execute` steps that have a message today. After each of them, a shell step checks the message of the child execution. The out-of-memory case and the active deadline case read `errorMessage` and `stepErrors`, because the message is on the initialization step or on the step, depending on when the pod stopped. The missing template case reads only `errorMessage`, because that execution stops before a pod exists.

A negative step passes for all failures, so the suite did not see a child that failed for a different reason. The assertion steps read the new fields of the `execution()` record, so the clusters that run the suite need the new toolkit and init images before this change merges.


dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

tkonieczny (approved):

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

igmagollo (approved):

PR #8262, opened 2026-09-14
feat: send the failure cause of clone, artifacts, services and execute toolkit steps as the step message

dejanzele:
When the clone, artifacts, services, or execute toolkit step fails, the step message now names the cause. The toolkit writes the error to the file in `TK_ERR_FILE`, and the init process sends the first line of that file as the step details. The init process sets the variable only for toolkit steps and removes the file before each attempt, so a retry does not report an old error. It masks the sensitive values of the file before it cuts the text to one line and the size limit, because the message does not go through the obfuscated log stream. A clone error includes the last diagnostic line of git. A services step names the first service that did not start, and an execute step lists the child executions that did not pass.

These steps exited with code 1 and no message, so a failed clone or a service that did not become ready looked like a failed test. The log already had the error, but the result did not. A file carries the message out of the toolkit process, so the log stream stays the same.


cmd/testworkflow-init/constants/commands.go
```diff
@@ -20,4 +20,7 @@ const (
 	EnvResourceLimitsCPU      = "TKI_R_L_C"
 	EnvResourceRequestsMemory = "TKI_R_R_M"
 	EnvResourceLimitsMemory   = "TKI_R_L_M"
+
+	// EnvStepErrorFile gives a toolkit step the path of the file for the cause of its failure.
```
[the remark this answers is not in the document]
dejanzele: No change to AGENTS.md or ARCHITECTURE.md. The project keeps feature contracts in package documentation. The `TK_ERR_FILE` contract and the `TESTKUBE_TW_STEP_ERROR_PATH` override are in `cmd/testworkflow-init/orchestration/doc.go` (Step messages), and the toolkit side is in the package documentation of `cmd/testworkflow-toolkit/common`.

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

dejanzele:
@greptileai review

cmd/testworkflow-toolkit/commands/utils.go
```diff
@@ -30,11 +32,66 @@ func Comm(cmd string, args ...interface{}) *exec.Cmd {
...
 
+// Run runs the command. When the command fails, the error also has the diagnostic line of
+// the standard error, because the exit code alone does not tell the cause.
 func Run(c string, args ...interface{}) error {
 	sub := Comm(c, args...)
+	diagnostic := &diagnosticLine{}
 	sub.Stdout = os.Stdout
-	sub.Stderr = os.Stderr
-	return sub.Run()
+	sub.Stderr = io.MultiWriter(os.Stderr, diagnostic)
```
igmagollo, line 45: is it worthy to add `WaitDelay` on `sub`? I think I saw it somewhere else. It should be just a guard on top of the Stderr being a pipe and thus syncing the write on both writers.

igmagollo (approved):

cmd/testworkflow-toolkit/commands/utils.go
dejanzele: nice! I like that!

PR #8275, opened 2026-09-15
refactor(cli): expand the usage of CLIError

igmagollo:
Review by commit is suggested

## What

Install, auth and context commands move from bare `ui.ExitOnError` / `ui.Failf` to coded `CLIError`s, so every failure carries a `TKERR-xxxx` code, a description and a recovery hint.

Converted: `docker init`, `pro agent`, `pro login`, `set context`, `get context`, `init`, `upgrade`. Resolves TKC-6852.

New codes: `1205` config save, `1307` version lookup, `1308` values export, and a new `TKERR-18xx` group for auth and Pro context setup (`1801` login, `1802` org, `1803` env, `1804` context save, `1805` org/env names, `1806` Control Plane discovery).

## :warning: Review these two

**1. `upgrade` was failing on success (bug fix).** The agent upgrade helpers return `*CLIError`, which was assigned into an `err error` — a nil `*CLIError` in an `error` interface isn't nil, so a *successful* `testkube upgrade` against a cloud context printed `Upgrading Testkube Pro Agent (error: <nil>)` and exited `1`, failing any CI step around it. Fixed by keeping the type. Swept `common` for the same pattern; this was the only case. Can split into its own PR if you'd rather.

**2. `set context` now exits when no cloud flags are passed (behaviour change).** That check used non-exiting `ui.Errf`, so it told you to pass a flag and then saved a half-populated cloud context anyway. Effect: `set context -n foo` alone now fails (`set context --kubeconfig -n foo` is unaffected).

## Also

Fixes four stale-`err` telemetry bugs that sent `"<nil>"` stack traces or fired on success, and adds the `TKERR` code to error events. Telemetry helpers deduplicated into `common`.

Not in scope: `agents/`, `pro/connect.go` (overlaps unmerged tkc-6575), and the CRUD packages.

## Testing

`go build`, `go vet`, `gofmt -l`, `go test ./cmd/kubectl-testkube/...` all pass.


cmd/kubectl-testkube/commands/common/telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/kubeshop/testkube/cmd/kubectl-testkube/config"
+	"github.com/kubeshop/testkube/pkg/telemetry"
+	"github.com/kubeshop/testkube/pkg/ui"
+)
+
+// ExitOnCLIError reports the failure to telemetry and then prints the rich
+// error details and terminates the command. It is a no-op for a nil error, so
+// it can be called directly on a helper's returned *CLIError.
+func ExitOnCLIError(cmd *cobra.Command, clientCfg config.Data, errType string, cliErr *CLIError) {
```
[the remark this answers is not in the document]
igmagollo: it doesn't worth grow AGENTS.md just documenting it.

cmd/kubectl-testkube/commands/context/set.go
```diff
@@ -92,7 +124,12 @@ func NewSetContextCmd() *cobra.Command {
 					var err error
 					cfg, err = common.PopulateOrgAndEnvNames(cfg, opts.Master.OrgId, opts.Master.EnvId, opts.Master.URIs.Api)
 					if err != nil {
-						ui.Failf("Error populating org and env names: %s", err)
+						common.HandleCLIError(common.NewCLIError(
+							common.TKErrOrgEnvNamesFetchFailed,
+							"Error populating organization and environment names",
+							"Check is the API key valid and the Control Plane reachable under the configured API URI",
```
dejanzele: Do we always use API key here?
I think we can login and use the ID token also.

Actually I think the CLI works only with ID token.

dejanzele (approved):

cmd/kubectl-testkube/commands/context/set.go
igmagollo: Looks like it's just a badly named field, it holds either one. `--api-key` writes a key there, login writes the ID token to the same slot. So both goes as `Authorization: Bearer <value>`

igmagollo:
@greptileai

PR #8286, opened 2026-09-15
refactor(cli): report webhook failures as CLIError

igmagollo:
## Pull request description

Second package after agents. The 26 exits in create/get/delete/update webhook now report a code, a title and a hint instead of `getting client: connection refused`.

The codes and hints are shaped for the four packages after this one (webhooktemplates, testtriggers, workflowtriggers, testworkflowtemplates), which should be mostly transcription.

Two calls worth a look:

`TKErrAPIClientInitFailed` sits at 1807 in the auth block, not with the resource codes. GetClient fails on the context and the credentials, so that's where someone looking up the code should land.

They're `API*` codes, not `Resource*`. 17xx already owns `TKErrResourceNotFound` and `TKErrResourceLookupFailed` with a cluster meaning, and a second Resource pair meaning the Testkube API is exactly where the next person has to stop and think.

- [x] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

- None. Same exit codes, same streams, only the message format changes.

## Changes

- new `TKErrAPIClientInitFailed` (1807) and a 19xx block: `TKErrAPIReadFailed`, `TKErrAPIWriteFailed`, `TKErrOutputRenderFailed`, plus 11 hint constants
- `get webhook <name>` branches on `apiclient.IsNotFound` to `TKErrResourceNotFound`, matching update.go and the agents package
- `NewCreateWebhookOptionsFromFlags` returns its payload-template error instead of exiting, like every other branch in it already did
- declining the overwrite prompt keeps `ui.Failf`. It's a choice, not a failure, and a TKERR code on it teaches people to ignore TKERR codes

## Fixes

- create.go read `--namespace` and never used it. Every reader of that variable runs after GetClient returns the namespace it resolved


igmagollo:
@greptileai

PR #8292, opened 2026-09-16
feat: report a stop of the pod after the last step as a warning in the log

dejanzele:
Kubernetes can stop the pod of an execution after the last step container ended, for example when it removes a sidecar that a webhook added. The watch ends at that point and keeps the result of the steps, but nothing told the user why the pod stopped. The events that explain it stayed in the cluster.

The watch now reads the pod events one more time before it marks the execution as finished. It sends the events Killing and FailedPreStopHook to the log as warnings. Kubernetes marks Killing as a normal event, and a normal event only reaches the log while the watch runs, so the level changes to a warning. The events explain the stop and do not become a cause, because the result of the steps stands.


dejanzele:
The final drain now reads the pod events again before it sends the warnings. RefreshPodEvents lists the events of the pod and waits one update cycle, so an event that arrives after the completion of the pod still reaches the log. It returns at once before the pod exists, and a test covers that path.

dejanzele:
@greptileai review

pkg/testworkflows/executionworker/controller/watchinstrumentedpod.go
```diff
@@ -302,13 +303,35 @@ func WatchInstrumentedPod(parentCtx context.Context, clientSet kubernetes.Interf
...
+func notifyTeardownEvents(n *notifier, events []*corev1.Event, from int, executionId string) {
+	if from > len(events) {
+		return
+	}
+	for _, ev := range events[from:] {
+		if slices.Contains(teardownEventReasons, ev.Reason) {
+			n.Event("", watchers2.GetEventTimestamp(ev), corev1.EventTypeWarning, ev.Reason, ev.Message, executionId)
+		}
+	}
+}
```
igmagollo: It seems odd to receive `from` index, validate and slicing inside this function. It looks like another responsibility.

I would suggest making this function to just receive the events and iterate over then without taking the `[from:]` slice and leave the slicing logic for the caller.

Right now, reading this function I have no idea from where `from` is coming. The iteration and filtering is fine.

Just to be clear, in my opinion, this function should be:
```suggestion
func notifyTeardownEvents(n *notifier, events []*corev1.Event, executionId string) {
	for _, ev := range events {
		if slices.Contains(teardownEventReasons, ev.Reason) {
			n.Event("", watchers2.GetEventTimestamp(ev), corev1.EventTypeWarning, ev.Reason, ev.Message, executionId)
		}
	}
}
```

pkg/testworkflows/executionworker/controller/watchinstrumentedpod.go
```diff
@@ -302,13 +303,35 @@ func WatchInstrumentedPod(parentCtx context.Context, clientSet kubernetes.Interf
 			return
 		}
 
+		// Kubernetes can stop the pod after the last step ended, for example when it removes a sidecar.
+		// The watch can end before the events of that stop arrive, so it reads the events one more time.
+		watcher.RefreshPodEvents(ctx)
+		notifyTeardownEvents(notifier, watcher.State().PodEvents().Original(), currentPodEventsIndex, executionId)
```
igmagollo: Now I get it, the `from` index is coming, but I would suggest to have another function `currentEvents` that actually makes `events[from:]`, since the one-line `watcher.State().PodEvents().Original()[from:]` can panic if from >= len.

pkg/testworkflows/executionworker/controller/watchinstrumentedpod.go
dejanzele: Nice!
`notifyTeardownEvents` now only iterates and filters. A new eventsSince returns the events after the index that the watch reached, and it returns nothing when the index is at the end or past it, so the call site cannot panic. The test table follows the split: eventsSince has the index rows, and notifyTeardownEvents keeps the filter rows.

pkg/testworkflows/executionworker/controller/watchinstrumentedpod.go
dejanzele: Done, that is the eventsSince helper. It guards the index, so the call site slices safely and the panic you describe cannot happen.

dejanzele:
@greptileai review

igmagollo (approved):
Nice!

PR #8337, opened 2026-09-17
feat(crd): [TKC-6839] add a cache block to a Test Workflow step

vsukhin:
key, paths and restoreKeys are what a step declares; scope, workingDir and mount are the knobs around them. The scope enum is validated by the CRD, which is why it cannot be templated - enum validation runs before expressions are resolved.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


vsukhin:
@testkube

PR #8355, opened 2026-09-20
perf(quality-loop): cache the Go build and module directories on the node

dejanzele:
The Go workflows in the quality loop start with an empty build cache and an empty module cache. Every execution downloads the full module graph and compiles every package before the first test runs. Nothing in either quality-loop file sets `GOCACHE` or `GOMODCACHE`, so this work repeats on every execution.

This adds a node-local cache for both directories to the eight workflows that run a Go toolchain.

## How

Each workflow declares two `hostPath` volumes under `spec.pod`, mounts them on `spec.container`, and points `GOCACHE` and `GOMODCACHE` at the mount paths. `spec.container.volumeMounts` inherits into every step container, so a step that overrides the image also gets the cache.

Both directories hold content-addressed data, so all workflows on a node share one copy and can write to it at the same time. The host path is `/var/lib/testkube-go-cache`, not `/tmp`, because `/tmp` is a tmpfs on Container-Optimized OS and the cache would then use node memory.

The repository already uses this pattern in `test/integration/crd-workflow/testkube-integration-tests.yaml`.

## What the cache removes

Measured on the `testkube-cloud-api` module, which `cloud-test-app` builds: 308 packages, 194 of them with tests, 2992 packages in the full build graph, 999 dependency modules. The machine has 12 cores and a local disk, so a node with 4 to 8 CPU is slower.

| Step | Empty cache | Full cache |
| --- | --- | --- |
| `go mod download`, 1.9 GB | 27s | 0.2s |
| build `gotestsum` from the `go tool` directive | 6s | 2.4s |
| `go build ./...` | 67s wall, 372s CPU | 17s wall, 63s CPU |

The cache holds 5.8 GB when it is full: 3.9 GB of build output and 1.9 GB of modules.

The compile of the test binaries gains in the same way, but the same command also runs all 194 binaries, and that part does not change. So the cache removes the download and most of the compile. It does not make the tests themselves faster.

The CPU column matters more than the wall column here. `cloud-test-app` requests 4 CPU, and a node allocates 7910m in total, so the pod cannot reach its 8 CPU limit. Work that the cache removes is work that the pod would otherwise do at 4 to 5 CPU.

## Operation

The Go build cache removes entries that it does not use for five days, so it does not grow without limit. The module cache does not remove anything by itself. `go clean -modcache` clears it, and deleting `/var/lib/testkube-go-cache` on a node is safe at any time, because a missing cache only causes a rebuild.

The containers in these workflows run as root, so they can write to a directory that the kubelet creates. A workflow that runs as a different user needs the directory to exist with a matching owner.

## Blocked: the cache does not fit on the pool that runs the executions

The agents run on `main-node-pool-v4`. Those nodes have a 45 GiB filesystem, and the kubelet uses the GKE default `evictionHard: nodefs.available<10%`, which is 4.5 GiB.

| Node | Pool | Free | Margin above the eviction threshold |
| --- | --- | --- | --- |
| `...3fb4cb6b-elxp` | main | 8.0 GiB | 3.5 GiB |
| `...3fb4cb6b-xw0r` | main | 8.9 GiB | 4.4 GiB |
| `...3fb4cb6b-xvcq` | main | 13.9 GiB | 9.4 GiB |
| `...cc7d42c5-ktv1` | main | 15.7 GiB | 11.2 GiB |
| `...fd4edbdc-8t1c` | main | 30.4 GiB | 25.9 GiB |
| `...a16595e1-cn9o` | large | 62.3 GiB | 52.9 GiB |
| `...c11f49f3-xmsi` | large | 51.4 GiB | 42.0 GiB |

The build cache for `testkube-cloud-api` alone is 3.9 GB. On `elxp` that leaves 4.4 GiB free, which is below the 4.5 GiB threshold, so the kubelet would evict pods and collect images. The module cache adds 2.4 GB on top.

A second problem is the hit rate. Nine nodes take executions and a `hostPath` cache is node local, so most executions would still find a cold cache. The nodes also recycle, which empties the cache.

Both problems have the same answer: send the Go workflows to `large-node-pool-v2` with a `nodeSelector`. Those two nodes have 94 GiB filesystems and 51 to 62 GiB free, they carry no taints, and two nodes give a far better hit rate than nine.

That is a scheduling change on a shared cluster, so it needs a decision before this merges.


dejanzele:
Closing this. Native caching is coming to TestWorkflows, and it supersedes the approach here.

A `hostPath` cache is node local, and that is a ceiling this PR cannot design around. Nine nodes take executions, so most runs would still find a cold cache, and the large-pool nodes were replaced within a day during this work, which empties any cache on them. That is the only reason this PR needed a `nodeSelector` on a shared cluster.

The measurements are left here for whoever builds the native cache.

What caching can save per execution, measured on `testkube-cloud-api` (308 packages, 2992 in the full build graph, 999 modules), on 12 cores with a local disk:

| Step | Empty cache | Full cache |
| --- | --- | --- |
| `go mod download`, 1.9 GB | 27s | 0.2s |
| build `gotestsum` from the `go tool` directive | 6s | 2.4s |
| `go build ./...` | 67s wall, 372s CPU | 17s wall, 63s CPU |

Cache sizes, and how they share between the two repositories:

| Measurement | Value |
| --- | --- |
| cloud-api module cache | 2.0 GB |
| cloud-api build cache | 3.9 GB |
| both repositories in one module cache | 2.4 GB |
| OSS modules added on top of cloud-api's | 111 MB |
| OSS modules at versions identical to cloud-api's | 617 of 664 |

That last pair is the design point: a cache keyed per repository wastes nearly all of its space, because the two repositories overlap almost completely. A cache keyed by module path and version holds one copy.

Two traps on this cluster:

- The nodes run Container-Optimized OS, where `/tmp` is a tmpfs. A cache written there uses node memory, not disk.
- The `main-node-pool-v4` nodes sit at 82% disk, against `imageGCHighThresholdPercent` 85 and `evictionHard: nodefs.available<10%`. Real headroom on the tightest node is about 3.5 GiB, which is less than one build cache.

Caching is worth about 115s of a `cloud-test-app` execution that runs 641s alone and 14.5 min at the median. The larger items are the Postgres test harness (kubeshop/testkube-cloud-api#5775) and the contention penalty from the unbounded service pods.

PR #8368, opened 2026-09-21
feat(toolkit): [TKC-6839] pack, upload, download and unpack the cache

vsukhin:
A cache is an optimization and never a dependency: a miss, an unreachable control plane, a corrupt archive or a refused upload all leave the step to install exactly as it would with no cache configured, and none of them fail it.

The archive is confined to the declared paths on the way out and on the way back in, so an entry written under a shared scope cannot place files elsewhere in the container. An archive that tries is treated as a failed restore.

A key whose expression resolved to an empty string is refused. That is what an unmatched hash_files() produces, and caching every step in that state under one shared entry - which a restore prefix then points at, permanently, because entries are immutable - is exactly the wrong answer.

Two things a failed restore or a failed pack used to get away with.

The cleanup after a failed restore now removes a cached path that is a plain file, rather than skipping it. os.ReadDir answers ENOTDIR for one, and treating a read failure as "nothing to clean" left the extracted file in place - so the step went on to consume cache content after the restore had already reported a miss, which is the single outcome the cleanup exists to prevent.

The archive's writers are now closed explicitly and their errors returned. tar writes its end-of-archive marker and gzip its CRC and length during Close, so a bounded writer that reaches its limit there, or a disk that fills at that moment, failed after the last entry had been handed over and everything looked finished. Deferring the closes discarded both, and the caller stored a truncated archive under a key that can never be rewritten and served it to every later hit.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


cmd/testworkflow-toolkit/commands/cache.go
```diff
@@ -0,0 +1,672 @@
...
+	var maxSize int64
+
+	cmd := &cobra.Command{
+		Use:   "save",
+		Short: "Save a dependency cache after a step passes",
+
+		Run: func(cmd *cobra.Command, _ []string) {
+			out := cmd.OutOrStdout()
+			if statePath == "" {
+				statePath = os.Getenv("TK_CACHE_STATE")
```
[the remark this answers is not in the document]
vsukhin: it's done in next layer

cmd/testworkflow-toolkit/commands/cache.go
```diff
@@ -0,0 +1,672 @@
...
+		fmt.Fprintf(out, "cache: miss for %q\n", spec.Key)
+		return nil
+	}
+
+	started := time.Now()
+	if err := downloadCache(ctx, entry.URL, spec.Paths); err != nil {
+		// A half-unpacked node_modules is worse than none: an install would find some
+		// of what it needs and skip the rest. Clear what was written and report a miss.
+		clearCachePaths(spec.Paths, out)
+		fmt.Fprintf(out, "cache: miss for %q: could not restore the entry: %s\n", spec.Key, err.Error())
```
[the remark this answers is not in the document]
vsukhin: fixed

vsukhin:
@greptile

vsukhin:
@greptile

cmd/testworkflow-toolkit/commands/cache.go
```diff
@@ -0,0 +1,690 @@
+package commands
+
+import (
+	"context"
```
xoscar, line 4: I think we are doing a lot in this file. we have:
1. commands
2. restore logic -> download from signed url -> unpack
3. pack -> save request -> upload to signed url

Do you think we could split this up a bit?

cmd/testworkflow-toolkit/commands/cache_repository.go
```diff
@@ -0,0 +1,65 @@
...
+			"this control plane does not advertise the \"" + string(capabilities.CapabilityDependencyCache) +
+				"\" capability, so nothing is cached")
+	}
+
+	cfg := config.Config()
+	if cfg.Execution.Id == "" {
+		return executioncache.Unsupported("this step has no execution id, so nothing is cached")
+	}
+
+	client, err := env.Cloud()
```
xoscar, line 55: This a cloud only feature right?

xoscar (approved):
LGTM

cmd/testworkflow-toolkit/commands/cache_repository.go
vsukhin: no, it will work for both OSS and Cloud, but we use local control panel to work with data on local minio and cloud control panel to work with data in GCP

cmd/testworkflow-toolkit/commands/cache.go
vsukhin: I followed the same approach we use for artifacts, so it should be okay for now

PR #8379, opened 2026-09-21
feat(control-plane): [TKC-6839] serve the cache grants, and record git provenance

vsukhin:
Only the scope kind comes from the request. The workflow and the namespace are read from the stored execution, which is the whole reason a workflow-scoped entry cannot be addressed by another workflow however the request is shaped, and why a pull request cannot name the namespace it writes.

A restore consults the execution's own namespace first, then the one it may read but not write, so a pull request starts from what the default branch already built without being able to change what the default branch will later restore.

StoreGitMetadata records the git keys a trigger reported whether or not the workflow declares them, and outside the configuration size limit. StoreConfig keeps only declared parameters, and nothing declares TESTKUBE_GIT_*, so the marker would never have arrived - and its absence means "trusted", which is the one direction that fails open.

A restore now lists under each restore key rather than across the whole scope. The candidate limit applies per listing and a store returns that page in lexical order, so listing the scope and filtering afterwards spent the entire budget on keys that could not match - and in a scope with many entries the matches sat past the end of the page. That surfaced as a miss, or as an older entry winning over a newer one that was never listed. Entries are immutable and leave only on expiry, so a scope reaches that size in the ordinary course of things.

The cost is one listing per restore key rather than one in total, bounded by the ten a step may declare, and each is far narrower than the scope was. Entries are deduplicated across them, because restore keys nest - "npm-" covers everything "npm-v3-" does - and MatchRestore counts a repeated entry as a genuine candidate.

A restore with no usable restore key now lists nothing at all, where it used to list the scope and discard every result.

- [ ] tested locally
- [ ] tested on cluster
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Breaking changes

-

## Changes

-

## Fixes

-


pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,312 @@
+package controlplane
```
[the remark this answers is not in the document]
vsukhin: it's prepared in next layer

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,312 @@
...
+//
+// Only the scope *kind* comes from the request. Which workflow it belongs to is read
+// from the stored execution, and that is the whole reason a workflow-scoped entry cannot
+// be addressed by another workflow however the request is shaped. The namespace comes
+// from the same place and for the same reason: a pull request's run executes code its
+// author wrote, so it must not be able to name the namespace it writes. A finished
+// execution is refused: a cache write from one is either a bug or an attempt to reach a
+// scope the caller should no longer hold.
+func (s *Server) resolveCacheScope(ctx context.Context, executionID string, requested cloud.ExecutionCacheScope) (cacheScope, error) {
+	execution, err := s.resultsRepository.Get(ctx, executionID)
```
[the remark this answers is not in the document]
vsukhin: it's not used in OSS standalone mode

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,312 @@
...
+			// Restore keys may nest - "npm-" covers everything "npm-v3-" does - and
+			// MatchRestore reads a repeated entry as a genuine candidate each time.
+			if _, duplicate := seen[entry.Key]; duplicate {
+				continue
+			}
+			seen[entry.Key] = struct{}{}
+			entries = append(entries, entry)
+		}
+	}
+
```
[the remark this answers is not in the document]
vsukhin: fixed

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,312 @@
...
+			return &cloud.SaveExecutionCachePresignedResponse{AlreadyExists: true}, nil
+		}
+	}
+
+	// What actually makes an entry immutable is the condition on the write. The headers
+	// carrying it are signed in, so an upload that drops them is rejected instead of
+	// becoming an overwrite, and of two executions racing on one key exactly one upload
+	// can succeed. The loser is refused and told, rather than replacing what the winner
+	// stored.
+	url, headers, err := s.storageClient.PresignCreateFileToBucket(ctx, s.cfg.StorageBucket, "", objectName, CachePresignedURLExpiration)
```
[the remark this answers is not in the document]
vsukhin: It's documented behaivour and acceptable rsik

pkg/controlplane/scheduling/enqueuer.go
```diff
@@ -292,6 +292,11 @@ func (e *Enqueuer) prepareExecutions(ctx context.Context, req *cloud.ScheduleReq
 			current.StoreConfig(exec.Config)
 		}
 
+		// Git provenance is recorded whether or not the workflow declares it, and whatever
+		// the configuration weighs, because its absence is meaningful: the dependency cache
+		// reads it to tell a pull request's run from a trusted one.
+		current.StoreGitMetadata(exec.Config)
```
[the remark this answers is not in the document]
vsukhin: fixed

vsukhin:
@greptile

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	for _, entry := range exactEntries {
+		if entry.Key != exactObject {
+			continue
+		}
+		url, err := s.storageClient.PresignDownloadFileFromBucket(ctx, s.cfg.StorageBucket, "", entry.Key, CachePresignedURLExpiration)
+		if err != nil {
+			return nil, err
+		}
+		return &cloud.GetExecutionCachePresignedResponse{
```
xoscar, line 221: Would be possible to combine extact key + restoreKeys or those are mutually exclusive?

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,348 @@
...
+		}
+	}
+	if match.Key == "" {
+		return nil, nil
+	}
+	exact := match.Key == exactObject
+
+	// The folder is left empty and the whole object name passed as the file, so the key
+	// that was matched is exactly the key that gets signed.
+	url, err := s.storageClient.PresignDownloadFileFromBucket(ctx, s.cfg.StorageBucket, "", match.Key, CachePresignedURLExpiration)
```
xoscar, line 286: Does this mean that for a list of restore keys we only return at max one presigned url?

pkg/controlplane/agent_grpc_execution_cache.go
```diff
@@ -0,0 +1,348 @@
...
+	// Skip the upload when the key is already stored. This is only the cheap path: the
+	// lookup and the grant are two steps, so two executions saving the same key can
+	// both be told it is absent.
+	existing, err := s.listCacheEntries(ctx, objectName, 1)
+	if err != nil {
+		return nil, err
+	}
+	for _, entry := range existing {
+		if entry.Key == objectName {
+			return &cloud.SaveExecutionCachePresignedResponse{AlreadyExists: true}, nil
```
xoscar, line 333: We never replace, but rather ignore an overwrite?

xoscar (approved):
LGTM, I added some questions, let me know what you think

pkg/controlplane/agent_grpc_execution_cache.go
vsukhin: yes, exactly

pkg/controlplane/agent_grpc_execution_cache.go
vsukhin: we try to find by exact key first, if it's not found, we fall back to select best candidate from restore keys

pkg/controlplane/agent_grpc_execution_cache.go
vsukhin: yes, we try to find the best available option

